PR #25, opened 2019-10-04
[delete-user-data] Add new param for enabling recursive delete

Ehesp:
This PR allows the `delete-user-data` to perform a recursive delete via the `recursive` select option. By default, this performs a `shallow` delete.

I had to change the `datastore` to `owner` permissions, otherwise the follow error occurs: 

image


delete-user-data/extension.yaml
```diff
@@ -99,6 +102,20 @@ params:
...
+    label: Cloud Firestore delete mode
+    description: >-
+      How would you like Firestore to delete documents? Set to `recursive` to also delete nested Collections and
+      Subcollections.
+    options:
+      - label: Shallow
+        value: shallow
+      - label: Recursive
+        value: recursive
+    default: shallow
```
laurenzlong: should probably default to recursive, since I would imagine that in almost all cases, nested collections would also contain user data that should be deleted.

delete-user-data/functions/src/index.ts
```diff
@@ -117,12 +118,24 @@ const clearFirestoreData = async (firestorePaths: string, uid: string) => {
...
+        logs.firestorePathDeleting(path, false);
+        await admin
+          .firestore()
+          .doc(path)
+          .delete();
+        logs.firestorePathDeleted(path, false);
+      } else {
+        logs.firestorePathDeleting(path, true);
+        await firebase_tools.firestore.delete(path, {
+          project: process.env.GCLOUD_PROJECT,
```
laurenzlong: use process.env.PROJECT_ID, since we set that env variable when deploying an extension so we know it's guaranteed to be there. process.env.GCLOUD_PROJECT is not reliable.

delete-user-data/extension.yaml
Ehesp: I can change this, however it'll be a breaking change for users who update (as it's currently shallow). If that's not an issue I'll change it 😊

laurenzlong: That's not a big concern since when they update, they'll be prompted to pick a value for FIRESTORE_DELETE_MODE (it won't silently set it to recursive). But yes it does technically qualify as a breaking change. This reminds me, can you update the version field in this PR also? (To 0.2.0 to indicate the breaking change).

Ehesp:
@laurenzlong Let me know if it's ok to merge with master. Not sure on the release process?

karayu:
We're working through the release process now. Is it possible to include release notes with this change?

laurenzlong:
Let's discuss during our call tomorrow how we do release notes going forward.

andreimateiro:
Do you have any updates on this issue? Any plans to merge it soon?

delete-user-data/extension.yaml
```diff
@@ -99,6 +102,20 @@ params:
...
+    label: Cloud Firestore delete mode
+    description: >-
+      How would you like Firestore to delete documents? Set to `shallow` to delete only parent documents
+      and ignore documents in subcollections.
+    options:
+      - label: Recursive
+        value: recursive
+      - label: Shallow
+        value: shallow
+    default: recursive
```
karayu: Chatted to @laurenzlong a bit further, can we change the default to be shallow so that this isn't a breaking change? This way the functionality of the extension doesn't automatically change when the user updates, in case the user just goes with defaults when updating.

Ehesp: No problem, updated it.

delete-user-data/extension.yaml
laurenzlong: After internal discussions, we decided to keep it as shallow.

jimmykane:
Hey guys. Just checking. Is this live ?

laurenzlong:
Hey @jimmykane no, it's not live yet, we plan to release it on Thurs. You can check our releases tab to see what's live: https://github.com/firebase/extensions/releases

jimmykane:
Ah yes. Now saw how I can expect an update. Did not know that before.
Thanks for the great work.

PR #32, opened 2019-10-09
fix broken link to functions location page in FGC

rachelsaunders:
- fix broken link to functions location page in firebase.google.com
- make link to top of page rather than to a section anchor
- add missing info about how to select a location


laurenzlong:
Thanks for updating Rachel! @karayu and @rafikhan  I don't think this requires a release since the links weren't actually broken (they went to the top of the page, instead of the intended header)

Issue #34, opened 2019-10-10
firebaseStorageDownloadTokens from object metadata


envyram:
### [READ] Step 1: Are you in the right place?

 Yes

### [REQUIRED] Step 2: Extension name

This feature request is for extension: **\_** (`storage-resize-images`)

### What feature would you like to see?

Any object on firebase storage contains firebaseStorageDownloadTokens from object metadata. However this is not available for resized images

### How would you use it?

I use firebaseStorageDownloadTokens to construct the downloadUrl for the object

rosariopfernandes:
@envyram I'm not sure what you mean. Looking at the [StorageMetadata reference](https://firebase.google.com/docs/reference/android/com/google/firebase/storage/StorageMetadata) I don't see any downloadTokens. Can you tell us how do you usually get this metadata? Is it custom metadata?

envyram:
@rosariopfernandes  

In firebase cloud functions:

exports.thumbnailHandler = functions.storage.object().onFinalize(async (object) => {
   console.log(object.metadata.firebaseStorageDownloadTokens);
}

This is available for all of my other objects in the bucket. But is not available only for the thumbnails created from the extension

rosariopfernandes:
@envyram Thanks for the useful code snippet. I couldn't find this firebaseStorageDownloadTokens anywhere on the docs, and [this post](https://stackoverflow.com/a/56010225/5861618) actually explains why:

> Doug Stevenson pointed out that firebaseStorageDownloadTokens is not an official Google Cloud Storage feature. You won't find it in any Google documentation, and there's no promise it will be in future version of @google-cloud. I like firebaseStorageDownloadTokens because it's the only way to get what I want, but it has a "smell" that it's not safe to use.

Since this is not official, I'm not sure if we should have it on Firebase Extensions, but I'll let the Firebase Team decide.

envyram:
@rosariopfernandes that's a great discussion. "getDownloadURL() from Node" is what I am looking for actually. Without that the extension on Firebase is not quite useful

jhuleatt:
I chatted with the Cloud Storage for Firebase team about `firebaseStorageDownloadTokens`. The field should be unique per object. If it isn't set on an object, it will be set automatically by the Cloud Storage for Firebase api when `getDownloadURL()` is called.

TL;DR we shouldn't copy it over in the extension.

rosariopfernandes:
@jhuleatt thanks!

[...]

Issue #20, opened 2019-10-01
[BUG] name of resized image is missing original image name if there was no file type extension


emrebicer:
### [READ] Step 1: Are you in the right place?
Yes
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **\_** (`storage-resize-images`)
- Configuration values (redact info where appropriate):
  -Size:64x64
  -Cloud Storage Path: empty
### [REQUIRED] Step 3: Describe the problem
I just added storage-resize-images extension to my firebase project, it is working as expected when i upload an image manually from the firebase console, but the problem is that whenever i use 

```terminal
storageRef.putString()
```
function to upload base64 url encoded image to the firebase storage cloud, the compressed image will be created but its name is only _64x64 (not original_64x64).

Obviously i am resizing images as 64x64.

Since the created file doesn't contain a unique name it is unaccessible for me dynamically...Is there something am i missing?

#### Steps to reproduce:
Install storage-resize-images extension to my firebase project.(web)
##### Expected result
storage-resize-images extension to create an image file using the "original.jpeg" image as "original_64x64.jpeg"
##### Actual result
The extension generated the image as expected but the naming was wrong, it is _64x64.jpeg instead of original_64x64.jpeg

Salakar:
Hey @emrebicer thanks for reporting this issue, will get this looked at.

Salakar:
You can set a file extension when using `putString` by  specifying a the file name when creating a storage reference:

```js
const imageStorageRef = firebase.storage().ref('images/space.jpg');
// ...
imageStorageRef.putString(/* ... */) // ..
```

Is there some other case where it would never set a file extension?

emrebicer:
Looks like 

> You can set a file extension when using `putString` by specifying a the file name when creating a storage reference:
> 
> ```js
> const imageStorageRef = firebase.storage().ref('images/space.jpg');
> // ...
> imageStorageRef.putString(/* ... */) // ..
> ```
> 
> Is there some other case where it would never set a file extension?

After adding ".jpg" at the end of my file names the problem went away. It looks like a proper solution for me. Closing the issue, thank you :).

christophe-g:
Still experiencing this when an uploaded file does not have an extension.

I am aware that I can check whether filename has an extension before uploading, but IMO the resize image extension should be able to handle case without file extension. 

Can this be re-opened ?

[...]

PR #6, opened 2019-09-26
[storage-resize-images] delete original image from bucket after resize

rosariopfernandes:
As requested in #4, this PR should give developers an option to remove the original image after it has been resized.

PR #42, opened 2019-10-14
feat(storage-resize-images): copy metadata to resized image

rosariopfernandes:
Extension name: storage-resize-images

This PR should make the extension copy the following metadata from the original file to the resized image:

- content disposition
- content encoding
- content language
- content type (was already copied over before)
- user-provided metadata (except Firebase storage download tokens)
- cache control (will get overridden by param if it is configured)

Full list of metadata is described at https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON


storage-resize-images/functions/src/index.ts
```diff
@@ -153,6 +157,12 @@ const resizeImage = async ({
         resizedImage: "true",
       },
     };
+    if (config.keepFileMetadata) {
```
Ehesp: Reduce nesting?:

`if (config.keepFileMetadata && originalMetadata) {`

storage-resize-images/extension.yaml
```diff
@@ -114,6 +114,19 @@ params:
       Invalid sizes, must be a comma-separated list of WIDTHxHEIGHT values.
     required: true
 
+  - param: KEEP_FILE_METADATA
+      type: select
+      label: Keep Original Custom Metadata? (Yes/No)
+      description: >-
+        Copy the custom metadata from the original image to the resized image?
```
Ehesp: It may be best to imply there may not be any existing metadata? cc @rachelsaunders 

> Copy any custom metadata from the original image to the resized image?

Ehesp:
Thanks for this PR :)

cc @karayu should be put a blocker on this one once we decide on https://github.com/firebase/extensions/pull/6? We'll have to resync this branch with master once it lands anyway.

storage-resize-images/functions/src/index.ts
rosariopfernandes: Done!

rosariopfernandes: Done :)

laurenzlong:
I wonder if would be better to always copy over metadata from the original image to the resized images, rather than having it as an option? I want to prevent option overload, and I also cannot think of a good reason for why someone would not want their custom metadata copied over to the resized images. Anyone have thoughts on this?

rosariopfernandes:
@laurenzlong I agree this shouldn't necessarily be an option. I'll wait for one more opinion before updating the PR.

laurenzlong:
@karayu @rachelsaunders thoughts?

GeeDollaHolla:
> @laurenzlong I agree this shouldn't necessarily be an option. I'll wait for one more opinion before updating the PR.

I'd second not having this be an option. :) 

Also, I just created [issue 72](https://github.com/firebase/extensions/issues/72). I think this PR will satisfy this.

laurenzlong:
Hey @rosariopfernandes Would you mind updating to make this not an option? Thanks!

rosariopfernandes:
Sure thing @laurenzlong !

laurenzlong:
Thanks! @rachelsaunders can you please review for copy?

jhuleatt:
While we're at it, is it worth copying over the other metadata too? I believe [this is the full list](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON) of settable metadata fields.

laurenzlong:
@jhuleatt Good point! We should probably treat the fields on a case to case basis. For example: acl[], eventBasedHold, and temporaryHold probably do not make sense to copy over as they affect access and usage of the file (and an app developer may not want  to treat both original and resized image in the same way when it comes to access). crc32c and md5hash probably also do not make sense, as they relate directly to the data of the original image. cacheControl is already a param in our extension installation. The rest may make sense to copy over.

rachelsaunders:
@jhuleatt and I were also chatting about whether we _would_ want to copy over Cache-Control metadata to the resized image, BUT if the dev set a Cache-Control param then the param value would overwrite the copied over value for the resized image.  Thoughts?

rosariopfernandes:
@jhuleatt @rachelsaunders I like both ideas!

@laurenzlong Should we also copy the `name` param? Won't it cause any conflict with the name we set for resized images?

laurenzlong:
@rosariopfernandes Agreed we shouldn't copy over name, since the resized image has a different name. I also wonder if we actually need to copy over `contentDisposition`, `contentEncoding`, `contentLanguage` and `contentType`. I have a feeling they are already preserved in the new image, that's how they remain an image file (otherwise they would just end up as binary blobs). It would be interesting to print out the full metadata object of the resized image to see what metadata already exists. @rosariopfernandes is this something you can do?

@rachelmyers I do like the idea of copying over cacheControl, but overriding it if that param has been configured. I think in a future update, we should probably just remove the cache control param. During user research, it confused a lot of people during the installation process. And if we preserved the cache control header from the original image,that should satisfy power users that actually set cache control headers. But that would count as a breaking update, so we shouldn't make it now. But in this PR, do what you suggest (copy over the original cache control header, but allow it to be overriden if the param has been set). 

To summarize, depending on findings on the question I had in my first paragraph. `metadata.metadata`, `metadata.storageClass`, and `metadata.cacheControl` might actually be the only ones that make sense to copy over.

GeeDollaHolla:
@laurenzlong, I do like the idea of using the original's `contentType`, if the name does not have a file extension. I prefer to not use file extensions (opting to use `contentType` to manage that for me), and the current resizer only looks at the ext in the name. Besides being annoying, I think this is the root of [issue 20](https://github.com/firebase/extensions/issues/20)

rosariopfernandes:
@laurenzlong I tried printing the metadata of both images to see what is actually preserved. I uploaded an image (using the Firebase Console) which doesn't contain `contentLanguage` or `contentEncoding`. But after resize, I noticed `contentType` and `storageClass` were the only properties preserved, as shown in the log bellow (for brevity, I omitted non-writable properties and the ones we had already discussed):
```javascript
// Original Image
object { 
  contentDisposition: 'inline; filename*=utf-8\'\'my_image.png',
  contentType: 'image/png',
  metadata: { firebaseStorageDownloadTokens: '75a36eed-357a-4c2b-b2e6-667542774c3c' },
  storageClass: 'STANDARD',
}

// Resized image
object {
  contentType: 'image/png',
  metadata: { resizedImage: 'true' },
  storageClass: 'STANDARD',
}
```

Just like `contentDisposition`, I suspect `contentEncoding` and `contentLanguage` are also not preserved. Should we copy those over?

---

@rachelsaunders As suggested, I've updated the PR to copy over the original `cacheControl` if not set on the extension config. Please take a look

jhuleatt:
@rosariopfernandes thanks for printing out the metadata! It looks like your test revealed a field that we *don't* want to copy over: `metadata.metadata.firebaseStorageDownloadTokens`. I've added more context on that in [this issue](https://github.com/firebase/extensions/issues/34#issuecomment-552595619).

I agree that it makes sense to copy `cacheControl`, `contentDisposition`, `contentEncoding`, and `contentLanguage`. They seem like fields that would apply to both the original and resized images.

rosariopfernandes:
Thanks @jhuleatt ! Maybe we should check if the original image contains the `metadata.metadata.firebaseStorageDownloadTokens` field, and if it does we'll set it to `undefined` on the resized image, since it will be updated with its correct value when `getDownloadUrl()` is called. 

I've also updated this PR to copy `contentDisposition`, `contentEncoding`, and `contentLanguage`.

storage-resize-images/functions/lib/config.js
```diff
@@ -20,5 +20,5 @@ exports.default = {
     cacheControlHeader: process.env.CACHE_CONTROL_HEADER,
     imageSizes: process.env.IMG_SIZES.split(","),
     resizedImagesPath: process.env.RESIZED_IMAGES_PATH,
-    deleteOriginalFile: process.env.DELETE_ORIGINAL_FILE,
+    deleteOriginalFile: process.env.DELETE_ORIGINAL_FILE === "true",
```
jhuleatt: Was this an unintentional change? It doesn't look like it is related to metadata.

jhuleatt:
@rosariopfernandes removing the field sounds good to me! I think [`delete`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) would be the proper way: 

```js
delete metadata.metadata.firebaseStorageDownloadTokens;
```

`delete` works whether the field is there or not, so I don't think we need to check for the existence `firebaseStorageDownloadTokens` before doing the `delete` call.

storage-resize-images/functions/lib/config.js
laurenzlong: That will go away once all the branches are sync'ed. That was the artifact of a previous PR.

laurenzlong: That's due to the "next" branch having outdated "lib" files, which is what caused our rollback last Friday. Let's be double sure that the lib files in this PR are up to date before merging this time.

laurenzlong:
@rosariopfernandes @jhuleatt all of those are great ideas!

rosariopfernandes:
All done! Ready for another review.

laurenzlong:
Thanks! Looks good to me (assuming that you've run "npm run build" one more time since the changes). 

@rachelsaunders I've revised the PR description to enumerate the list of what's now being copied. Please review and give feedback on whether any copy needs to change.

rosariopfernandes:
@laurenzlong I did run `npm run build`. Thanks!

@rachelsaunders I've added the TODOs.

rosariopfernandes:
@rachelsaunders The PRE/POST files LGTM.

Added the TODO to the extension.yaml file! :)

storage-resize-images/extension.yaml
```diff
@@ -143,9 +143,11 @@ params:
...
     description: >
-      Do you want to specify a `Cache-Control` header for the resized image
-      files? Learn more about
+      This extension copies over to the resized images any `Cache-Control` header metadata from the original image.
+      For the resized images, do you want to overwrite this `Cache-Control` header metadata or add
+      `Cache-Control` header metadata? Learn more about
       [`Cache-Control` headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).
-      If you prefer not to use a `Cache-Control` header, leave this field empty.
+      If you prefer not to overwrite or add `Cache-Control` header metadata, leave this field empty.
+      TODO: @rachelsaunders
```
rosariopfernandes: I think we can now remove this
```suggestion
```

rachelsaunders: lol - yep!  Done :-D

storage-resize-images/POSTINSTALL.md
```diff
@@ -12,13 +12,24 @@ You can test out this extension right away:
...
 
-You can upload images using the [Cloud Storage for Firebase SDK](https://firebase.google.com/docs/storage/) for your platform (iOS, Android, or Web). Alternatively, you can upload images directly in the Firebase console's Storage dashboard.
+The extension also copies over to the resized image(s) the following [metadata](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON) associated with the original image:
+
+- content disposition
+- content encoding
+- content language
+- content type
+- user-provided metadata (except Firebase storage download tokens)
+- Cache-Control header
```
jhuleatt: I think these should either match the [docs](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON) exactly ("content disposition" -> "contentDisposition"), or match the IETF spec names ("content disposition" -> "[Content-Disposition](https://tools.ietf.org/html/rfc6266)") so that it is as easy as possible for a developer to figure out what each header does.

storage-resize-images/POSTINSTALL.md
```diff
@@ -12,13 +12,24 @@ You can test out this extension right away:
...
+
+Note that if you configured the `Cache-Control header for resized images` param, the specified value will overwrite the value copied over with the original image.
```
jhuleatt: Tiny nit: maybe "copied from the original image" instead of "copied over with" the original image?

storage-resize-images/PREINSTALL.md
```diff
@@ -3,12 +3,12 @@ Use this extension to create resized versions of an image uploaded to a Cloud St
...
 - Names the resized image using the same name as the original uploaded image, but suffixed with your specified width and height.
+- Stores the resized image in the same Storage bucket as the original uploaded image.
 
 You can even configure the extension to create resized images of different dimensions for each original image upload. For example, you might want images that are 200x200, 400x400, and 680x680 - this extension can create these three resized images then store them in your bucket.
 
-Another optional feature of this extension is to specify a [`Cache-Control` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) for your resized image files.
+This extension automatically copies over to the resized image(s) the following metadata associated with the original image: `Cache-Control` header, content disposition, content encoding, content language, content type, and user-provided metadata (except Firebase storage download tokens). Note that you can optionally configure the extension to overwrite the [`Cache-Control` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) metadata for the resized image(s).
```
jhuleatt: Suggested shortening of the first sentence: "This extension automatically copies the following metadata from the original image to the resized image"

jhuleatt: same suggestion as above: let's either make these names match the docs exactly, or match the IETF spec names exactly

also a formatting note, want to format these as bullets like in the postinstall?

storage-resize-images/POSTINSTALL.md
rachelsaunders: Cool - agree.  I've changed them to the IETF spec names and added links to the mozilla dev docs for each.  See the recent commit.

storage-resize-images/PREINSTALL.md
rachelsaunders: good suggestion!  done (throughout, too)

rachelsaunders: changed to IETF names

purposefully made an inline list b/c of space concerns in extension details card

storage-resize-images/POSTINSTALL.md
rachelsaunders: good suggestion!  done (throughout, too)

storage-resize-images/POSTINSTALL.md
```diff
@@ -12,13 +12,24 @@ You can test out this extension right away:
...
 
-You can upload images using the [Cloud Storage for Firebase SDK](https://firebase.google.com/docs/storage/) for your platform (iOS, Android, or Web). Alternatively, you can upload images directly in the Firebase console's Storage dashboard.
+The extension also copies the following metadata, if present, from the original image to the resized image(s):
+
+- [`Cache-Control`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Cache-Control)
+- [`Content-Disposition`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Disposition)
+- [`Content-Encoding`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Encoding)
+- [`Content-Language`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Language)
+- [`Content-Type`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Type)
+- user-provided metadata (except Firebase storage download tokens)
```
laurenzlong: Perhaps link to https://cloud.google.com/storage/docs/metadata#custom-metadata for user-provided metadata.

rachelsaunders: done :-)

PR #44, opened 2019-10-15
[storage-resize-images] Update to v0.1.1

laurenzlong:
Format of changelog will probably change in the future (since we may be automating the generation of it). But this is just something to have for now.

### DIFF
--- a/storage-resize-images/CHANGELOG.md
+++ b/storage-resize-images/CHANGELOG.md
@@ -1,0 +1,2 @@
+## Version 0.1.1
+fixed - Fixed bug where certain edge cases led to already resized image being resized again. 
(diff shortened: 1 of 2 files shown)


rafikhan:
LGTM

karayu:
LGTM; Let's not forget to create a release in GitHub when this goes out

laurenzlong:
No. We should probably just remove the version number in package.jsons for all the extensions. It's only the version number in extension.yaml that counts, and it's just upkeep to try to keep the two in sync, and confusion when they're out of sync.

bkendall:
That's not a horrible idea, in all honesty. These will never be published via `npm`, so it shouldn't break anything to remove them

PR #45, opened 2019-10-15
Refactor service-agnostic URL shortener code into common module

kevinthecheung:
This PR is to prepare for a Dynamic Links based extension (to follow).


firestore-shorten-urls-bitly/extension.yaml
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 name: firestore-shorten-urls-bitly
-displayName: Shorten URLs
+displayName: Shorten URLs with bit.ly
```
laurenzlong: Don't change the display name in this PR, as that requires marketing review. (We'll do that review shortly before releasing the FDL version)

laurenzlong:
Looks good from my perspective. Mike and Elliot, could you check to ensure the refactor didn't break anything?

PR #46, opened 2019-10-15
[firestore-translate-text] allow single language list

### LINKED ISSUE #43: Target languages field fails to accept one language
- Extension name: `firestore-translate-text`
  - **Target langs: ar**

#### Steps to reproduce:

1. Use the console or CLI to install the `firestore-translate-text` extension. 
2. In the target languages field list only one language for example `ar` 

##### Expected result

I except for the target languages field to be able to accept ONE target languages

##### Actual result

The target languages field only accepts MORE than one language

image

OR

image

rosariopfernandes:
This PR should:
- Change the regex validation of the `languages` field to accept a single language list;
- Use a `Set` so that the extension doesn't translate to the same language more than once;
- Fix #43

### DIFF
--- a/firestore-translate-text/extension.yaml
+++ b/firestore-translate-text/extension.yaml
@@ -90,7 +90,7 @@ params:
       For these codes, visit the [supported languages list](https://cloud.google.com/translate/docs/languages).
     default: en,es,de,fr
     example: en,es,de,fr
-    validationRegex: "^[a-zA-Z,-]+[a-zA-Z-]{2,}$"
+    validationRegex: "^[a-zA-Z,-]*[a-zA-Z-]{2,}$"
     validationErrorMessage:
       Languages must be a comma-separated list of ISO-639-1 language codes.
     required: true
--- a/firestore-translate-text/functions/src/config.ts
+++ b/firestore-translate-text/functions/src/config.ts
@@ -15,7 +15,9 @@
  */
 
 export default {
-  languages: new Set(process.env.LANGUAGES.split(",")),
+  languages: Array.from(
+    new Set(process.env.LANGUAGES.split(","))
+  ),
   location: process.env.LOCATION,
   inputFieldName: process.env.INPUT_FIELD_NAME,
   outputFieldName: process.env.OUTPUT_FIELD_NAME,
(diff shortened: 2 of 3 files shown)


firestore-translate-text/extension.yaml
```diff
@@ -90,7 +90,7 @@ params:
       For these codes, visit the [supported languages list](https://cloud.google.com/translate/docs/languages).
     default: en,es,de,fr
     example: en,es,de,fr
-    validationRegex: "^[a-zA-Z,-]+[a-zA-Z-]{2,}$"
+    validationRegex: "^[a-zA-Z,-]*[a-zA-Z-]{2,}$"
```
Ehesp: Just to confirm this works:

https://regex101.com/r/99E00o/4

firestore-translate-text/functions/src/config.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 export default {
-  languages: process.env.LANGUAGES.split(","),
+  languages: new Set(process.env.LANGUAGES.split(",")),
```
Ehesp: To save duplication, it may be better to do:

```
languages: Array.from(
  new Set(process.env.LANGUAGES.split(","))
),
``

That way, `config.languages` is always an array throughout the extension rather than a Set.

File: firestore-translate-text/functions/src/index.ts [hunk 3]
```diff
@@ -147,9 +147,10 @@ const translateDocument = async (
 ): Promise<void> => {
   const input: string = extractInput(snapshot);
 
-  logs.translateInputStringToAllLanguages(input, config.languages);
+  const languages = Array.from(config.languages);
```
Ehesp: See config file note.

firestore-translate-text/functions/src/index.ts
```diff
@@ -55,7 +55,7 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
       validators.fieldNameIsTranslationPath(
         inputFieldName,
         outputFieldName,
-        languages
+        Array.from(languages)
```
Ehesp: See config file note.

rosariopfernandes:
@Ehesp Thanks for your review! I've addressed the suggested changes. :)

PR #49, opened 2019-10-16
changed(firestore-counter): use pubsub trigger for controller function

Salakar:
This PR makes the following changes:

 - swap out the `https` controller function for a `pubsub` topic on publish function (as discussed previously)
 - updates the `POSTINSTALL` doc to use a pubsub schedule `gcloud` command. [[docs]](https://cloud.google.com/scheduler/docs/creating#)
 - fixed some small typos in the code

----

I also noticed there's several unused imports & exports in the extension, but I've not touched these - I can update this PR to remove them if you'd accept it as part of this PR;

 - https://github.com/firebase/extensions/blob/39a845f8a0eb6fc84a4342d8b4a8cf941e113a33/firestore-counter/functions/src/common.ts#L85 (`delay` fn is not used anywhere)
 - https://github.com/firebase/extensions/blob/39a845f8a0eb6fc84a4342d8b4a8cf941e113a33/firestore-counter/functions/src/aggregator.ts#L18 (FieldPath is unused)
 - https://github.com/firebase/extensions/blob/39a845f8a0eb6fc84a4342d8b4a8cf941e113a33/firestore-counter/functions/src/common.ts#L18-L20 (all 3 import statements unused)

---

Thanks


firestore-counter/extension.yaml
```diff
@@ -47,7 +47,9 @@ resources:
       sourceDirectory: .
       location: ${LOCATION}
       maxInstances: 1
-      httpsTrigger: {}
+      eventTrigger:
+        eventType: google.pubsub.topic.publish
+        resource: projects/${PROJECT_ID}/topics/firestore-counter
```
laurenzlong: Can you make the topic the same as the ID of the instance? It is available as ${MOD_INSTANCE_ID}. That way, if someone installs this extension more than once, it will still work properly. 

This also needs to be changed in the POSTINSTALL. 

Meanwhile, I'm going to change the backend to also populate EXT_INSTANCE_ID, since the product is no longer called Mods. Hopefully, that'll be in prod before we release this update, then you can change MOD_INSTANCE_ID with EXT_INSTANCE_ID before merging the PR.

firestore-counter/functions/lib/index.js
```diff
@@ -38,7 +38,7 @@ const WORKERS_COLLECTION_ID = "_counter_workers_";
  * there's less than 200 of them. Otherwise it is scheduling and monitoring
  * workers to do the aggregation.
  */
-exports.controller = functions.https.onRequest((req, res) => __awaiter(void 0, void 0, void 0, function* () {
+exports.controller = functions.pubsub.topic("firestore-counter").onPublish(() => __awaiter(void 0, void 0, void 0, function* () {
```
laurenzlong: `functions.handler.pubsub.topic.onPublish(...`

firestore-counter/functions/src/index.ts
```diff
@@ -31,7 +31,7 @@ const WORKERS_COLLECTION_ID = "_counter_workers_";
  * there's less than 200 of them. Otherwise it is scheduling and monitoring
  * workers to do the aggregation.
  */
-export const controller = functions.https.onRequest(async (req, res) => {
+export const controller = functions.pubsub.topic("firestore-counter").onPublish(async () => {
```
laurenzlong: `functions.handler.pubsub.topic.onPublish`

firestore-counter/functions/lib/index.js
laurenzlong: See https://docs.google.com/document/d/1cthryFQXsV5kqS366W5emljvmePPo1c7etwTShvDnHQ/edit#heading=h.vk7i214jq4av

laurenzlong:
Yes, please remove the unused imports. Thanks!

firestore-counter/extension.yaml
Salakar: Ah this makes more sense, pushed up a change for this - thanks!

(will update to `EXT_INSTANCE_ID` when available)

firestore-counter/functions/src/index.ts
Salakar: Done, thanks!

laurenzlong:
@Salakar you can now update to EXT_INSTANCE_ID. Let me know if it doesn't work,

Salakar:
> @Salakar you can now update to EXT_INSTANCE_ID. Let me know if it doesn't work,

Cool, thanks - I've pushed up the change, will test on Monday (as I can't deploy extensions from local source on the latest tools EAP on my home Mac - I'll follow up on the issue separately but have been using another Mac which has the old EAP installed which I don't have access to right now)

Salakar:
Have tested and confirmed all working. I made some small changes to the gcloud commands in the `POSTINSTALL` file to fix some small issues with them.

This is ready for final review & merge now, thanks

cc @laurenzlong

laurenzlong:
We are going to delay the merge (and launch) of this PR for a few weeks. We need to create better UX for breaking changes first. At the moment, users won't be sufficiently alerted of breaking changes when they update.

laurenzlong:
@Salakar I realized that cloud functions does not support update of trigger type. So this won't work. Can you rename the function another name? (That way, updating the extension will actually delete the old HTTPS function and create a new PubSub function)

Ehesp:
@googlebot I consent.

Ehesp:
@laurenzlong `controller` has now been renamed to `aggregator`

firestore-counter/functions/lib/controller.js
```diff
@@ -191,7 +191,7 @@ class ShardedCounterController {
                     yield t.get(this.controllerDocRef);
                 }
                 catch (err) {
-                    console.log("Failed to read controler doc " + this.controllerDocRef.path);
+                    console.log("Failed to read controller doc " + this.controllerDocRef.path);
```
rachelsaunders: Should this be "Failed to read aggregator doc "

rachelsaunders: Actually, this whole script and its contents use "controller" rather than "aggregator".  Should everything be updated to align with the other updates to the extension in this PR?  Or am I not understanding where the "controller" vs "aggregator" changes are applicable.

laurenzlong: Hmmm I'm thinking that we don't rename to `aggregator` since it may confuse existing users (we haven't change the functionality of it, we've just moved it from HTTP to PubSub). @rachelsaunders for context, we have to rename the function in order to change the trigger type. Can you think of a better name? `controller-ps`?

firestore-counter/POSTINSTALL.md
```diff
@@ -20,13 +20,13 @@ match /databases/{database}/documents/pages/{page} {
 
 #### Set up a scheduled function
 
-Review the [scheduled function documentation](https://firebase.google.com/docs/functions/schedule-functions) to set up a call to `${function:controller.url}` every minute. You may need to enable some APIs in your Firebase project to use scheduled functions.
+Review the [scheduled function documentation](https://firebase.google.com/docs/functions/schedule-functions) to set up a call to `${function:aggregator.url}` every minute. You may need to enable some APIs in your Firebase project to use scheduled functions.
```
laurenzlong: Revise to:

Review the [scheduled function documentation](https://firebase.google.com/docs/functions/schedule-functions) to send a message to the PubSub topic `${param:EXT_INSTANCE_ID}` every minute.

firestore-counter/POSTINSTALL.md
```diff
@@ -20,13 +20,13 @@ match /databases/{database}/documents/pages/{page} {
...
 
 As an example, to set up a scheduled function, you can run the following [`gcloud`](https://cloud.google.com/sdk/gcloud/) commands:
```
laurenzlong: Revise to: 
You can set up a new Cloud Scheduler Job using the [Cloud Console](https://console.cloud.google.com/cloudscheduler/jobs/new) or by running the following [`gcloud`](https://cloud.google.com/sdk/gcloud/) commands:

firestore-counter/functions/lib/controller.js
laurenzlong: Rachel and I discussed, and our best idea so far is changing it to `controllerFunc` so we are changing the name as little as possible and misleading people into thinking that the functionality has changed.

laurenzlong:
@Salakar We actually realized that there's a way to do this update without it being a breaking change. (Context: HTTP functions that get updated after Jan 15 will still remain public, only newly created HTTP functions will be private)

Can you:
- Keep the new pubsub function, but add back an http function called `controller`. The http function becomes a trampoline that gets a request and simply dumps a message into the pubsub topic. The pubsub function will actually have the business logic
- the http function's name will be unchanged (`controller`) so that it would constitute an in-place update of the function, and it will remain public for existing users
- the new pubsub topic can perhaps be named `controller-core`? @rachelsaunders thoughts?
- the postinstall should still instruct users to set up a Cloud Scheduler that publishes to the pubsub topic. But existing users that have schedules that call the HTTP function will still have everything working as before. In the release notes, we put up a "deprecation notice" saying that they should change their schedule to call the pubsub topic instead. (You don't have to worry about the release notes in this PR)

If you can make the change early next week that would be ideal, so we can test it out. Thanks!

firestore-counter/functions/lib/controller.js
laurenzlong: This discussion is now out of date, see https://github.com/firebase/extensions/pull/49#issuecomment-559302743

russellwheatley:
Hi Lauren,
	     As per your request, we have:
- Reinstated the `controller` as a http function that triggers an event on the relevant subscriber (`onPublish`).
- Renamed the subscriber to `coreController`.
- Updated the relevant metadata in `extension.yaml`.

Notes
- We have updated the roles to include `pubsub.publisher`, and we aren’t sure if this constitutes a breaking change.
- I have signed the CLA but I hadn't updated my username on my local machine when I committed the code which is why googlebot doesn't recognise one of the 'co-authors'.

firestore-counter/PREINSTALL.md
```diff
@@ -18,7 +18,7 @@ Before installing this extension, make sure that you've [set up a Cloud Firestor
 After installing this extension, you'll need to:
 
 - Update your [database security rules](https://firebase.google.com/docs/rules).
-- Set up a [scheduled function](https://firebase.google.com/docs/functions/schedule-functions) to regularly call the controller function, which is created by this extension and monitors the extension's workload.
+- Set up a [scheduled function](https://firebase.google.com/docs/functions/schedule-functions) to regularly call the aggregator function, which is created by this extension and monitors the extension's workload.
```
laurenzlong: Change `aggregator`  to `controller`

firestore-counter/functions/src/index.ts
```diff
@@ -16,36 +16,51 @@
...
 firestore.settings({ timestampsInSnapshots: true });
 
+const pubsub = new PubSub();
+
 const SHARDS_COLLECTION_ID = "_counter_shards_";
 const WORKERS_COLLECTION_ID = "_counter_workers_";
 
 /**
- * Controller is scheduled every minute. It tries to aggregate shards if
+ * The aggregator is scheduled every minute. It tries to aggregate shards if
```
laurenzlong: Aggregator > Controller

russellwheatley:
I've updated the naming in the `PREINSTALL.md` & `POSTINSTALL.md` to match the exported function naming. 

Please be aware that I've named the pubsub function `controllerCore` as per your previous comment (pending agreement from @rachelsaunders), and reinstated `controller` as the legacy http function. therefore, the docs make reference to `controllerCore`.

In light of testing this extension numerous times (installing and uninstalling). It came to a point where I was unable to publish further changes due to this error:

`
Error: Error occurred while deploying resources: RESOURCE_ERROR: '/deployments/firebase-ext-firestore-counter/resources/worker' Deployment Field Violation Error: gcp-types/cloudfunctions-v1:projects.locations.functions: INVALID_ARGUMENT: Field: event_trigger (Expected value test_shard/workers/{workerId} to match regular expression [^/]+/[^/]+(/[^/]+/[^/]+)*)
`

I asked @Ehesp to publish from his machine in order to test in my place. He also encountered another error which stopped him from publishing any further:

`
Error: Error occurred while deploying resources: RESOURCE_ERROR: '/deployments/firebase-ext-firestore-counter/resources/controllerCore' {“ResourceType”:“gcp-types/cloudfunctions-v1:projects.locations.functions”,“ResourceErrorCode”:“400",“ResourceErrorMessage”:“Build failed: Build has timed out”}
`

We haven't been able to test that this still works since changing to `coreController`.

I haven't encountered errors publishing any of the other extensions, only `firestore-counter`.

laurenzlong:
@russellwheatley Thanks for discovering these really important issues! The first error is due to mis-formed document path (I've pushed up a commit that adds regex validation). I managed to repro the second error when I installed this extension on a project that has the Cloud Function and PubSub APIs turned off (i.e. never used Cloud Functions or Extensions before). I think it's a subtle race condition with when APIs are enabled that caused the build to fail. We'll work on fixing the race condition on the backend, but in the meanwhile I've moved the initialization code for the PubSub instance into the function body itself. So it no longer gets run during the build step when the function is first deployed. @Ehesp can you try again to see if the issue has been resolved?

@rachelsaunders I think this PR is now ready for a last review of copy, and any modifications you wish to make to change Scheduled Function to Cloud Scheduler job like we discussed. I think it's probably easier for you to directly make a commit.

firestore-counter/POSTINSTALL.md
```diff
@@ -18,17 +18,26 @@ match /databases/{database}/documents/pages/{page} {
...
+**IMPORTANT:** Note the following about v0.1.1 of this extension:
+- **If you updated your extension from v0.1.0 to v0.1.1:**  We recommend that you edit your Cloud Scheduler job to instead send a message to the extension's Pub/Sub topic, as described in this section. Although it's not recommended, if you leave your Cloud Scheduler job calling `${function:controller.url}`, your extension will continue to run as expected. For more information about the changes for v0.1.1, refer to the [changelog](https://github.com/firebase/extensions/blob/next/firestore-counter/CHANGELOG.md).
+- **If you installed this extension for the first time at v0.1.1:** Follow the instructions as described in this section.
+
+
+Set up a [Cloud Scheduler job](https://firebase.google.com/docs/functions/schedule-functions) to regularly send a message to the extension's Pub/Sub topic (`${param:EXT_INSTANCE_ID}`). This Pub/Sub topic then automatically triggers the Pub/Sub controller function (`${function:controllerPubSub}`). This Pub/Sub controller function is created by the extension and monitors the extension's workload.
```
rachelsaunders: @laurenzlong What is the final name of the Pub/Sub controller function?  It looks like it's "controllerCore"?

rachelsaunders: And I think that we should _probably_ update our copy throughout to just use "controllerCore function" to describe this function, i.e., leave out the "Pub/Sub" bit.  WDYT?

For example:
"... triggers the controllerCore function (...). This controllerCore function is created by..."

firestore-counter/extension.yaml
```diff
@@ -36,13 +36,28 @@ contributors:
...
+      maxInstances: 1
+      eventTrigger:
+        eventType: google.pubsub.topic.publish
+        resource: projects/${PROJECT_ID}/topics/${EXT_INSTANCE_ID}
+
+  - name: controller
+    type: firebaseextensions.v1beta.function
+    description:
+      Maintained for backwards compatibility.
+      This function will relay the message to the onPublish function.
```
rachelsaunders: If I understand what this function does in the grand scheme of things, I would recommend the following edit:

This function relays a message to the extension's Pub/Sub topic to trigger the controllerCore function.

firestore-counter/extension.yaml
```diff
@@ -107,4 +122,8 @@ params:
       What is the path to the document where the extension can keep its internal state?
     default: _firebase_ext_/sharded_counter
     example: _firebase_ext_/sharded_counter
+    validationRegex: "^[^/]+/[^/]+(/[^/]+/[^/]+)*$"
+    validationErrorMessage:
+      Please enter a document path, not a collection path. It should have an even
```
rachelsaunders: _(If I make this change in the file directly, we'll lose the comment above)_

Enter a document path, not a collection path. The path must have an even number of segments, for example: `my_collection/doc` or `my_collection/doc/subcollection/doc`, but not `my_collection`.

rachelsaunders: _(If I make this change in the file directly, we'll lose the comment above)_

Enter a document path, not a collection path. The path must have an even number of segments, for example: `my_collection/doc` or `my_collection/doc/subcollection/doc`, but not `my_collection`.

firestore-counter/PREINSTALL.md
```diff
@@ -18,7 +18,7 @@ Before installing this extension, make sure that you've [set up a Cloud Firestor
 After installing this extension, you'll need to:
 
 - Update your [database security rules](https://firebase.google.com/docs/rules).
-- Set up a [scheduled function](https://firebase.google.com/docs/functions/schedule-functions) to regularly call the controller function, which is created by this extension and monitors the extension's workload.
+- Set up a [Cloud Scheduler job](https://firebase.google.com/docs/functions/schedule-functions) to regularly call the controllerCore function, which is created by this extension and monitors the extension's workload.
```
rachelsaunders: I'd like to update this docs hyperlink to: https://cloud.google.com/scheduler/docs/quickstart  WDYT?
(we'd do it throughout)

There's really no content on the Firesite page about Cloud Scheduler jobs.

firestore-counter/extension.yaml
```diff
@@ -36,13 +36,28 @@ contributors:
 roles:
   - role: datastore.user
     reason: Allows the extension to aggregate Cloud Firestore counter shards.
+  - role: pubsub.publisher
+    reason: Allows the HTTPS function to trigger the PubSub handler.
```
rachelsaunders: 1) PubSub -> Pub/Sub

2) We've not used the phrase "the Pub/Sub handler" anywhere else.  Should this be "... to trigger the Pub/Sub controllerCore function."?

russellwheatley:
@laurenzlong We have now tested. It is ready to merge once naming & comments have been agreed to, and updated.

firestore-counter/PREINSTALL.md
laurenzlong: SGTM

firestore-counter/extension.yaml
laurenzlong: Yes,that's more accurate.

firestore-counter/extension.yaml
laurenzlong: More specifically: "Allows the HTTPS controller function to publish a message to Pub/Sub, to trigger the controllerCore function"

firestore-counter/PREINSTALL.md
laurenzlong: SGTM

firestore-counter/POSTINSTALL.md
laurenzlong: Yes that sounds good. 

Also `${function:controllerPubSub}` should become `${function:controllerCore}`

firestore-counter/extension.yaml
laurenzlong: SGTM

firestore-counter/POSTINSTALL.md
```diff
@@ -71,7 +80,7 @@ gcloud scheduler jobs create http firestore-sharded-counter-controller --schedul
 
 ### Using the extension
 
-After you complete the post-installation configuration above, your extension will create subcollections in all the documents that your app uses as counters. The client SDK will write to these subcollections to distribute the write load, and the scheduled function you deployed will sum the subcollections' values into the single `visits` field (or whichever field you configured).
+After you complete the post-installation configuration above, your extension will create subcollections in all the documents that your app uses as counters. The client SDK will write to these subcollections to distribute the write load, and the controllerCore function will sum the subcollections' values into the single `visits` field (or whichever field you configured).
```
rachelsaunders: @laurenzlong PTAL this change?

firestore-counter/extension.yaml
rachelsaunders: updated.  thx!

firestore-counter/PREINSTALL.md
rachelsaunders: updated.  thx!

firestore-counter/extension.yaml
rachelsaunders: updated.  thx!

firestore-counter/extension.yaml
rachelsaunders: updated.  thx!

firestore-counter/POSTINSTALL.md
rachelsaunders: updated.  thx!

firestore-counter/POSTINSTALL.md
laurenzlong: LGTM

firestore-counter/POSTINSTALL.md
```diff
@@ -18,17 +18,26 @@ match /databases/{database}/documents/pages/{page} {
...
 ```
 
-#### Set up a scheduled function
 
-Review the [scheduled function documentation](https://firebase.google.com/docs/functions/schedule-functions) to set up a call to `${function:controller.url}` every minute. You may need to enable some APIs in your Firebase project to use scheduled functions.
+#### Set up a Cloud Scheduler job
 
-As an example, to set up a scheduled function, you can run the following [`gcloud`](https://cloud.google.com/sdk/gcloud/) commands:
+**IMPORTANT:** Note the following about v0.1.1 of this extension:
+- **If you updated your extension from v0.1.0 to v0.1.1:**  We recommend that you edit your Cloud Scheduler job to instead send a message to the extension's Pub/Sub topic, as described in this section. Although it's not recommended, if you leave your Cloud Scheduler job calling `${function:controller.url}`, your extension will continue to run as expected. For more information about the changes for v0.1.1, refer to the [changelog](https://github.com/firebase/extensions/blob/next/firestore-counter/CHANGELOG.md).
```
laurenzlong: https://github.com/firebase/extensions/blob/master/firestore-counter/CHANGELOG.md

Should point to master,not next branch.

firestore-counter/POSTINSTALL.md
```diff
@@ -18,17 +18,26 @@ match /databases/{database}/documents/pages/{page} {
...
+- **If you installed this extension for the first time at v0.1.1:** Follow the instructions as described in this section.
+
```
laurenzlong: did you mean to add 2 newlines?

firestore-counter/POSTINSTALL.md
```diff
@@ -18,17 +18,26 @@ match /databases/{database}/documents/pages/{page} {
...
+
+Set up a [Cloud Scheduler job](https://cloud.google.com/scheduler/docs/quickstart) to regularly send a message to the extension's Pub/Sub topic (`${param:EXT_INSTANCE_ID}`). This Pub/Sub topic then automatically triggers the controllerCore function (`${function:controllerCore}`). This controllerCore function is created by the extension and monitors the extension's workload.
```
laurenzlong: I'm not sure if "monitors the extension's workload" is the best way to describe it. It contains a part of the business logic of the extension, so it's more like it's carrying out the extension's workload rather than monitoring it.

firestore-counter/POSTINSTALL.md
rachelsaunders: good catch!

firestore-counter/POSTINSTALL.md
rachelsaunders: good catch!

firestore-counter/POSTINSTALL.md
rachelsaunders: This is unchanged from the launch of the extension.  Is the functionality of this controllerCore function different from the functionality of the controller function?  We landed on this phrasing with @lepatryk .  Happy to update if y'all agree that it should be updated (we'd do it in several places throughout several files).

lepatryk: I'm fine with changing the wording here to be more precise. In the description of the function, we say: "This function either aggregates shards itself, or it schedules and monitors workers to aggregate shards"

rachelsaunders: I'm fine with using the same phrasing throughout each place this is mentioned. ok with y'all?

- extension.yaml, description of the function
- preinstall file, additional setup section
- postinstall file, cloud scheduler job section

firestore-counter/PREINSTALL.md
rachelsaunders: this has been updated

firestore-counter/functions/src/index.ts
rachelsaunders: I believe that this has been updated to "controllerCore"

firestore-counter/POSTINSTALL.md
laurenzlong: SGTM

rachelsaunders: done :-D

PR #50, opened 2019-10-16
[storage-resize-images] Update README.md with fixed link.

laurenzlong:


rachelsaunders:
Curious why this update wasn't made with the regeneration of all READMEs before the registry was updated?  Or is this the PR that regenerates/updates the READMEs?

laurenzlong:
I just updated the registry now, after merging this PR. We are aiming to have README get regenerated automatically either as a presubmit hook or as a part of our release script. I had to do it manually this time. Bear with us as our process is still in flux :)

laurenzlong:
Also READMEs only show up in Github, not on the website nor Console nor CLI, so it actually doesn't matter the ordering of when readme is updated vs when registry is updated.

PR #51, opened 2019-10-17
[storage-resize-images] move "how param works" link to IMG_SIZES parameter

rachelsaunders:
The parameter RESIZED_IMAGES_PATH does not need this link to the /products page; it should be on the IMG_SIZES param as this is the param that needs more in-depth explanation for proper configuration.


laurenzlong:
Trying out the new process, I'm merging approved PRs into the `next` branch.

PR #53, opened 2019-10-18
add note to FIRESTORE_DELETE_MODE about applicability

rachelsaunders:
This parameter is only applicable for those extensions that have a configured value for the FIRESTORE_PATHS parameter.

### DIFF
--- a/delete-user-data/extension.yaml
+++ b/delete-user-data/extension.yaml
@@ -106,8 +106,9 @@ params:
     type: select
     label: Cloud Firestore delete mode
     description: >-
-      How do you want to delete Cloud Firestore documents? To also delete documents in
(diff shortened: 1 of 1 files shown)


delete-user-data/extension.yaml
```diff
@@ -106,8 +106,9 @@ params:
     type: select
     label: Cloud Firestore delete mode
     description: >-
-      How do you want to delete Cloud Firestore documents? To also delete documents in
-      subcollections, set this parameter to `recursive`.
+      Only applicable if you use the `Cloud Firestore paths` parameter. How do you want
```
laurenzlong: Should the first sentence be inside parentheses?

rachelsaunders: sure!

Issue #48, opened 2019-10-16
Error on cold start firestore-send-email


rphlmr:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: `firestore-send-email`
- Configuration values (redact info where appropriate):
  - all config require (extension is working, except after a cold start)

#### Steps to reproduce:

firestore-send-email throw an unhandled error at cold start : 

```
Unhandled error occurred during processing: { Error
    at Http2CallStream.call.on (/srv/node_modules/@grpc/grpc-js/build/src/client.js:96:45)
    at emitOne (events.js:121:20)
    at Http2CallStream.emit (events.js:211:7)
    at process.nextTick (/srv/node_modules/@grpc/grpc-js/build/src/call-stream.js:71:22)
    at _combinedTickCallback (internal/process/next_tick.js:132:7)
    at process._tickDomainCallback (internal/process/next_tick.js:219:9)
  code: 13,
  details: '',
  metadata: Metadata { options: undefined, internalRepr: Map {} },
  note: 'Exception occurred in retry method that was not classified as transient' }
```

##### Expected result

Mail send / or send retry

##### Actual result

Mail not sent and not retried

Ehesp:
Thanks for the report! @laurenzlong would you be able to help take a look at this?

rphlmr:
I think it'll be linked to https://github.com/firebase/firebase-functions/issues/536

A problem in the grpc implementation.

In my own functions I have the same problem and I rollback dependencies to  :  

```
"firebase-admin": "7.4.0",
"firebase-functions": "2.3.1"
```

laurenzlong:
[...]

PR #65, opened 2019-10-22
Migrate to Firestore transactions for writes

Ehesp:
This PR introduces transaction based writes/deletes for extensions using Firestore, as discussed. 

Related PR: https://github.com/firebase/extensions/issues/48

### DIFF
--- a/delete-user-data/functions/src/index.ts
+++ b/delete-user-data/functions/src/index.ts
@@ -121,9 +121,13 @@ const clearFirestoreData = async (firestorePaths: string, uid: string) => {
       const isRecursive = config.firestoreDeleteMode === 'recursive';
 
       if (!isRecursive) {
+        const firestore = admin.firestore();
         logs.firestorePathDeleting(path, false);
-        await admin.firestore().runTransaction((async transaction => {
-          return transaction.delete(admin.firestore().doc(path));
+
+        // Wrapping in transaction to allow for automatic retries (#48)
+        await firestore.runTransaction((transaction => {
+          transaction.delete(firestore.doc(path));
+          return Promise.resolve();
         }));
         logs.firestorePathDeleted(path, false);
       } else {
--- a/delete-user-data/functions/src/index.ts
+++ b/delete-user-data/functions/src/index.ts
@@ -122,10 +122,9 @@ const clearFirestoreData = async (firestorePaths: string, uid: string) => {
 
       if (!isRecursive) {
         logs.firestorePathDeleting(path, false);
-        await admin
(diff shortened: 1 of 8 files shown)


delete-user-data/functions/src/index.ts
```diff
@@ -122,10 +122,9 @@ const clearFirestoreData = async (firestorePaths: string, uid: string) => {
 
       if (!isRecursive) {
         logs.firestorePathDeleting(path, false);
-        await admin
-          .firestore()
-          .doc(path)
-          .delete();
+        await admin.firestore().runTransaction((async transaction => {
```
schmidt-sebastian: Here and everywhere: Do you mind adding a comment on why you are doing this ("Wrapping in transaction to allow for automatic retries").

delete-user-data/functions/src/index.ts
```diff
@@ -122,10 +122,9 @@ const clearFirestoreData = async (firestorePaths: string, uid: string) => {
...
+          return transaction.delete(admin.firestore().doc(path));
```
schmidt-sebastian: Nit: Store `admin.firestore()` in a local variable.

firestore-send-email/functions/src/index.ts
```diff
@@ -86,14 +86,16 @@ function validateFieldArray(field: string, array?: string[]) {
...
 async function processCreate(snap: FirebaseFirestore.DocumentSnapshot) {
-  return snap.ref.update({
-    delivery: {
-      startTime: admin.firestore.FieldValue.serverTimestamp(),
-      state: "PENDING",
-      attempts: 0,
-      error: null,
-    },
-  });
+  return admin.firestore().runTransaction((async transaction => {
```
schmidt-sebastian: Note sure if you have a specific style that you are following, but in general it seems like your functions don't need to be marked async.

delete-user-data/functions/src/index.ts
Ehesp: Sure thing, will get them added.

firestore-send-email/functions/src/index.ts
Ehesp: This was to ensure the transaction returns a Promise. I'll change them to resolve a Promise from the callback instead.

Ehesp:
@schmidt-sebastian I've updated the code with the feedback. I'll go through and re-test these are all working with the latest CLI release too. We'll also need a version bump before merging.

laurenzlong:
I'll update the version numbers in a separate PR

PR #67, opened 2019-10-23
test(firestore-translate-text): Implement tests for translate extension

Salakar:
This PR adds full tests for full code coverage for the `firestore-translate-text` extension.

image

**Note**: The local patch file & `patch-package` are to workaround https://github.com/firebase/firebase-functions-test/pull/49 - will remove once available in a release.

---

**Notes**:

 - image
   - This test works on master - PR mentioned in code comment makes the test fail, can fix in another PR or this one - didn't want to change much other than adding tests
 - several minor unreachable code paths refactored or removed (see specific commits)


Salakar:
@rafikhan / @laurenzlong - would you be able to connect Circle CI or Travis CI to this repo? I can then look at creating the CI config as part of this PR if you'd like. [Coveralls](https://coveralls.io/) would be good too 😅

Ehesp:
@googlebot I consent.

Salakar:
We've updated to remove the local patch & `patch-package` now that a new version of `firebase-functions-test` is released with a fix.

firestore-translate-text/__tests__/config.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  });
+
+  afterEach(() => restoreEnv());
+
+  test("config loaded from environment variables", () => {
+    const functionsConfig = require("../functions/src/config").default;
+    expect(functionsConfig).toMatchSnapshot({});
+  });
+
+  test("config is logged on initialize", () => {
```
laurenzlong: This is fine to have, but I wouldn't worry too much about testing log behavior with all the extensions.

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+
+const testTranslations = {
+  de: "hallo",
+  en: "hello",
+  es: "hola",
+  fr: "salut",
+};
+
+const defaultEnvironment = {
+  PROJECT_ID: "extensions-testing",
```
laurenzlong: is this a real project?

.travis.yml
```diff
@@ -11,11 +15,10 @@ stages:
 
 jobs:
   include:
-    - stage: validate
-      name: "TypeScript Compile"
-      script: npm run clean && npm run build
```
laurenzlong: so when would TS compile happen now?

firestore-translate-text/__tests__/config.test.ts
Salakar: Just did this as a proof of concept mainly because the logging is not consistent across the extensions. 

I think we discussed a generic logging api/service that all extensions can use/inherit from but this is still pending/needs discussing through an API review.

Would you like me to remove log testing for now?

firestore-translate-text/__tests__/functions.test.ts
Salakar: It is a real project that we have - but in this instance it's just mock testing so can be any value here.

.travis.yml
Salakar: Currently already happens post-install on NPM from the root - so I removed this as it'd be doing it twice otherwise.

https://github.com/firebase/extensions/pull/86 would additionally add TS compile to pre-commit also to ensure compiled code is up to date before committing.

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      jest.resetModules();
+      mockTranslateClass.mockClear();
+
+      require("../functions/src");
+
+      expect(mockTranslateClass).toHaveBeenCalledTimes(1);
+      expect(mockTranslateClass).toHaveBeenCalledWith({
+        projectId: defaultEnvironment.PROJECT_ID,
+      });
+    });
```
rafikhan: I was confused by this test. On the surface I don't see what the relationship is between the require and the assertions.  Perhaps add some comments or change the string to describe the expected side effect?

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+
+      beforeSnapshot = functionsTest.firestore.makeDocumentSnapshot(
+        { input: "hello" },
+        "translations/id1"
+      );
+
+      afterSnapshot = functionsTest.firestore.makeDocumentSnapshot(
+        { input: "hello", changed: 123 },
+        "translations/id1"
+      );
```
rafikhan: See my comment from earlier about putting all programmatically generated mocks into their own file.

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      restoreEnv();
+      restoreEnv = mockedEnv({
+        ...defaultEnvironment,
+        INPUT_FIELD_NAME: "input",
+        OUTPUT_FIELD_NAME: "input",
+      });
+
+      wrappedFunction = functionsTest.wrap(
+        require("../functions/src").fstranslate
+      );
```
rafikhan: I haven't used jest before so perhaps this is an dumb question but this logic seems like its already being handled in the `beforeEach` calls earlier and I was confused as to why some of it is being repeated.  Why was this necessary?

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      restoreEnv();
+      restoreEnv = mockedEnv({
+        ...defaultEnvironment,
+        INPUT_FIELD_NAME: "output.en",
+        OUTPUT_FIELD_NAME: "output",
+      });
+
+      wrappedFunction = functionsTest.wrap(
+        require("../functions/src").fstranslate
+      );
```
rafikhan: Same comment as above.

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      expect(mockFirestoreUpdate).toHaveBeenCalledWith(
+        defaultEnvironment.OUTPUT_FIELD_NAME,
+        testTranslations
+      );
+
+      // confirm logs were printed
+      Object.keys(testTranslations).forEach((language) => {
+        // logs.translateInputString
+        expect(mockConsoleLog).toHaveBeenCalledWith(
+          `Translating string: 'hello' into language(s): '${language}'`
```
rafikhan: I really like that you're verifying the the logs were printed b/c we and our users will rely on this to diagnose issues. I'm concerned about the fragility of this though if we ever want to redo our messages.

**Random thought:** Perhaps refactor the logging to include a code in the front and pass in a structured object separately as part of `console.log`? The test assertions could just check the presence of the code and inspect structured data which will be more typesafe. I'm open to other ideas as well.

firestore-translate-text/__tests__/config.test.ts
laurenzlong: No need to remove. Thanks for the clarification.

.travis.yml
laurenzlong: Got it.

firestore-translate-text/__tests__/functions.test.ts
laurenzlong: can you revise to "fake-project" or something to make it obvious it is not using a real project?

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      exists: true,
+      get: documentSnapshot.get.bind(documentSnapshot),
+      ref: {
+        path: documentSnapshot.ref.path,
+      },
+    };
+  })();
+}
+
+jest.mock("@google-cloud/translate", mockTranslateModuleFactory);
```
rafikhan: This code is all tied to the mocks required for the tests.  What do you think about having a folder like `mocks` and putting this into something like a `translate.ts` file?

firestore-translate-text/__tests__/functions.test.ts
```diff
@@ -0,0 +1,384 @@
...
+      admin.firestore().runTransaction = jest.fn().mockImplementation(() => {
+        return (transactionHandler) => {
+          transactionHandler({
+            update(ref, field, data) {
+              mockFirestoreUpdate(field, data);
+            },
+          });
+        };
+      })();
+    });
```
rafikhan: Same as my comment from above. What about putting this into `mocks/firestore.ts` and use descriptive names to initialize these or just export the final values?

firestore-translate-text/__tests__/functions.test.ts
rafikhan: Agreed.

russellwheatley:
@googlebot I consent.

russellwheatley:
Hey @rafikhan ,
             I've been trying to address the concerns you raised, and I have the code up to a point where I would be grateful for your feedback.

Thus far:
-  Moved mocks into their own modules. Namely: `console`, `firestore` & `translate` as requested.
- Created a `jest-setup` module that runs before each test file.
- Created a separate `messages` module to ensure logs in the app and the tests are equivalent.
- I've added a few notes that ought to be a bit more informative about the test logic.
- Renamed `PROJECT_ID` to 'fake-project'

firestore-translate-text/__tests__/functions.test.ts
russellwheatley: As it is currently implemented,`beforeEach` runs generic environment setup logic. This is fine for most tests but in this instance, we wanted `INPUT_FIELD_NAME ` & `OUTPUT_FIELD_NAME ` to have the same value. So we set the environment variables again. If you see logic that looks duplicated, it's a safe bet that the test requires custom setup outside of the generic environment setup.

rafikhan: This makes sense. I think this is worth a comment or putting in a function w/ a descriptive name like `ensureInputOutputFieldsAreTheSame()` or something more concise.

rafikhan:
> Hey @rafikhan ,
> I've been trying to address the concerns you raised, and I have the code up to a point where I would be grateful for your feedback.

Hi @russellwheatley, I think the changes you've made thus far look good. I've commented on one small thing regarding the environment reset but otherwise I like the changes you made.

russellwheatley:
@rafikhan I've added a couple of code comments to make it clear what is happening around those env variable updates. I think this PR is good to go otherwise.

firestore-translate-text/__tests__/config.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  // LANGUAGES
+  describe("config.LANGUAGES", () => {
+    test("param exists", () => {
+      const extensionParam = extensionParams["LANGUAGES"];
+      expect(extensionParam).toMatchSnapshot();
+    });
+
+    test("removes duplicated any duplicated languages from user input", () => {
```
laurenzlong: Typo.

Salakar:
```suggestion
    test("removes any duplicated languages from user input", () => {
```

Issue #17, opened 2019-09-27
firestore-bigquery-export fs-bq-import-collection doesn't allow {widecard} path


levibn123:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: firestore-bigquery-export
- Configuration values (redact info where appropriate):
  - fs-bq-import-collection

#### Steps to reproduce:

1) Clone this repo.
2) Go to extensions/firestore-bigquery-export/scripts/import and install dependencies (npm install)
3) Execute `npm run import`
4) Pass the ${PROJECT_ID}
5) Try pass a widecard path as described on docs, like `chats/{chatid}/posts`
6) Get an error: `>> The collection path must only contain letters or spaces`
##### Expected result
Allow widecard paths to import.
##### Actual result
Get an error: `>> The collection path must only contain letters or spaces`

fredzqm:
Thanks for the feedback. This is not yet supported because we do not yet have a way to translate wild card path into a Firestore query. Wildcard are supported in triggers by default.

A low-hanging fruit might be adding support for collection group in the import script. Will that solve your use case?

levibn123:
Thanks. I realized this after looking at the code and the way imports were being made. I was able to work around this problem by using CollectionGroup and importing subcollections. For me that was extremely necessary.

skupsala:
> A low-hanging fruit might be adding support for collection group in the import script. Will that solve your use case?

That would be great!

jttaynton:
This is a great tool but definitely need a way to be able to do wildcards and collection groups would be good to.  I have a small dataset so for now I'm going to cheat and just run updates on my existing records.

MattReidArnold:
> Thanks. I realized this after looking at the code and the way imports were being made. I was able to work around this problem by using CollectionGroup and importing subcollections. For me that was extremely necessary.

[...]

PR #75, opened 2019-10-30
Add --non-interactive flag to views generator script

IanWyszynski:
This PR adds a `--no-interactive` flag to the schema view generator script so that developers may invoke the script with all necessary parameters specified as command line options. The mode is all-or-nothing: if `--no-interactive` is supplied, all parameters must be present or the script will fail. 

The `--schema-files` flag is repeatable and accepts globbed patterns such as `./schemas/*array*.txt`. Such patterns can point to a collection of schema definition files, or a collection of directories. If pointed to a directory, the script will assume that the directory contains exclusively schema definition files and process all of them. 

Schema names still share one namespace: if you have two different definition files in separate directories, if they have the same name, only one schema will get created.


firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,52 +43,80 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
+    name: "dataset",
     type: "input",
     validate: (value) =>
       validateInput(value, "dataset", BIGQUERY_VALID_CHARACTERS),
   },
   {
     message:
       "What is the name of the Cloud Firestore Collection that you would like to generate a schema view for?",
-    name: "collectionName",
+    name: "collection",
```
fredzqm: This is an existing issue, but worth mentioning.

It is not a good idea to couple Firestore `collectionName` with BigQuery `tableName`. For example, https://github.com/firebase/extensions/issues/17.
Let's rename this to `tableNamePrefix`.

We should describe it as the name prefix of BigQuery table generated by the BigQuery export extension.

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,52 +43,80 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
 ];
 
 async function run(): Promise<number> {
-  const schemaDirectory = [process.cwd(), "schemas"].join('/');
-  const schemaDirExists = existsSync(schemaDirectory);
-  if (!schemaDirExists) {
-    console.log(`Expected directory "${schemaDirectory}" not found!`);
-    process.exit(1);
-  }
+  program.parse(process.argv);
```
fredzqm: Extract a helper that returns an 
```
interface SchemaGenCliConfig {
 ...
}
```

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -112,17 +141,71 @@ async function run(): Promise<number> {
   return 0;
 };
 
-function readSchemas(directory: string): { [schemaName: string]: FirestoreSchema } {
+function readSchemas(globs: string[]): { [schemaName: string]: FirestoreSchema } {
```
fredzqm: These are fairly complicated logics. It would nice to have tests if not too hard.

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -112,17 +141,71 @@ async function run(): Promise<number> {
...
+        warnDuplicateSchemaName(schemaName);
+      }
+      schemas[schemaName] = readSchemaFromFile(dirent);
+    }
+  }
+  return schemas;
+}
+
+function warnDuplicateSchemaName(schemaName: string) {
+  console.log(`Found multiple schema files named ${schemaName}! Only the latest one will be created!`);
```
fredzqm:
```suggestion
  console.log(`Found multiple schema files named ${schemaName}! Only the last one will be created!`);
```
I don't think there is a time order associated with schema files?

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -112,17 +141,71 @@ async function run(): Promise<number> {
...
+    results = results.concat(globResults);
+  }
+  return results;
+}
+
+function readSchemasFromDirectory(directory: string): { [schemaName: string]: FirestoreSchema } {
   let results = {};
   let files = readdirSync(directory);
-  let schemaNames = files.map(fileName => path.basename(fileName).split('.').slice(0, -1).join('.').replace(/-/g,'_'));
+  let schemaNames = files.map(fileName => filePathToSchemaName(fileName));
```
fredzqm: Always use `const` unless they are assigned more than once.

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -112,17 +141,71 @@ async function run(): Promise<number> {
...
+}
+
+function warnDuplicateSchemaName(schemaName: string) {
+  console.log(`Found multiple schema files named ${schemaName}! Only the latest one will be created!`);
+
+}
+
+function resolveFilePath(filePath: string): string {
+  if (filePath.startsWith(".") || !filePath.startsWith("/")) {
+    filePath = [process.cwd(), filePath].join("/")
```
fredzqm:
```suggestion
    return [process.cwd(), filePath].join("/")
```

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
fredzqm: Optional nit:
Add `schemaName` to `FirestoreSchema` and return `FirestoreSchema[]` instead, which is slightly more nature to read.

You can still dedup `schemaName` internally with a separate set.

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,52 +43,80 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
 };
 
+function collect(value, previous) {
+  return previous.concat([value]);
+}
+
+program
+  .name("gen-schema-views")
+  .option("--no-interactive", "Parse all input from command line flags instead of prompting the caller.", false)
+  .option("-p, --project-id <project-id>", "Firebase Project ID for project containing Cloud Firestore database.")
```
fredzqm: Should those be consistent with names in the questions?

`project-id` vs `project`. I am happy with either.

mbleigh:
Please follow conventions of the Firebase CLI

-P --project
--non-interactive

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
IanWyszynski: I have them different now because the names that are specified in `questions` need to match the names of the variables that the input is captured in (https://github.com/firebase/extensions/pull/75/files#diff-fde3bc7b28048e80c1d9ed7c18fe1f47R109-R114). So if the names were consistent, then they would shadow the variables I'm actually passing to the schema generator class: https://github.com/firebase/extensions/pull/75/files#diff-fde3bc7b28048e80c1d9ed7c18fe1f47R93-R95.

What do you think of changing the names of the questions inputs to `<name>Input` (i.e. projectId => projectIdInput)?

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
IanWyszynski: I'd prefer to keep the `FirestoreSchema` interface as clean as possible (it should be possible to read in a json file directly into a valid FirestoreSchema object). Injecting a filesystem specific name feels like leaking information.

IMO part of the job of this script is to manage these names and I think they should probably be stored separately.

IanWyszynski:
The diff grew a bit because I ran a formatter over this code, but the main changes are:

- Extracting CLI parsing into a separate function
- Move utilities for loading schemas from the filesystem into their own package
- Add tests for glob pattern interpretation and other schema loading behavior

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -69,141 +82,99 @@ const questions = [
...
-      schemaFiles,
-    } = await inquirer.prompt(questions);
-    projectId = project;
-    datasetId = dataset;
-    collectionName = collection;
-    schemas = readSchemas(schemaFiles.split(",").map(schemaFileName => schemaFileName.trim()));
-  }
+async function run(): Promise<number> {
+  // Get all configuration options via inquirer prompt or commander flags.
+  const config: CliConfig = await getConfig();
```
fredzqm: nit: `parseConfig`

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -69,141 +82,99 @@ const questions = [
...
+    if (
+      program.project === undefined ||
+      program.dataset === undefined ||
+      program.tableNamePrefix === undefined ||
+      program.schemaFiles.length === 0
+    ) {
+      program.outputHelp();
+      process.exit(1);
+    }
+    config.projectId = program.project;
```
fredzqm: [General readability guideline] Avoid setting a variable if you don't have too. Immutability is always preferred.

In a helper method, you can return `CliConfig` from each branch directly without setting.

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,86 +38,143 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
-  if (Object.keys(schemas).length === 0) {
-    console.log(`Found no schemas in ${schemaDirectory}!`);
+  if (Object.keys(config.schemas).length === 0) {
+    console.log(`No schema files found!`);
   }
   const viewFactory = new FirestoreBigQuerySchemaViewFactory();
-
-  for (const schemaName in schemas) {
-    await viewFactory.initializeSchemaViewResources(datasetId, collectionName, schemaName, schemas[schemaName]);
+  for (let schemaName in config.schemas) {
```
fredzqm: const

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,86 +38,143 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
+    "-d, --dataset <dataset>",
+    "The ID of the BigQuery dataset containing a raw Cloud Firestore document changelog."
+  )
+  .option(
+    "-t, --table-name-prefix <table-name-prefix>",
+    "A common prefix for the names of all views generated by this script."
+  )
+  .option(
+    "-f, --schema-files <schema-files>",
+    "A path in the filesystem specifying a globbed collection of files to read schemas from.",
```
rachelsaunders: NIT:
... collection of files from which to read schemas.

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,86 +38,143 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
     type: "input",
     validate: (value) =>
-      validateInput(value, "dataset", BIGQUERY_VALID_CHARACTERS),
+      validateInput(value, "dataset ID", BIGQUERY_VALID_CHARACTERS),
   },
   {
     message:
       "What is the name of the Cloud Firestore Collection that you would like to generate a schema view for?",
-    name: "collectionName",
+    name: "tableNamePrefix",
```
rachelsaunders: NIT:
... name of the Cloud Firestore collection for which you want to generate a schema view?

firestore-bigquery-export/scripts/gen-schema-view/src/index.ts
```diff
@@ -41,86 +38,143 @@ const validateInput = (value: any, name: string, regex: RegExp) => {
...
-    confirmed
-  } = await inquirer.prompt(questions);
+async function run(): Promise<number> {
+  // Get all configuration options via inquirer prompt or commander flags.
+  const config: CliConfig = await getConfig();
 
   // Set project ID so it can be used in BigQuery intialization
-  process.env.PROJECT_ID = projectId;
+  process.env.PROJECT_ID = config.projectId;
   // BigQuery aactually requires this variable to set the project correctly.
```
rachelsaunders: aactually -> actually

firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
```diff
@@ -171,14 +192,18 @@ export const buildSchemaViewQuery = (
...
+      )
+      .join(" ")}`;
   }
   query = sqlFormatter.format(query);
   return query;
-}
+};
 
 /**
  * Given a firestore schema which may contain values for any type present
```
rachelsaunders: NIT:
firestore -> Cloud Firestore

firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
```diff
@@ -278,46 +318,87 @@ const processLeafField = (
...
+      const longitude = jsonExtract(
+        dataFieldName,
+        extractPrefix,
+        field,
+        `._longitude`,
+        transformer
+      );
+      /*
+       * We return directly from this branch because it's the only one that
+       * generate multiple selector clauses.
```
rachelsaunders: generate -> generates

firestore-bigquery-export/scripts/gen-schema-view/src/udf.ts
```diff
@@ -18,44 +18,59 @@ import * as sqlFormatter from "sql-formatter";
 
 // Persistent UDFS to be created on schema view initialization.
```
rachelsaunders: UDFS -> UDFs

firestore-bigquery-export/scripts/gen-schema-view/src/util.ts
```diff
@@ -0,0 +1,99 @@
...
+      }
+      schemas[schemaName] = readSchemaFromFile(dirent);
+    }
+  }
+  return schemas;
+}
+
+function warnDuplicateSchemaName(schemaName: string) {
+  console.log(
+    `Found multiple schema files named ${schemaName}! Only the last one will be created!`
```
rachelsaunders: Maybe change last sentence to:

Only the most recent one will be used to create the schema view!

IanWyszynski: The word "recent" implies that we are somehow iterating over these in a temporal order, but we aren't. @fredzqm left a comment about this requesting the opposite of what you're asking. I like adding "will be used to create the schema view" though.

IanWyszynski:
I force-pushed a change the removed the formatting bits from a previous commit and addressed the latest round of review feedback.

cc/ @fredzqm @rachelsaunders

firestore-bigquery-export/scripts/gen-schema-view/src/util.ts
```diff
@@ -0,0 +1,99 @@
+/*
```
fredzqm: Maybe name it more descriptive, like `schema-loader.ts`.

firestore-bigquery-export/scripts/gen-schema-view/src/util.ts
```diff
@@ -0,0 +1,99 @@
...
+import * as path from "path";
+
+import { FirestoreSchema } from "./schema";
+
+import { existsSync, readdirSync, lstatSync } from "fs";
+
+export function readSchemas(
+  globs: string[]
+): { [schemaName: string]: FirestoreSchema } {
+  let schemas = {};
```
fredzqm: `const` unless modified please.

PR #82, opened 2019-11-01
[fs-bq-export] add prettier formatting for fs-bq-schema-views script

IanWyszynski:
These are strictly formatting changes.


Thanhtai83:
Ok

PR #85, opened 2019-11-04
Update release notes URL and delete versions in package.json

laurenzlong:
- Revise release note url for all extensions to point to Github releases page
- Remove version number from all package.jsons, since they are not used for anything, and it is a burden to keep them in sync with the versions in extension.yaml
- Add "private:true" to package.json so that npm will not complain about the lack of a version number in package.json


rafikhan:
LGTM

PR #86, opened 2019-11-05
build(*): configure pre-commit hooks

Salakar:
This PR introduces pre-commit hooks (scripts that run before git committing locally) that do the following to all extensions on commit:

 - Clean and generate JS files via `npm run clean` & `npm run build`
 - Regenerate extension `README.md` via `firebase ext:info . --markdown > README.md`
 - Lint commit message format to ensure commit messages match the [conventional commit standard](https://www.conventionalcommits.org/en/v1.0.0/)
   - The tools we need to generate changelogs will require commits to follow this convention

----

This is blocked due to the following reasons;

- [x] `firebase ext:info . --markdown > README.md` requires a project to run the command, via `--project` or `firebase use` - this means the pre-commit scripts will fail without these, though I don't think a project is required to generate the README locally - can this restriction be removed?
- [ ] CLI issue on `delete-user-data` -> `firebase ext:info . --markdown > README.md` command fails with `Could not find existing source, checking for local source.` - not sure why, nothing on debug logs and checked yaml is valid. I have excluded the `delete-user-data` for now here: https://github.com/firebase/extensions/blob/9282f492b8c2da1ab011f626f343c0e230e98d22/package.json#L15
- [x] For this to work for everyone and to be merged, all contributors (external & internal) need to work from the repo root and not in the individual extensions, e.g. `npm install` at repo root, not inside individual packages, we're looking at updating the contributors guide for this, cc @ehesp who will send up a PR that updates the `CONTRIBUTING.md` file with local development guides etc


laurenzlong:
@joehan would you mind looking into why delete-user-data can't generate README?

Salakar:
> Looks great! Let's wait until the blocking issues are resolved before merging.

Thanks for the review, happy to wait

> After merging, how do we actually set it up to run as a precommit hook?

I've set it up so that it is automatically configured when you `npm install` in the root (via [`husky`](https://github.com/typicode/husky)) - no manual setup required (we can additionally use the same husky setup later on to validate that commit messages match an expected format for the change log generation).

joehan:
Just opened up a PR to fix your first blocker - https://github.com/FirebasePrivate/firebase-tools/pull/495

firebase.json
```diff
@@ -0,0 +1 @@
+{}
```
joehan: Do we need this file?

.firebaserc
```diff
@@ -0,0 +1 @@
+{}
```
joehan: Do we need this file?

firebase.json
Salakar: After your tools PR it won't be needed, this was added so I could do `firebase use` for the project requirement - which requires a valid Firebase project directory.

Will remove, thanks

.firebaserc
Salakar: After your tools PR it won't be needed, this was added so I could do `firebase use` for the project requirement - which requires a valid Firebase project directory.

Will remove, thanks

Salakar:
I've pushed up another change to this to add pre-commit message linting (updated PR description with info).

I've also been reconsidering the following pre-commit scripts:

 - Clean and generate JS files via `npm run clean` & `npm run build`
 - Regenerate extension `README.md` via `firebase ext:info . --markdown > README.md`

These are quite intensive take time to run, perhaps we should leave these out of pre-commit but keep commit messaging linting, but keep the scripts there for use in the release process?

laurenzlong:
@Salakar I think it makes a lot of sense to keep these out of precommit and keep them in release script since they are time intensive.

Ehesp:
@googlebot I consent.

Salakar:
I think this is now good to go, the blockers are a non-issue for now as these scripts are the readme script is not currently run on pre-commit anymore

PR #90, opened 2019-11-06
Bump versions and add release notes for Nov 7 release

laurenzlong:


delete-user-data/extension.yaml
```diff
@@ -15,7 +15,7 @@
 name: delete-user-data
 displayName: Delete User Data
 specVersion: v1beta
-version: 0.2.0
+version: 0.1.2
```
laurenzlong: Bumped down, because originally we thought this would be a breaking change, but we made it non-breaking by making the default of the param "shallow". So now it's just 0.1.1 > 0.1.2

storage-resize-images/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Version 0.1.2
+
+feature - Add new param for deleting the original image.
+fixed - Fixed "cold start" errors experienced when the Extension runs after a period of inactivity (issue #48).
```
rosariopfernandes: I don't think this was done on the `storage-resize-images` extension, since it doesn't use a firestore trigger.

laurenzlong: Great catch! That was meant for delete-user-data. Thanks!

rafikhan:
LGTM

PR #98, opened 2019-11-07
chore(*): Initial setup of CODEOWNERS file (automatic review assignments)

Salakar:
Small PR to setup an initial GitHub [`CODEOWNERS`](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) file.

Code owners are automatically requested for review when someone opens a pull request that modifies code that they own. Code owners are not automatically requested to review draft pull requests. 

I've added the following rules for now;

 - when any markdown (*.md) file is modified, request review from @rachelsaunders
 - when any `package.json` or `lerna.json` file is modified, request review from myself and @Ehesp 
 - when any of the new testing files/setup are modified, request review from myself and @Ehesp 

Hope these initial ones are ok? I can add more specific to each extension if required - let me know :)

PR #99, opened 2019-11-07
chore(*): update CODEOWNERS

Salakar:
Updated as per https://github.com/firebase/extensions/pull/98#issuecomment-551266401

cc @rachelsaunders

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,6 @@
 # All documentation changes, e.g. changelogs & readmes.
-*.md             @rachelsaunders @annie-ryan
-*/extension.yaml @rachelsaunders @annie-ryan
+*.md             @rachelsaunders
+*/extension.yaml @rachelsaunders
 
 # Dependencies, package.json and mono-repo files
 package.json     @Salakar @Ehesp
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,6 @@
 # All documentation changes, e.g. changelogs & readmes.
-*.md             @rachelsaunders
+*.md             @rachelsaunders @annie-ryan
```
laurenzlong: I'm not sure Annie needs to be in the CODEOWNERS list, she should no be expected to review things via Github. We will surface UX concerns during our Tuesday cross-functional syncs.

laurenzlong: Actually I just saw the comment @rachelsaunders left on the other PR. Rachel, let's discuss this with Annie offline.

rachelsaunders: I think that at minimum for this PR, we should have me be a "codeowner" for the the .md files and the extension.yaml files.  We can review the addition of Annie in the RACI review?

laurenzlong: That sounds good to me. @Salakar can you revise this PR to be just for adding Rachel as an owner of extension.yaml files, and we can revisit adding Annie in a subsequent PR if needed.

laurenzlong: Also, when I first wrote this comment, I accidentally edited Rachel's (that's scary that Github lets me edit other people's comments). I reverted Rachel's comment back to what it was before.

Issue #77, opened 2019-10-30
Send mails with custom headers with firestore-send-email extension


giammin:
### Extension name

firestore-send-email

### What feature would you like to see?

the ability to add custom header in a mail

### How would you use it?

We use Mandrill API to send emails and we can add some [custom metadata ](https://mandrill.zendesk.com/hc/en-us/articles/205582417-Using-Custom-Message-Metadata) to every message.

We can also add metadata sending email with Mandrill smtp but we need to pass them in the X-MC-Metadata header.

It would be great if we can specify in each message and at template level some headers to include in the message

Ehesp:
Thanks for the feature request. I'll pick this up internally & figure out how/if this can be integrated.

dasmeet:
Hi

Would love to have this feature. Am using SES but can't even enable basic SES open and click tracking without custom header.

karayu:
This looks like it's covered by #108

Issue #76, opened 2019-10-30
Firebase extension for transactional mails


tobiaseisenschenk:
### What feature would you like to see?

I see transactional e-mails as a common use case firebase users will implement. If this feature gets picked I'd be happy to contribute to its development. However I'd appreciate some guidelines as to how a function technically differs from an extension.

### How would you use it?

In my company we currently use [mailjet in combination with firebase functions](https://github.com/nui-os/mailjet-function). This is a minimal example to send welcome mails on authentication events, but other triggers could be useful as well.

giammin:
in my case it is enough [to implement email headers](https://github.com/firebase/extensions/issues/77) but having the ability to access the API would be really usefull

Ehesp:
Reference: https://github.com/firebase/extensions/issues/77

karayu:
Thanks so much for the feature request! I'm closing this for now since we're tracking in #108

PR #108, opened 2019-11-12
feat(firestore-send-email) Attach headers to nodemailer

Ehesp:
Fixes:
 - https://github.com/firebase/extensions/issues/77
 - https://github.com/firebase/extensions/issues/76 (part fix)

This PR allows users to attach additional headers to the mail.

### DIFF
--- a/firestore-send-email/POSTINSTALL.md
+++ b/firestore-send-email/POSTINSTALL.md
@@ -47,7 +47,7 @@ The top-level fields of the document supply the email sender and recipient infor
 * **ccUids:** An array containing the CC recipient UIDs.
 * **bcc:** A single recipient email address or an array containing multiple recipient email addresses.
 * **bccUids:** An array containing the BCC recipient UIDs.
-* **headers:** An object or array of additional header fields (e.g. `{"X-Key-Name": "key value"}` or `[{key: "X-Key-Name", value: "val1"}, {key: "X-Key-Name", value: "val2"}]`).
+* **headers:** An object of additional header fields (e.g. `{"X-Custom-Header": "value"}`).
 
 **NOTE:** The `toUids`, `ccUids`, and `bccUids` options deliver emails based on user UIDs keyed to email addresses within a Cloud Firestore document. To use these recipient options, you need to specify a Cloud Firestore collection for the extension's "Users collection" parameter. The extension can then read the `email` field for each UID specified in the `toUids`, `ccUids`, and/or `bccUids` fields.
 
--- a/firestore-send-email/POSTINSTALL.md
+++ b/firestore-send-email/POSTINSTALL.md
@@ -47,7 +47,7 @@ The top-level fields of the document supply the email sender and recipient infor
 * **ccUids:** An array containing the CC recipient UIDs.
 * **bcc:** A single recipient email address or an array containing multiple recipient email addresses.
 * **bccUids:** An array containing the BCC recipient UIDs.
-* **headers:** An object of additional header fields (for example, `{"X-Custom-Header": "value", "Y-Second-Custom-Header": "value"}`).
+* **headers:** An object of additional header fields (for example, `{"X-Custom-Header": "value", "X-Second-Custom-Header": "value"}`).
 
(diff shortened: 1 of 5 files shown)


firestore-send-email/functions/src/index.ts
```diff
@@ -73,6 +73,7 @@ interface QueuePayload {
   bccUids?: string[];
   from?: string;
   replyTo?: string;
+  headers?: any;
```
Ehesp: FYI: Taken from [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/nodemailer/v3/index.d.ts#L157). Headers are flexible so `any` works in this case.

jhuleatt: Since the postinstall says to provide an array or object, should we enforce that here?

firestore-send-email/POSTINSTALL.md
```diff
@@ -47,6 +47,7 @@ The top-level fields of the document supply the email sender and recipient infor
 * **ccUids:** An array containing the CC recipient UIDs.
 * **bcc:** A single recipient email address or an array containing multiple recipient email addresses.
 * **bccUids:** An array containing the BCC recipient UIDs.
+* **headers:** An object or array of additional header fields (e.g. `{"X-Key-Name": "key value"}` or `[{key: "X-Key-Name", value: "val1"}, {key: "X-Key-Name", value: "val2"}]`).
```
mbleigh: I don't think we want to mirror Nodemailer's messy and unintuitive header input. Let's just do an object with string keys and values unless you have a compelling reason to do otherwise...

firestore-send-email/functions/src/index.ts
rachelsaunders: Pls see mbleigh@'s comment below.  Is this comment thread here still applicable?

Ehesp: This is only for development, it won't have impact on the user experience. As NodeMailer accepts `any` I replicated it here. That being said, I'll change the docs to accept just a key/value object - I don't think this type should change?

firestore-send-email/POSTINSTALL.md
```diff
@@ -47,6 +47,7 @@ The top-level fields of the document supply the email sender and recipient infor
 * **ccUids:** An array containing the CC recipient UIDs.
 * **bcc:** A single recipient email address or an array containing multiple recipient email addresses.
 * **bccUids:** An array containing the BCC recipient UIDs.
+* **headers:** An object of additional header fields (for example, `{"X-Custom-Header": "value"}`).
```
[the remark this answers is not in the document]
Ehesp: @rachelsaunders to add multiple, the user would add a new object key/value pair, eg:

```js
{
  "X-Custom-Header": "value",
  "X-Second-Custom-Header": "value",
}
```

firestore-send-email/functions/src/index.ts
rachelsaunders: @mbleigh For confirmation pls - thx!

firestore-send-email/POSTINSTALL.md
rachelsaunders: @mbleigh Can this conversation be resolved now?

firestore-send-email/POSTINSTALL.md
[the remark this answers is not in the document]
rachelsaunders: added the 2nd header field to the example - PTAL?

firestore-send-email/POSTINSTALL.md
```diff
@@ -47,6 +47,7 @@ The top-level fields of the document supply the email sender and recipient infor
 * **ccUids:** An array containing the CC recipient UIDs.
 * **bcc:** A single recipient email address or an array containing multiple recipient email addresses.
 * **bccUids:** An array containing the BCC recipient UIDs.
+* **headers:** An object of additional header fields (for example, `{"X-Custom-Header": "value", "Y-Second-Custom-Header": "value"}`).
```
Ehesp: It'd be best to keep the second header as `"X-Second-Custom-Header"` - Prefixing a header with `X-` is usually practice for non-standard headers.

rachelsaunders: thx! done

firestore-send-email/POSTINSTALL.md
nwparker: This probably should have been kept to allow array entry

In emails, it's normal to have duplicate tags with multiple values

Issue #101, opened 2019-11-08
Export Collections to BigQuery creates identical entries in raw_changelog if multiple documents are created at the same time by a batch


jonathanrose9:
### Configuration
- Extension name: **Export Collections to BigQuery**

### The problem

Is it a problem to use the function in combination with batch operations and multiple documents in Firestore? Thank you very much for your help!

#### Steps to reproduce:
  - The extension is configured twice in the same Firebase project
  - We do not use wildcards in both configurations of the extension.
  - The two configurations of the extension each point to a sub-collection. (Collection path1: collection/doc1/subcollection; Collection path2: collection/doc2/subcollection)
  - The two configurations store the events in the same BigQuery dataset but in different tables within the dataset.
  - In our case, there are two ways in which documents are created in the subcollections: 
    - Documents in the sub-collections can be created separately by batch. The command batch.set() is used for this, because in this case the parent document is also updated.
    - Multiple documents in the sub-collections can be created per batch. The command batch.set() is used for this. The number can reach the maximum of 500 available from Firestore (see: https://firebase.google.com/docs/firestore/manage-data/transactions#batched-writes).

##### Expected result
For each created document in the sub-collection, an event in BigQuery should be stored in the raw_changelog table. There should be no duplicate entries.

##### Actual result
Multiple documents are created by a batch process in Firestore in the subcollections (as described above). The background trigger of the function starts as desired for each new document and exports the new documents as an event with the operation: "Create" to BigQuery. When we query the latest view table we get the expected result and the current state of the documents in Firestore. If you look at the raw_changelog table in BigQuery, some of the create operation entries are doubled. For some entries that were previously loaded into Firestore by batch, document_name, timestamp, event_id, operation and data are identical. If the batch process is executed with a small number of documents ( <10 ), no duplicate entries are created in Firestore. If the batch process is executed with a high number (e.g.: 400), identical entries (max. 2 per document) are created for some documents.

rphlmr:
Hi,
We have 3 bigquery extensions running (each one on different collections, no subcollections) and we didn't notice duplicating data.
Maybe the duplication is due to retries made by your function executing batch (there is a limitation of 1s to write on the same document).
You said that your batch processing write on the parent document, for this use case I would use transaction instead of batch.
Transaction will stop and retry if the parent document changed since the transaction's start.

jonathanrose9:
Hello rphlmr,

[...]

PR #118, opened 2019-11-20
fix(firestore-bigquery-export): fix occasional duplicate entries in BigQuery

IanWyszynski:
https://github.com/firebase/extensions/issues/101 identifies the possibility that the raw change log may contain duplicate rows with matching eventIds. This is due to the at-least-once nature of cloud function execution. One potential safeguard against this is to use the function's eventId as a BigQuery streaming insert `insertId`. This will give BigQuery streaming works the chance to deduplicate redundant rows coming from repeated Cloud function executions occurring within a couple of minutes of one another [1]. This should probably cover most instances of this issue. 

I've stress tested mirroring around 500K document changes to a collection and compared the number of rows in the raw changelog to the number of distinct event_ids and found them to be equal. This indicates that the fix should at least mitigate the concern raised in the issue.

[1] https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataconsistency

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
@@ -51,29 +51,33 @@ export class FirestoreBigQueryEventHistoryTracker
 
   async record(events: FirestoreDocumentChangeEvent[]) {
     await this.initialize();
-
+    const options = {
+      raw: true
+    };
     const rows = events.map((event) => {
-      // This must match firestoreToBQTable().
       return {
-        timestamp: event.timestamp,
-        event_id: event.eventId,
-        document_name: event.documentName,
-        operation: ChangeType[event.operation],
-        data: JSON.stringify(event.data),
+        insertId: event.eventId,
+        json: {
+          timestamp: event.timestamp,
+          event_id: event.eventId,
+          document_name: event.documentName,
+          operation: ChangeType[event.operation],
+          data: JSON.stringify(event.data),
+        },
       };
     });
-    await this.insertData(rows);
+    await this.insertData(rows, options);
   }
 
   /**
    * Inserts rows of data into the BigQuery raw change log table.
    */
-  private async insertData(rows: bigquery.RowMetadata[]) {
+  private async insertData(rows: bigquery.RowMetadata[], options: object) {
     try {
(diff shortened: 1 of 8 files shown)


firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
```diff
@@ -51,29 +51,33 @@ export class FirestoreBigQueryEventHistoryTracker
...
     const rows = events.map((event) => {
-      // This must match firestoreToBQTable().
       return {
-        timestamp: event.timestamp,
-        event_id: event.eventId,
-        document_name: event.documentName,
-        operation: ChangeType[event.operation],
-        data: JSON.stringify(event.data),
+        insertId: event.eventId,
+        json: {
```
Ehesp: Will nesting this data not create a breaking change for users with existing data?

IanWyszynski: This doesn't change the content of the row that is mirrored in bigquery, it is just the required format for the streaming insert payload if we want to use insertIds (https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/insertAll). The format of the json payload changes so that we can specify a custom insertId, but that id is transparent to existing users.

Ehesp: Makes sense thanks for explaining

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
```diff
@@ -51,29 +51,38 @@ export class FirestoreBigQueryEventHistoryTracker
 
   async record(events: FirestoreDocumentChangeEvent[]) {
     await this.initialize();
-
+    const options = {
+      raw: true,
```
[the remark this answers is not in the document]
IanWyszynski: It tells the BigQuery streaming API to expect the payload we're sending instead of plain row data. https://googleapis.dev/nodejs/bigquery/latest/Table.html#insert.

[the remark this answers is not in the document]
laurenzlong: Got it, thanks!

PR #119, opened 2019-11-20
docs(firestore-bigquery-export) Capitalize "IMPORT" in example query

evenbrenna:
Using lowercase "import" will give zero results since the value of the `operation` field is capitalized in rows written by the script.


evenbrenna:
@googlebot I signed it!

PR #127, opened 2019-12-06
typo fix

jnlopar:
s/reqad/read

### DIFF
--- a/firestore-counter/functions/src/controller.ts
+++ b/firestore-counter/functions/src/controller.ts
@@ -240,7 +240,7 @@ export class ShardedCounterController {
         );
       } catch (err) {
         console.log("Failed to read worker docs.", err);
-        throw Error("Failed to reqad worker docs.");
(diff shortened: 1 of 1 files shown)


rosariopfernandes:
@jnlopar nice catch!

Maybe we should run `npm run build` before merge so that it also fixes the typo on the js file? 

https://github.com/firebase/extensions/blob/c044297c5281c1af44a3f40f84d0c6613e64564c/firestore-counter/functions/lib/controller.js#L204

jnlopar:
I don't currently have a dev setup for building the extensions lib, as I did the PR directly via github's web UI after finding it as a runtime error on my app. @rosariopfernandes  would you like me to clone the repo and build it locally or is it something that can be done quickly by someone who has it set up already?

laurenzlong:
@jnlopar Thanks for the fix! I'll merge it and run the build on the `next` branch so the JS files are updated also. Good catch @rosariopfernandes !

Issue #130, opened 2019-12-10
Error when resizing image TypeError: Cannot set property 'resizedImage' of undefined


Timtech4u:
Hello, I installed the Resize Image Extension but it breaks with the following error:
```
Error when resizing image TypeError: Cannot set property 'resizedImage' of undefined
```

Screenshot:
![Screenshot from 2019-12-10 09-50-06](https://user-images.githubusercontent.com/11256481/70513731-772d6580-1b32-11ea-9b0d-927c2b7dc476.png)

rosariopfernandes:
@Timtech4u Thanks for filing this issue. But we might need some more information in order to be able to reproduce the problem:
- Which version of the storage-resize-images extension are you using?
- What configuration parameters did you use to setup the extension?

Timtech4u:
I used the default parameters and the Firebase UI.

It works now when I switch using the CLI command.

`firebase ext:install storage-resize-images --project=projectId`

utsav1322:
Hello, I'm Facing Same Issue
Error when resizing image TypeError: Cannot set property 'resizedImage' of undefined
&
Error when deleting temporary files { Error: ENOENT: no such file or directory, unlink

storage-resize-images extension version 0.1.3
configuration parameters 
-Deployment location 
us-central1
-Sizes of resized images
200*200,500*500,1240*1240
-Deletion of original file
false
-other's are default unset

I have upload it from sdk 
    firebase": "6.3.0",
    "firesql": "2.0.2",
    "linq2fire": "1.0.22",
    "react": "16.8.3",
    "react-native": "0.59.9",
I try installing extension from both way console and cli none of them work for me


![1Screenshot from 2019-12-11 13-21-48](https://user-images.githubusercontent.com/31151641/70602505-9b647180-1c1a-11ea-9388-f1173ce12fb6.png)

utsav1322:
yet its not resizing its getting undefined on line 140 of index.js in lib

Cybergebi:
I setup the Extension for the First Time yesterday and i get exactly the same Error

laurenzlong:
@Ehesp can you try to repro?

Ehesp:
I'm struggling to reproduce this. I've tried an existing instance and a totally fresh one but I can't trigger this error.

Could I ask how you are uploading an image?

[...]

PR #134, opened 2019-12-12
fix(storage-resize-images): fix bug where name of resized file was missing original name if there was no file extension

yuichkun:
I'm not sure if it is the expected behavior or not, and if so, please disregard this PR.

## Issue
As someone has pointed out in [this comment](https://github.com/firebase/extensions/issues/130#issuecomment-564425978), when a user uploads an image to the storage with a ref name **without** an extension, resized image's name gets to be trimmed like `_200x200`.

Below is a sample code that reproduces this.

```typescript
const storageRef = firebase.storage().ref('RANDOM_PATH')
const folderRef = storageRef.child('my_beautiful_cat')
await folderRef.put(e.target.files[0]) // got this from File API
```

Screenshot 2019-12-13 02 44 02

## Root Cause

It seems like the problem is happening where it [extracts the file name without extension](https://github.com/firebase/extensions/blob/next/storage-resize-images/functions/src/index.ts#L66). 
The current implementation is based on the premise that the ref would always have an extension, in other words, `fileExtension.length` is never 0. But, as it is not a restriction that firebase SDK enforces, when the file ref does not contain an extension, it should simply use the ref name, I guess.

## Fix

Node.js's `path.basename` can extract filename without extension.

reference: https://nodejs.org/docs/latest-v8.x/api/path.html#path_path_basename_path_ext

### DIFF
--- a/storage-resize-images/__tests__/extractFileNameWithoutExtension.ts
+++ b/storage-resize-images/__tests__/extractFileNameWithoutExtension.ts
@@ -1,0 +1,26 @@
+import { extractFileNameWithoutExtension } from "../functions/src/util";
+describe("extractFileNameWithoutExtension", () => {
+  const filePathWithExtension = "/ref/to/my/image.png";
+  const filePathWithoutExtension = "/ref/to/my/image";
+  const ext = ".png";
+
+  describe("when file path includes file extension", () => {
+    it("extracts the basename without extension", () => {
+      const fileNameWithoutExtension = extractFileNameWithoutExtension(
+        filePathWithExtension,
+        ext
+      );
+      expect(fileNameWithoutExtension).toBe("image");
+    });
+  });
+
+  describe("when file path does not include file extension", () => {
+    it("extracts the basename without extension", () => {
+      const fileNameWithoutExtension = extractFileNameWithoutExtension(
+        filePathWithoutExtension,
+        ext
+      );
+      expect(fileNameWithoutExtension).toBe("image");
+    });
+  });
+});
--- a/storage-resize-images/functions/lib/index.js
+++ b/storage-resize-images/functions/lib/index.js
@@ -33,6 +33,7 @@ const path = require("path");
 const config_1 = require("./config");
 const logs = require("./logs");
 const validators = require("./validators");
+const util_1 = require("./util");
 // Initialize the Firebase Admin SDK
 admin.initializeApp();
 logs.init();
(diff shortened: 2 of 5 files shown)


yuichkun:
#135

Ehesp:
Currently, if a user uploads an image called `image.jpg`, it would get resized to say `image_200x200.jpg`, as expected.

With this change, although it would correctly handle images without an extension, wouldn't it always output an image with no extension name, which is different than the current implementation? For example:

`image.jpg` -> `image_200x200.jpg`
vs
`image.jpg` -> `image_200x200`

My thinking is if the user uploads an image with an extension, we should persist it when resizing, if not we don't add it. What do you think?

Ehesp:
Also I think issue #130 is unrelated to this.

yuichkun:
Hi @Ehesp , thank you for your review.

`My thinking is if the user uploads an image with an extension, we should persist it when resizing, if not we don't add it.`  
I completely agree with you, and I believe this PR does abide by that!

As far as I understand, `fileNameWithoutExtension`, which literally refers to the file name **without** an extension,  is passed on to the `resizeImage` function along with `fileExtension`.

https://github.com/firebase/extensions/blob/947d533eb4531dc311b10f8a94e1c60fde9e12cc/storage-resize-images/functions/src/index.ts#L92-L105

The `resizeImage` function then reconstructs the output file name like below,

https://github.com/firebase/extensions/blob/947d533eb4531dc311b10f8a94e1c60fde9e12cc/storage-resize-images/functions/src/index.ts#L158

So, when the image does not have a file extension, in other words `fileExtension` is `''`,  `resizedFileName` does **not** have a file extension but the **original file name** with **size suffix**, and when the image does have a file extension, it **preserves** the file extension.

I haven't written unit tests for above since it kind of exceeds the scope of this topic, but if it's desirable, I guess I can provide one. 

Btw, are there anyways I can actually use my local version of this extension with an actual Firebase project? I couldn't find a way to confirm that everything works in harmony 🤔

yuichkun:
Also, my apologies for mentioning an unrelated issue! Edited the description.

Ehesp:
Ah yes you are correct, thanks for explaining. I also didn't realise that a `.slice` with a `-0` value totally removes the string (`"foo".slice(0, -0)` = `""`).

Let me give it a go and I'll get back to you.

Ehesp:
Tested this out and it's working:

image

Funnily enough, I added an image through gsutil & then realised I hit issue #130. So although this works, https://github.com/firebase/extensions/pull/142 needs merging too :)

Issue #148, opened 2019-12-30
Can't Delete Directory of Files


ccjernigan:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

### [REQUIRED] Step 2: Extension name

This feature request is for extension: delete-user-data

### What feature would you like to see?

What I currently see is that deleting files from Google Cloud Storage requires an explicit path for a single file.  This works if all users have filenames that follow the exact same naming rules.  E.g. {DEFAULT}/{UID}-pic.png

What I want to see is the ability to delete everything under a particular path.  E.g. {DEFAULT}/{UID}/*

### How would you use it?

I store user files under a path like:
{DEFAULT}/{UID}/

Under that path, there are an arbitrary number of files with randomly generated UUIDs.  When the user delete's their account, I'd like to be able to purge all of these files automatically.  With the current version of the extension, I am not able to delete these files because I don't know their names ahead of time.

Ehesp:
I *think* this should be possible... There isn't a way to directly delete a directory in Firebase/Google storage, however it looks as though `deleteFiles` accepts a `prefix` parameter.

I'll give this a go and see if it works with a directory & file.

PR #152, opened 2019-12-31
feat(delete-user-data): allow deleting directories & files

Ehesp:
Implements #148

`bucket.file(x)` only allows deleting specific files, not directories. This change uses `deleteFiles` with a prefix arg to delete both files and directories.

```
-- my-bucket
  -- 123
    -- foo.jpg
    -- bar.jpg
  -- 567.png
```

Given the structure above, the user could enter the following into the setup args to delete both `foo.jpg` & `bar.jpg` files:

`my-bucket/{UID}` (UID = 123)

To delete a single file, they'd do the usual:

`my-bucket/{UID}.png` (UID = 567)

### DIFF
--- a/delete-user-data/extension.yaml
+++ b/delete-user-data/extension.yaml
@@ -144,14 +144,14 @@ params:
       Where in Google Cloud Storage do you store user data? Leave empty if you
       don't use Cloud Storage.
 
-      Enter the full paths to a file or directory, separated by commas. You can
-      represent the User ID of the deleted user with `{UID}`. You can use
-      `{DEFAULT}` to represent your default bucket.
+      Enter the full paths to files or directories in your Storage buckets,
+      separated by commas. Use `{UID}` to represent the User ID of the deleted
+      user, and use `{DEFAULT}` to represent your default Storage bucket.
 
-      For example, if you are using your default bucket, and the bucket has
-      files with the naming scheme `{UID}-pic.png`, then you can enter
-      `{DEFAULT}/{UID}-pic.png`. If you also have files in another bucket called
-      `my-awesome-app-logs`, and that bucket has files with the naming scheme
-      `{UID}-logs.txt`, then you can enter
-      `{DEFAULT}/{UID}-pic.png,my-awesome-app-logs/{UID}-logs.txt`. You can
-      delete a directory by proving the full path, for example `{DEFAULT}/media/{UID}`.
+      Here's a series of examples. To delete all the files in your default
+      bucket with the file naming scheme `{UID}-pic.png`, enter
+      `{DEFAULT}/{UID}-pic.png`. To also delete all the files in another bucket
+      called my-app-logs with the file naming scheme `{UID}-logs.txt`, enter
+      `{DEFAULT}/{UID}-pic.png,my-app-logs/{UID}-logs.txt`. To *also* delete a User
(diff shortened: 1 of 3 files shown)


delete-user-data/extension.yaml
```diff
@@ -18,8 +18,8 @@ specVersion: v1beta
 version: 0.1.2
 
 description:
-  Deletes data keyed on a userId from Cloud Firestore, Realtime
-  Database, and/or Cloud Storage when a user deletes their account.
```
rachelsaunders: jFYI - the console and the CLI do not respect line breaks in these parameter descriptions.  Given this, I wouldn't spend a lot of time adjusting these line breaks in the extension.yaml file, unless it's just for readability in the extension.yaml file (which is, of course, a nice-to-have).

delete-user-data/extension.yaml
```diff
@@ -140,12 +144,14 @@ params:
...
-      Enter the full paths, separated by commas. You can represent the User ID of the deleted user with `{UID}`.
-      You can use `{DEFAULT}` to represent your default bucket.
-
-      For example, if you are using your default bucket,
-      and the bucket has files with the naming scheme `{UID}-pic.png`,
-      then you can enter `{DEFAULT}/{UID}-pic.png`.
-      If you also have files in another bucket called `my-awesome-app-logs`,
-      and that bucket has files with the naming scheme `{UID}-logs.txt`,
-      then you can enter `{DEFAULT}/{UID}-pic.png,my-awesome-app-logs/{UID}-logs.txt`.
+      Enter the full paths to a file or directory, separated by commas. You can
```
rachelsaunders: Let's update these paragraphs a bit.  How 'bout:

Enter the full paths to files or directories in a Storage bucket, separated by commas. Use `{UID}` to represent the User ID of the deleted user, and use `{DEFAULT}` to represent your default Storage bucket.

Here's a series of examples.  To delete all the files in your default bucket with the file naming scheme `{UID}-pic.png`, enter `{DEFAULT}/{UID}-pic.png`. To _also_ delete all the files in another bucket called `my-app-logs` with the file naming scheme `{UID}-logs.txt`, enter `{DEFAULT}/{UID}-pic.png,my-app-logs/{UID}-logs.txt`. To _also_ delete a User ID-labeled directory and all its files (like `media/{UID}`), enter `{DEFAULT}/{UID}-pic.png,my-app-logs/{UID}-logs.txt,{DEFAULT}/media/{UID}`.

delete-user-data/extension.yaml
Ehesp: Looks like these were changed because of formatting (the line width rule). I can change back but if we introduce CI checks, this may need to be reverted to be linted correctly.

delete-user-data/extension.yaml
Ehesp: Sounds good to me, will update.

delete-user-data/extension.yaml
rachelsaunders: naw, I'd just leave the changes that are already in this PR.

If we do eventually get control of how these descriptions line break in the console and CLI, we'll need to be sure that any checks here in GitHub play nice :-)

delete-user-data/extension.yaml
rachelsaunders: LGTM!  thx for the quick update!

Issue #27, opened 2019-10-07
Resize Images - Firebase Storage Bucket Names Invalid


paulrenshaw-tls:
Expected behaviour:
Able to enter the name of firebase storage buckets in the extension configuration as they appear in the firebase console.

Actual behaviour:
'Invalid storage bucket' warning in extension configuration for names of storage buckets created in the firebase console. 'Install extension' button is not in a disabled state but does not respond in this scenario.

Further notes:
The firebase-admin sdk works perfectly well with storage names as they are entered when creating a new bucket in the console / displayed in the firebase console. It seems odd that you would restrict what can be entered in the extension configuration with validation rules different to whatever is used for creating buckets in the first place.

rosariopfernandes:
Sounds like a dupe of #18

paulrenshaw-tls:
> Sounds like a dupe of #18

#18 talks about defining a 'custom' bucket, implying something non-standard. The source for the extension has a regex filter for the bucket name requiring a string of text with a least 2 .s with strings around them i.e. 
```
 - param: IMG_BUCKET
    type: string
    label: Cloud Storage bucket for images
    description: >
      To which Cloud Storage bucket will you upload images that you want to resize?
      This bucket will store both the original and resized images.
    default: ${STORAGE_BUCKET}
    validationRegex: (.)+\.(.)+\.(.)+
    validationErrorMessage: Invalid storage bucket
    example: my-project-12345.appspot.com
    required: true
```
which is what is restricting entered values to something like the example value, but also allows anything else with 2 .s in, like a bucket URL such as in the example in the response by @Salakar or something like an IP address or even 'this.is.silly'

#18 talks about the issue that when you enter something incorrectly, it doesn't work. That's not an issue. Their expected result is 'It should use the defined bucket', but then they are giving an example not using a defined bucket and showing that that doesn't work - this doesn't demonstrate what the actual issue is, though it is a good demonstration of the wasted time it causes people to spend by not accepting valid entries.

It should not be expected that it uses whatever you define (since that might not be a bucket, as they demonstrate entering a value which is not a bucket), it should be expected that it accepts all and only valid bucket names, not just any value in the form [string].[string].[string]

My issue is that you can't enter something which would be correct.

[...]

Issue #18, opened 2019-09-28
[BUG] Define custom bucket for extention Resize Image


digimbyte:
### [REQUIRED] Step 2: Describe your configuration
Resize Image, default settings
set Bucket to name of any other bucket that is not your default

Deployment location Description
us-central1
Cloud Storage bucket for images Description
new-second-bucket 
Sizes of resized images Description
128x128
Cloud Storage path for resized images (Optional) Description
Parameter not set
Cache-Control header for resized images (Optional) Description
Parameter not set

### [REQUIRED] Step 3: Describe the problem
Cloud Storage Bucket is not clear on the naming convention
should it be "new-second-bucket" per "gs://new-second-bucket"
or <bucket\>.<project\>.appspot.com/
or something else?
needs to be clearer on how to define your custom bucket
#### Steps to reproduce:

Install new extention "Resize Image"
configure the bucket to match the url of the second or alternative bucket
ie "gs://new-second-bucket" to "new-second-bucket"
or other respectable forms

##### Expected result
it should use the defined bucket
##### Actual result
Error occurred while deploying resources: RESOURCE_ERROR: `/deployments/firebase-ext-storage-resize-images/resources/generateResizedImage` {"ResourceType":"gcp-types/cloudfunctions-v1:projects.locations.functions","ResourceErrorCode":"400","ResourceErrorMessage":{"code":400,"message":"The request has errors","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.BadRequest","fieldViolations":[{"field":"event_trigger.resource","description":"GCS bucket resource projects/_/buckets/https://console.cloud.google.com/storage/browser/new-second-bucket does not match the expected pattern, which is projects/{project}/buckets/{bucket}"}]}],"statusMessage":"Bad Request","requestPath":"https://cloudfunctions.googleapis.com/v1/projects/worldsbeyond-1/locations/us-central1/functions/ext-storage-resize-images-generateResizedImage","httpMethod":"PATCH"}}There was an unknown error while processing the request. Try again.

Salakar:
Hey @digimbyte 👋 I believe the expected format is `my-project-or-bucket-12345.appspot.com` based on the example in the [extension configuration yaml](https://github.com/firebase/extensions/blob/master/storage-resize-images/extension.yaml#L101) 

The example should be showing in the UI as below: 
[...]

PR #153, opened 2019-12-31
fix(storage-resize-images) relax bucket naming rules

Ehesp:
Handles #27 & allows for #18

Currently the rules assume the bucket name is always `x.x.x`, whereas it can actually be anything with letters, numbers, dashes, dots, underscores.

This only checks for characters, it doesn't validate the storage bucket actually exists.


rachelsaunders:
@laurenzlong (or someone she delegates) should be the reviewer on this change.  Thx!

Issue #40, opened 2019-10-13
Attach file in email when using trigger email extension.


edlgg:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

### [REQUIRED] Step 2: Extension name

This feature request is for extension: firestore-send-email

### What feature would you like to see?

If would like to be able to attach files when I use the send email extension.

### How would you use it?

I want to send my users some files when they request them.

Ehesp:
Thanks we'll raise this internally as a feature request and let you know if/when it will be implemented.

BluebambooSRL:
@Ehesp how you doing?
Is there any deadline for the delivery of this feature-request?
We are developing a "message & alerts" tier, and wonder to know if this firebase extension will be a good solution. Thanks in advance, Federico

Ehesp:
Hey @BluebambooSRL this is now being reviewed internally. Thanks for your patience 👍

bgwissam:
any update on this issue? have the extension been created?

russellwheatley:
Hey @bgwissam, this issue was closed because we updated the documentation for attaching file. Here's the [PR](https://github.com/firebase/extensions/pull/158/files) if you want to take a look.

PR #158, opened 2020-01-09
docs(firestore-send-email): add instructions for email attachments

russellwheatley:
Implements #40.

This PR updates the POSTINSTALL.md and shows how a file may be attached to an email.

### DIFF
--- a/firestore-send-email/POSTINSTALL.md
+++ b/firestore-send-email/POSTINSTALL.md
@@ -8,7 +8,7 @@ You can test out this extension right away:
 
 1.  Add a document with a `to` field and a `message` field with the following content:
 
-    ```
+    ```js
     to: ['someone@example.com'],
     message: {
       subject: 'Hello from Firebase!',
@@ -20,7 +20,7 @@ You can test out this extension right away:
 1.  In a few seconds, you'll see a `delivery` field appear in the document. The field will update as the extension processes the email.
 
 **Note:** You can also use the [Firebase Admin SDK][admin_sdk] to add a document:
-```
+```js
 admin.firestore().collection('${param:MAIL_COLLECTION}').add({
   to: 'someone@example.com',
   message: {
@@ -62,6 +62,30 @@ Available properties for the `message` field are:
 * **text:** The plaintext content of the email.
 * **html:** The HTML content of the email.
 * **amp:** The [AMP4EMAIL][amp4email] content of the email.
+* **attachments:** An array containing attachment(s) as noted below.
+
+To send email attachments, you have to include an `attachments` array to the `message` map: 
+
+*Please note that the* `pathname` *property requires a URL path to the file you wish to attach.*
+*For further options, please consult [nodemailer attachment options](https://nodemailer.com/message/attachments/), bearing in mind you have no access to the file system.*
+
+```js
+admin.firestore().collection('mail').add({
+  to: 'someone@example.com',
+  message: {
+    subject: 'Hello from Firebase!',
+    attachments:[
+      {
+        path: 'https://url-path-to-file/cat.png'
+      },
+      {
+        filename: 'newFileName.txt',
+        path: 'https://url-path-to-file/file.txt'
+      },
+    ]
+  },
+})
+```
 
 #### Using templates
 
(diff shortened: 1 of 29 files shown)


russellwheatley:
@laurenzlong Apologies, issue link is now in the PR description.

@rachelsaunders Thanks for letting me know! I've since moved it to the POSTINSTALL.md.

firestore-send-email/POSTINSTALL.md
```diff
@@ -62,6 +62,30 @@ Available properties for the `message` field are:
 * **text:** The plaintext content of the email.
 * **html:** The HTML content of the email.
 * **amp:** The [AMP4EMAIL][amp4email] content of the email.
+* **attachments:** An array containing attachment(s) as noted below.
+
+To send email attachments, you have to include an `attachments` array to the `message` map: 
+
+*Please note that the* `pathname` *property requires a URL path to the file you wish to attach.*
+*For further options, please consult [nodemailer attachment options](https://nodemailer.com/message/attachments/), bearing in mind you have no access to the file system.*
```
rachelsaunders: Are all the options in this reference supported? https://nodemailer.com/message/attachments/

If so, I'd like to just point devs to that reference rather than add an example specifically for attachments.  Perhaps in a separate PR, we can add a more comprehensive example with all the options available.

I propose the entirety of the change for this PR be something like this instead:

**attachments:** An array containing attachment(s); supports [Nodemailer options](https://nodemailer.com/message/attachments/) (be aware that your email has no access to the recipient's file system)

russellwheatley: Not all those options are available for the user I'm afraid. I've updated the code to reflect your idea. I've also tweaked your suggestion from `recipient's file system` to `cloud server's file system`. I hope you don't mind!

rachelsaunders: ah!  good to know!  Which ones _are_ supported?

russellwheatley: The following attachments are supported : "utf-8 string as an attachment", "define custom content type for the attachment", "use URL as an attachment", "encoded string as an attachment", "data uri as an attachment", "pregenerated MIME node". I've taken these from the code comments from here: https://nodemailer.com/message/attachments/ so you can easily reference each attachment type. If I had to hazard a guess, I would say the one user's really care about is the "use URL as an attachment" as this allows any file type attachment (.pdf, .txt, .doc, etc) provided it is hosted on the web and downloadable. All the others attachments are primitive.

rachelsaunders: okie - added the supported types. PTAL?

As I mentioned in a previous comment in this thread, in a follow-up PR, let's add a full example for this message field (and we can use the URL option in that example).  But let's limit this PR to just adding the attachment support info to get the info out to devs.

firestore-send-email/POSTINSTALL.md
```diff
@@ -62,7 +62,7 @@ Available properties for the `message` field are:
 * **text:** The plaintext content of the email.
 * **html:** The HTML content of the email.
 * **amp:** The [AMP4EMAIL][amp4email] content of the email.
-* **attachments:** An array containing attachment(s); supports [Nodemailer options](https://nodemailer.com/message/attachments/) (be aware that your email has no access to the cloud server's file system)
+* **attachments:** An array containing attachment(s); supports [Nodemailer options](https://nodemailer.com/message/attachments/) of utf-8 string, custom content type, URL, encoded string, data uri, and pregenerated MIME node (be aware that your email has no access to the cloud server's file system)
```
russellwheatley: I guess my suggestion would be to may be change from:
 `supports [Nodemailer options](https://nodemailer.com/message/attachments/) of utf-8 string...` 
to: 
`[Nodemailer options](https://nodemailer.com/message/attachments/) supported: utf-8 string...`

@rachelsaunders what do you think?

rachelsaunders: sure!  wanna make this change and commit it?

russellwheatley: @rachelsaunders changes have been made 👍

Issue #170, opened 2020-01-19
[Firestore-shorten-urls-bitly] Extension will break down if not updated to Bitly APIv4 by deadline March 1st, 2020


babkouew:
Hi Firebase Extension Team -

I recently installed the Firestore-shorten-urls extension via the Firebase Console but while auditing the extension I noticed that calls to the Bitly API were reliant on the library node-bitly which is not maintained anymore.

This is problematic because Bitly is deprecating its old APIv3 for a new one APIv4 and is asking developers to migrate before March 1st, 2020, when APIv3 will be deactivated.

The creator of node-bitly @tanepiper, mentioned that he is looking for a new maintainer for his library and has [no intention of updating it](https://github.com/tanepiper/node-bitly/issues/72) to the new APIv4.

Consequently, the shorten-urls-bitly extension will in all likelihood breakdown in about a month from now since it is using node-bitly v6.0.8, which is incompatible with Bitly APIv4.

Will you implement the necessary changes in time for all the users who are currently using your extension or is it safer to stay away from it for now?

tanepiper:
Hi there, author of the Bitly library here - yes I can confirm I have no intention of spending any time working on the existing bitly library - I've since moved on from needing to use it and have been calling for new maintainers, or for some support to maintain it, but sadly neither have materialised (despite 13k downloads per week - so I suspect it's not just Firebase that's about to break)

In this case, I'd advise switching to https://github.com/vijaypatoliya/bitly-node-api - it seems to support V4 API, although it does seem a bit more complex internally than mine. Also doesn't seem to be actively maintained.

As I've mentioned I have pipelines set up, so if someone wants to provide a PR to fix this and release a new version I'm happy to help with that.

Ehesp:
Hey, thanks for brining this up. I'll take a look at the library you linked and see if we can port it over. Thanks.

Ehesp:
Pushed up a non-breaking PR for this. Decided to just directly send a POST request to the v4 API rather that use the library above, as it's lightweight and also seems overly complicated for our own needs.

Need to test it a bit more before merging.

babkouew:
Excellent - looking forward to it.

tanepiper:
Up to you, but in the end I have decided (after seeing your PR that it might be easier than expected) to update the library. I'm aiming to have it over before end of January. But I agree even my library is probably overkill for what you need https://github.com/tanepiper/node-bitly/pull/73

babkouew:
[...]

PR #171, opened 2020-01-21
fix(firestore-shorten-urls-bitly) migrate to v4 api #170

Ehesp:
See #170 for more details. This directly connects with the v4 API via a POST request.


Ehesp:
Thanks - it was more just to confirm you were ok with direct API integration (but just seen your other comment, so good to go).

PR #178, opened 2020-01-27
Jan 30, 2020 Release

laurenzlong:


firestore-send-email/POSTINSTALL.md
```diff
@@ -8,7 +8,7 @@ You can test out this extension right away:
 
 1.  Add a document with a `to` field and a `message` field with the following content:
 
-    ```
+    ```js
```
rachelsaunders: Just a reminder that during review, we'll need to check that this syntax plays nice with the formatting library used by the Firebase console.  Since the POSTINSTALL file no longer can be printed by the CLI, I think that we're ok there.

PR #184, opened 2020-01-31
chore(firestore-counter): update firebase-admin & firebase-functions packages

### LINKED ISSUE #181: Update firebase-functions and firebase-admin for firestore-counter

firestore-counter

As a follow-up to https://github.com/firebase/extensions/pull/125, typing issues should be fixed and firebase-functions and firebase-admin updated to the latest versions.  See https://cloud.google.com/compute/docs/migrating-to-v1-metadata-server#supported-google-libraries, which indicates we should be using firebase-admin v8.2.0 due to the v1 metadata server migration.

russellwheatley:
fixes #181

### DIFF
--- a/firestore-counter/package.json
+++ b/firestore-counter/package.json
@@ -9,7 +9,7 @@
     "@google-cloud/pubsub": "^1.1.5",
     "deep-equal": "^1.0.1",
     "firebase-admin": "^8.3.0",
-    "firebase-functions": "^2.3.1",
+    "firebase-functions": "^3.3.0",
     "uuid": "^3.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


firestore-counter/package.json
```diff
@@ -5,11 +5,11 @@
   "license": "Apache-2.0",
   "description": "Auto-scalable counters for your app.",
   "dependencies": {
-    "@google-cloud/firestore": "^1.3.0",
+    "@google-cloud/firestore": "^3.0.0",
```
laurenzlong: is this still being used?

russellwheatley: Good catch, it has been removed now.

Issue #167, opened 2020-01-18
[firestore-send-email] Allow specifying from name in "Default FROM address" in configuration


jcapogna:
###  Extension name

This feature request is for extension: firestore-send-email

### What feature would you like to see?

I would like to be able to specify a name in the "Default FROM address" configuration field. 
Currently only a raw email address passes the validation rules for the field. 

This could either be done as a separate configuration option or by allowing email addresses of the following format: "My App \<noreply@myapp.com\>".

Specifying email addresses of the format "My App \<noreply@myapp.com\>" works when manually specifying the 'from' field, so I believe this just change is just to modify what is allowed for the configuration of the "Default FROM address" option. 

### How would you use it?

This would allow the default from address to be more user friendly, instead of just the raw email.

Ehesp:
Hey @jcapogna - could you confirm whether [this regex](https://regex101.com/r/H4gvGO/2) covers your use cases? Think it should!

jcapogna:
@Ehesp That looks good to me

PR #185, opened 2020-02-03
feat(firestore-send-email): update "Default FROM address" param to allow name to be added

russellwheatley:
This PR resolves this feature request #167.

### DIFF
--- a/firestore-send-email/extension.yaml
+++ b/firestore-send-email/extension.yaml
@@ -109,7 +109,7 @@ params:
     required: true
     description: >-
       The email address to use as the sender's address (if it's not specified in the added email document).
-    example: foobar <foobar@example.com>  
+    example: Friendly Firebaser <foobar@example.com>  
 
   - param: DEFAULT_REPLY_TO
     type: string
--- a/firestore-send-email/extension.yaml
+++ b/firestore-send-email/extension.yaml
@@ -105,7 +105,7 @@ params:
     type: string
     label: Default FROM address
     validationRegex: ^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$|^.*<(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})>$
-    validationErrorMessage: Must be a valid email address
+    validationErrorMessage: Must be a valid email address or valid name plus email address
     required: true
     description: >-
       The email address to use as the sender's address (if it's not specified in the added email document). 
--- a/firestore-send-email/extension.yaml
+++ b/firestore-send-email/extension.yaml
@@ -104,7 +104,7 @@ params:
   - param: DEFAULT_FROM
     type: string
     label: Default FROM address
-    validationRegex: ^\S+@\S+\.\S+$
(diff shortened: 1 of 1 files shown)


firestore-send-email/extension.yaml
```diff
@@ -104,11 +104,12 @@ params:
   - param: DEFAULT_FROM
     type: string
     label: Default FROM address
-    validationRegex: ^\S+@\S+\.\S+$
+    validationRegex: ^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$|^.*<(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})>$
     validationErrorMessage: Must be a valid email address
     required: true
     description: >-
       The email address to use as the sender's address (if it's not specified in the added email document).
+    example: foobar <foobar@example.com>
```
laurenzlong: "foobar" is not obviously someone's name, how about something like: `Friendly Firebaser <foobar@example.com>` ?

russellwheatley: No problem, I wasn't sure what would be reasonable.

firestore-send-email/extension.yaml
```diff
@@ -104,11 +104,12 @@ params:
   - param: DEFAULT_FROM
     type: string
     label: Default FROM address
-    validationRegex: ^\S+@\S+\.\S+$
+    validationRegex: ^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$|^.*<(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})>$
     validationErrorMessage: Must be a valid email address
     required: true
     description: >-
       The email address to use as the sender's address (if it's not specified in the added email document).
+    example: Friendly Firebaser <foobar@example.com>
```
rachelsaunders: Is the main use case that a dev will want to use a name + email address here?  If not, I would leave the example with just the email address so that a dev doesn't think that they need to use the name, too.

_Suggested change:_

description: >-
  The email address to use as the sender's address (if it's not specified in the added email document). You can optionally include a name with the email address (`Friendly Firebaser <foobar@example.com>`).

example: foobar@example.com

firestore-send-email/extension.yaml
```diff
@@ -104,11 +104,13 @@ params:
   - param: DEFAULT_FROM
     type: string
     label: Default FROM address
-    validationRegex: ^\S+@\S+\.\S+$
+    validationRegex: ^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$|^.*<(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})>$
     validationErrorMessage: Must be a valid email address
```
rachelsaunders: Is this error message still accurate / appropriate given the new options in the regex?

laurenzlong: Good catch! This is checking that the the input is either: an email address, or name plus email address

rachelsaunders: Can we produce an error message that's conditional upon which part of the regex got violated?

rachelsaunders: ping on this thread :-)

russellwheatley: I don't think that this is possible. `extension.yaml` is used to setup the extension with set parameters. (e.g. `validationRegex ` & `validationErrorMessage `). However, I'll defer that call to @laurenzlong as I have no access or knowledge of `extensions` deployment context.

laurenzlong: Not possible unfortunately.

rachelsaunders: directly updated the file to include "or valid name plus email address"

firestore-send-email/extension.yaml
rachelsaunders: this looks done, yeah?  Resolve?

PR #195, opened 2020-02-08
feat(firestore-translate-text) add shouldUpdate option to config

### LINKED ISSUE #194: [firestore-translate-text] add shouldUpdate option to config

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

This feature request is for extension: **\_** `firestore-translate-text`

I propose adding a configuration option to be able to decide if an update on the input should be propagated to the translations after creation.

It is a nice feature keeping translations in sync with the original text, in case it is not an issue that the text was machine-translated.

In other situations, I believe it is hardly the case the translated text is good enough to be published immediately. Therefore, I suggest that when creating the translation, translate the text as before, but on subsequent changes, update the translations only if the new `config.shouldUpdate` option is set to `true`. This could default to `true`, so current flows should not break.

Added a PR(#195) myself for review.
### How would you use it?

I develop a multi-language website, where the content on the main page is saved in Firestore, and there exists an Admin dashboard from where those texts can be edited. I would like to be nice with the admin and suggest him translations when he creates the texts. But if he decides to do small modifications to the original text, right now it would result in changing every other language texts as well (which he probably had to tinker). This is only my use case though, I can imagine there are others who would benefit from an "only recommend the first time translation".

balazsorban44:
A nice feature to make subsequent translations optional.

This closes #194, where the issue is discussed in detail.

### DIFF
--- a/firestore-translate-text/POSTINSTALL.md
+++ b/firestore-translate-text/POSTINSTALL.md
@@ -29,7 +29,7 @@ Whenever you write a string to the field `${param:INPUT_FIELD_NAME}` in `${param
 }
 ```
 
-If the `${param:INPUT_FIELD_NAME}` field of the document is updated, then the translations will be automatically updated, as well.
+By default, if the `${param:INPUT_FIELD_NAME}` field of the document is updated, then the translations will be automatically updated, as well. To skip re-translating the translations, set `${param:SHOULD_UPDATE}` to `false`.
 
 ### Monitoring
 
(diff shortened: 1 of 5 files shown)


balazsorban44:
@googlebot I signed it!

karayu:
Closing this for now. More details in: https://github.com/firebase/extensions/issues/194

PR #205, opened 2020-02-22
chore(*): aligned the order and spacing of fields within extension.yaml files

rachelsaunders:
Each extension.yaml file had a slightly different ordering and spacing for its fields.  This PR aligns this ordering and spacing across all extensions.

### DIFF
--- a/auth-mailchimp-sync/extension.yaml
+++ b/auth-mailchimp-sync/extension.yaml
@@ -53,6 +53,7 @@ resources:
       eventTrigger:
         eventType: providers/firebase.auth/eventTypes/user.create
         resource: projects/${PROJECT_ID}
+
   - name: removeUserFromList
     type: firebaseextensions.v1beta.function
     description:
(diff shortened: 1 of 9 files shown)


auth-mailchimp-sync/extension.yaml
```diff
@@ -51,6 +53,7 @@ resources:
       eventTrigger:
         eventType: providers/firebase.auth/eventTypes/user.create
         resource: projects/${PROJECT_ID}
+
```
laurenzlong: I'd prefer to not add line breaks between items in a list (`-` indicates items in a list). That's very non-standard for yaml files, although it does work (much to my surprise).

rachelsaunders: The other extension.yaml files with multiple resource objects have line breaks.  Also, when there are multiple parameters, we also have line breaks.

In most of the yaml files used internally, we use line breaks, as well.

laurenzlong: Oh interesting, I never noticed that. Ya learn something new everyday!

rachelsaunders: ya, yaml files are flaky, but with this, there're not bad :-)

PR #208, opened 2020-02-26
chore(storage-resize-images) Move package.json into functions folder

laurenzlong:
Having package.json inside extension root doesn't make as much sense as having it inside of the functions folder, since it's the package.json for the functions, not for the entire extension. 

Once you've verified that this works (I've already verified on my side, but good to double check.) Can one of you also do this for all the other extensions please?


Salakar:
Hey @laurenzlong - I see a couple issues with making this change;

 - Lerna mono-repo issues, e.g.;
   - will no longer install packages for each extension
   - post install at root will no longer `clean` & `build` js in each package
 - CI (tests in this extension now fail)
 - `generate-readme` script in each `package.json` will no longer work
 - `firestore-bigquery-export` has a root package.json alongside a separate `package.json` in functions

Salakar:
Closing as discussed - we'll draft another PR against the translate text extension for this

PR #219, opened 2020-03-11
fix(delete-user-data): update firebase-tools version

Ehesp:
Upgrades the `firebase-tools` version, which updates the underlying `google-auto-auth` package.

Fixes #206

## Testing

image

PR #221, opened 2020-03-12
chore(*): remove package-lock.json from repository

Ehesp:
Removing package-lock from git cache & adding to gitignore as per https://github.com/firebase/extensions/pull/219#pullrequestreview-372989588.

@laurenzlong I didn't add anything to the `.npmrc` as nothing is being published to npm at the moment.


laurenzlong:
`.npmrc` also affects how npm behaves locally, which is important since during the release process, what gets uploaded as a source is what is on the engineer's computer. So if a package-lock.json got created, then it gets uploaded into the archive that becomes the source for the extension. So we get into the same problem.

laurenzlong:
@Ehesp In case my comment above got missed, can you please make a follow up PR to add .npmrc? (Rationale above)

laurenzlong:
Never mind, I realized that it is not an issue, since the release process pulls from Github, and does not use local files.

PR #224, opened 2020-03-13
refactor(firestore-translate-text): clarify log message

joehan:
From chatting with Rachel, we've realized that the words 'configuration' and 'config' are overloaded in the context of extensions. Since these are actually parameter values, I think this will be slightly clearer.


laurenzlong:
Can you fix CI please?

PR #227, opened 2020-03-16
refactor(firestore-counter): use handlers

### LINKED ISSUE #191: [firestore-counter] Use handler namespace for all exported functions
Describe your configuration

- Extension name: firestore-counter
- Extension version: any
We currently define some exported functions using the normal functions namespace instead of the handlers (ie https://github.com/firebase/extensions/blob/next/firestore-counter/functions/src/index.ts#L77). This is misleading - any triggers declared within the code will be ignored in favor of those in extension.yaml. We should switch these to use the handler namespace.

No rush on this since it doesn't break anything, it just makes the code less readable.

russellwheatley:
Fixes #191.

### DIFF
--- a/firestore-counter/functions/lib/index.js
+++ b/firestore-counter/functions/lib/index.js
@@ -54,7 +54,7 @@ exports.controllerCore = functions.handler.pubsub.topic.onPublish(() => __awaite
 /**
  * Backwards compatible HTTPS function
  */
-exports.controller = functions.https.onRequest((req, res) => __awaiter(void 0, void 0, void 0, function* () {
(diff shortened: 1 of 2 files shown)


russellwheatley:
The first two images show the document count being updated for a document with a `hello-world` id  using the JavaScript counter library provided by Firebase Extensions.
Screenshot 2020-03-16 at 16 57 40
Screenshot 2020-03-16 at 16 58 00
Below is a log of the `controllerCore` function executing successfully.
Screenshot 2020-03-16 at 17 03 10
Below is a log of the `onWrite` function executing successfully.
Screenshot 2020-03-16 at 17 03 33

PR #242, opened 2020-04-04
Refer to these extensions as first-party, instead of official

puf:


laurenzlong:
Hey @puf "Official Extensions" is actually a term we decided on based on User Research. Do you have a reason for why it should be "first party" instead?

laurenzlong:
As discussed offline, we are going to stick with "official" for now.

PR #244, opened 2020-04-06
test(firestore-bigquery-export) Use`mocha` instead of `jest` for tests

IanWyszynski:
This PR documents the local-development flow for this extension, which likely to be different than what we have for the others because we rely on npm dependency resolution for the extension to run.

The PR also migrates this extension to uses `jest` as opposed to `mocha`. `jest` appears to be the standard practice for other extensions in this repo.


firestore-bigquery-export/functions/package.json
```diff
@@ -1,25 +0,0 @@
-{
```
jhuleatt, line 1: FYI @laurenzlong this PR takes out the functions-folder-level package.json

Issue #140, opened 2019-12-19
unable to use rezised images through public url


harshapulikollu:
### [READ] Step 1: Are you in the right place?
Yes.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **\_** (`storage-resize-images`)
- Extension version: **\_0.1.3**

### [REQUIRED] Step 3: I'm unable to access the resized images through a public(download) url from my flutter app, like I'm able to retrieve original image. I don't see any reason to use this extension if i can't access those resized images by manipulating the original Image URL.
please provide me a way how can I access those resized images from my flutter app.

brandonpearcy:
+1 to this. I have not found a convenient way to make the resized images publicly accessible without doing manual post-processing with additional cloud functions. 

The generated thumbs have no access token visible in Firebase console, there is no clickable link to view the images, and the preview of the image never loads (infinite spinner).

<img width="1267" alt="Screen Shot 2019-12-21 at 10 57 12 AM" src="https://user-images.githubusercontent.com/17098317/71312534-f8adaf00-23e0-11ea-9b29-e0936cfc026b.png">

I would expect that either: 

(1) there exists a simple transformation from the original firebase storage URL to access the resized images (e.g., something you could hardcode in client logic), or 
(2) this firebase extension comes with documentation describing how to best post-process the resized images to make them accessible

NicolasMaas:
> +1 to this. I have not found a convenient way to make the resized images publicly accessible without doing manual post-processing with additional cloud functions.
> 
> The generated thumbs have no access token visible in Firebase console, there is no clickable link to view the images, and the preview of the image never loads (infinite spinner).
> 

I'm facing the same issue... Seems kind of pointless at the moment to use this extension at all.

TiagoGouvea:
Happening with me also.

On my "dev" project, sending a new image trough firebase console or by code, it generates the public URL of the resized file.
- [Resized image on dev project](https://firebasestorage.googleapis.com/v0/b/whereami-dev.appspot.com/o/images%2F98453cef-5e3a-4423-93ce-7fde2e61c67b_650x500.jpg?alt=media&token=5ff48935-88dc-4973-9968-5c13001a9408)

On my "prod" project it don't.
[...]

PR #279, opened 2020-04-22
fix(storage-resizes-images) Fixes issue where resized image is not rendered in Firebase Console (Issue #140)

sitefinitysteve:
The problem with the current extension is outlined [here](https://github.com/firebase/extensions/issues/140) 

This should add a new configuration option to NOT delete the token, defaults to true to preserve existing functionality, users can go in to disable it if they don't want it happening.


laurenzlong:
@firebase/invertase Can you verify that this works? If the option is set to "false" then does it allow the resized image to be downloaded? If yes, then it seems that this should not be a param, but the download token should just always be preserved, and we would make a minor version bump.

sitefinitysteve:
No, well I mean right *now* you cant download the resized images, so defaulting to false I assumed would not break existing configurations for users who have it working this way now and for some reason want it like that?

It shouldn't be a param, and the delete command should just go away honestly... but I didn't want to rock the boat.

Like I can't for the life of my understand why this is here
https://github.com/firebase/extensions/blob/685443f5e44ac2e69304f10e2c497702471e0935/storage-resize-images/functions/src/index.ts#L205

Wouldn't it be the inverse, more people would want that access token there so the resized image just works?

laurenzlong:
> No, well I mean right _now_ you cant download the resized images, so defaulting to false I assumed would not break existing configurations for users who have it working this way now and for some reason want it like that?
> 
> I agree 100%, it shouldn't be a param, and the delete command should just go away honestly... but I didn't want to rock the boat.
> 
> Like I can't for the life of my understand why this is here
> https://github.com/firebase/extensions/blob/685443f5e44ac2e69304f10e2c497702471e0935/storage-resize-images/functions/src/index.ts#L205
> 
> Wouldn't it be the inverse, more people would want that access token there so the resized image just works?

That's why I suggested a minor version bump, which for versions under 1.0.0 indicates a breaking change. (I think this is worth doing a breaking change for, rather than adding an unnecessary param)

sitefinitysteve:
Ok, so if I have the blessing... remove the param and delete entirely, bump to 0.2.0? Sound good?

laurenzlong:
@sitefinitysteve Yes please, also you do not have to increment the version or add to the changelog in this PR, that gets automatically generated later when we do the release. Thanks for your contribution!

storage-resize-images/functions/lib/index.js
```diff
@@ -161,7 +164,6 @@ const resizeImage = ({ bucket, originalFile, fileDir, fileNameWithoutExtension,
         else {
             metadata.cacheControl = objectMetadata.cacheControl;
         }
-        delete metadata.metadata.firebaseStorageDownloadTokens;
         // Generate a resized image using Sharp.
```
NickFoden, line 167: Ah of course, line 164 the source of all our pain.  :-)

russellwheatley:
Hey @laurenzlong, I can confirm that this works. 

This is a resized image before the change is made. As you can see there is no image displayed because it need a new token creating:
Screenshot 2020-04-23 at 10 12 40

This is a resized image after the change. Image is displayed because the token is preserved:
Screenshot 2020-04-23 at 10 10 54

sitefinitysteve:
I mean i've been using this fix by editing the cloud function directly for a few weeks, but new releases keep blowing away the edit, I know she works.

On that note, what's the release schedule like for something like this?

laurenzlong:
Thanks @sitefinitysteve and @russellwheatley! This is failing the linter, could you please fix by running `npm run format`? 

We can slot this in for the May 7 release.

laurenzlong:
Thanks for the contribution!

PR #304, opened 2020-05-06
fix(firestore-bigquery-export): clean column names for generating schema views

### LINKED ISSUE #247: firestore-bigquery-export-gen-schema-view fails if field uses forbidden char

yep

- Extension name: **firestore-bigquery-export** (`gen-schema-view`)
- Extension version: **07d050ac1f39248e5b77e6503a7aab51a58b68f7**

#### Steps to reproduce:

BigQuery view generation from view schema fails if field contains not-alpha/digit/underscore. For example:

```json
{
  "fields": [
    {
      "name": "subscriber:cancelled",
      "type": "boolean"
    }
  ]
}
```

Error:
```
BigQuery creating schema view audiences_schema_audiences_latest:
Schema:
{"fields":[{"name":"subscriber:cancelled","type":"boolean"}]}
Query:
SELECT
  document_name,
  timestamp,
  operation,
  `PROJECT_ID.firestore_export.firestoreBoolean`(JSON_EXTRACT(data, '$.subscriber:cancelled')) AS subscriber :cancelled
FROM
  `PROJECT_ID.firestore_export.audiences_raw_latest`
{"code":400,"errors":[{"message":"Syntax error: Unexpected \":\" at [5:111]","domain":"global","reason":"invalidQuery","location":"q","locationType":"parameter"}],"message":"Syntax error: Unexpected \":\" at [5:111]"}
Syntax error: Unexpected ":" at [5:111]
```

Quoting qualified name does not fix the issue, since there is requirement for query:
> Fields must contain only letters, numbers, and underscores, start with a letter or underscore, and be at most 128 characters long.

There is no such limitation for fields in firebase documents.

This issue can be fixed with properly generated qualified field name, during SQL schema generation:
```typescript
function qualifyFieldName(prefix: string[], name: string): string {
  const notAlphaDigitUnderscore = /([^a-zA-Z0-9_])/;
  name = name.replace(notAlphaDigitUnderscore, '_');
  // TODO: first char. length limit

  return prefix.concat(name).join("_");
}
```

##### Expected result

Field name should be automatically normalized and no errors occurred

##### Actual result

fails to generate view

russellwheatley:
fixes #247

Screenshot of successfully generating schema view without error
Screenshot 2020-05-06 at 15 39 42

Screenshot of view in Big Query updated with column name successfully changed to `colon_colon`.
Screenshot 2020-05-06 at 16 03 32

### DIFF
--- a/firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
+++ b/firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
@@ -576,7 +576,10 @@ export function subSelectQuery(query: string, filter?: string[]): string {
 }
 
 function qualifyFieldName(prefix: string[], name: string): string {
-  return prefix.concat(name).join("_");
+  const notAlphaDigitUnderscore = /([^a-zA-Z0-9_])/;
+  const cleanName = name.replace(notAlphaDigitUnderscore, "_");
+
+  return prefix.concat(cleanName).join("_");
 }
 
 export function latest(tableName: string): string {
--- a/firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
+++ b/firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
@@ -576,8 +576,8 @@ export function subSelectQuery(query: string, filter?: string[]): string {
 }
 
 function qualifyFieldName(prefix: string[], name: string): string {
-  const notAlphaDigitUnderscore = /([^a-zA-Z0-9_])/;
-  const cleanName = name.replace(notAlphaDigitUnderscore, "_");
+  const notAlphanumericUnderscore = /([^a-zA-Z0-9_])/g;
+  const cleanName = name.replace(notAlphanumericUnderscore, "_");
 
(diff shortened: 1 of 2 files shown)


firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
```diff
@@ -576,7 +576,10 @@ export function subSelectQuery(query: string, filter?: string[]): string {
 }
 
 function qualifyFieldName(prefix: string[], name: string): string {
-  return prefix.concat(name).join("_");
+  const notAlphaDigitUnderscore = /([^a-zA-Z0-9_])/;
```
Salakar, line 579:
```suggestion
  const notAlphaDigitUnderscore = /([^a-zA-Z0-9_])/g;
```

I believe this regex needs the global modifier - otherwise it will only replace the first character it comes across and leaving the rest

Salakar: Nit: can we call this `notAlphanumericUnderscore`?

PR #306, opened 2020-05-11
fix(storage-resize-images): fix bug where resized image did not keep orientation of original image (Issue #290)

### LINKED ISSUE #290: auto-orient images and thumbnails when uploaded to firebase storage

This feature request is for extension: `storage-resize-images`

I have a public script which reads the EXIF data from jpeg images

As you are probably aware of... iPhone photos which are taken on portrait mode would appear sideways if you show the resulting thumbnail or uploaded image to the user inside an image tag even after your plugin completes generating the thumbnails.

I was successfully able to set a sample script to auto-orient the jpeg images by preselecting EXIF orientation image which value is read as: 6 (six)

I do NOT want to mess with the code you offer as prebuilt plugins... So I am kindly asking to include it as soon as possible.

If you need additional data please let me know.

Stay safe

### How would you use it?

russellwheatley:
fixes #290

Shows the image being displayed correctly using the image's EXIF orientation property in the Mac's image viewer:
Screenshot 2020-05-11 at 13 20 24

Shows the same image displayed in VScode which does not use the EXIF data orientation property:
Screenshot 2020-05-11 at 13 10 38

This shows how the resized image is displayed in the firebase console (The resized image loses EXIF data) without this PR change:
Screenshot 2020-05-11 at 13 09 42

This shows the resized image being displayed correctly. It doesn't have the EXIF data, but it was rotated at the time of resizing using the EXIF data orientation property:
Screenshot 2020-05-11 at 13 10 01

### DIFF
--- a/storage-resize-images/functions/lib/index.js
+++ b/storage-resize-images/functions/lib/index.js
@@ -138,6 +138,7 @@ function resize(originalFile, resizedFile, size) {
         throw new Error("height and width are not delimited by a ',' or a 'x'");
     }
     return sharp(originalFile)
+        .rotate()
         .resize(parseInt(width, 10), parseInt(height, 10), { fit: "inside" })
         .toFile(resizedFile);
 }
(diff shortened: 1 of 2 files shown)


EliasTouil:
Thank you :)

PR #315, opened 2020-05-19
release notes for May 20 release

tinaliang:


storage-resize-images/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## Version 0.1.8
+
+fixed - fix bug where resized image did not keep orientation of original image (Issue #290
+fixed - release automation script for Github release (#302)
+fixed - clear cache so that image names can be reused (Issue #286) (#297)
+fixed - switched ImageMagick for Sharp library (#207)
```
rachelsaunders, line 6: These two items are also listed in the previous release.  I think that we need some clarity about what exactly was changed to have the same fixes mentioned again.

tinaliang: @laurenzlong can you provide more context here?

laurenzlong: only the first line should be in this release.

rachelsaunders: When you say "first line"... do you mean that only the first bullet (the one about orientation) should be in this release?

laurenzlong: yes.

storage-resize-images/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## Version 0.1.8
+
+fixed - fix bug where resized image did not keep orientation of original image (Issue #290
```
rachelsaunders, line 3:
```suggestion
fixed - Resized images now maintain the same orientation as the original image. (Issue #290)
```

PR #329, opened 2020-05-28
feat(storage-resize-images): skip resizing of vector images (Fixes #326).

### LINKED ISSUE #326: storage-resize-images resizes SVG files, resulting in wrong content types

...

- Extension name: storage-resize-images
- Extension version: 0.1.8

Without an explicit output format, [Sharp infers it from the input file](https://sharp.pixelplumbing.com/api-output#tofile):

> If no explicit format is set, the output format will match the input image, except GIF and SVG input which become PNG output.

This causes problems with SVG images:

1. Upload an SVG (content type `image/svg+xml`)
2. The extension proceeds because the validation passes (`contentType.startsWith("image/")`)
3. Sharp converts the file to PNG
4. Metadata is copied from the original object, so ContentType is set to `image/svg+xml`
5. The browser (including Firebase Console) isn't able to render the image ("invalid xml")

The extension has to set a content type, otherwise it defaults to `application/octet-stream`.

Possible solutions:

1. Skip vector images. Vector images aren't really resized, so it's probably not an interesting use case for this extension.
2. Set the content type based on the resulting Sharp format

I can make a PR if you'd like. My proposal would be to limit this extension to the formats supported by Sharp (JPEG, PNG, WebP, GIF, SVG and TIFF) *except* SVG.

russellwheatley:
fixes #326
- SVG images are now skipped from being resized as they're already resizable.
- removed unnecessary abstraction for checking image type.
- TypeScript introduced further updates to compiled code. In fact, it has changed every extension's compiled code. I'll make a PR with latest compiled code.

Here is a screenshot of an SVG image being skipped:
Screenshot 2020-05-28 at 15 33 27


nicoqh:
What about the other `image/*` content types that aren't supported by Sharp? E.g. `image/bmp`, `image/vnd.microsoft.icon`, and lots of obscure ones like `image/x-cmu-raster`

Here's a list of the formats Sharp supports (excluding SVG) when passing a filesystem path to a file:

```javascript
const supportedContentTypes = [
  "image/gif",
  "image/jpeg",
  "image/png",
  "image/tiff",
  "image/webp",
];

if(!supportedContentTypes.includes(contentType)) { ... return; }
```

russellwheatley:
Hey @nicoqh, thank you for the suggestion, I've updated the code to match. I also realise that `gif` images aren't exactly supported as the resized image is just the first frame of the `gif`. Not sure whether `gif` images should also be excluded, @laurenzlong?

storage-resize-images/functions/src/index.ts
```diff
@@ -40,6 +39,18 @@ sharp.cache(false);
 admin.initializeApp();
 
 logs.init();
+
+/**
+ * Supported file types
+ */
+const supportedContentTypes = [
+  "image/gif",
+  "image/jpeg",
```
Ehesp, line 48: If the image extension is `jpg`, does it still come through as the `jpeg` mime type?

nicoqh:
@russellwheatley In addition to that, GIFs are converted by Sharp to PNGs, so it'll mess up the content type. The `contentType` is copied from the original to the resized object :\

storage-resize-images/functions/src/index.ts
russellwheatley: Yeah, it is the same content-type for both: https://stackoverflow.com/a/37266399

russellwheatley:
@nicoqh I think it makes sense to remove `gif` types as it currently stands.

storage-resize-images/functions/lib/logs.js
```diff
@@ -26,6 +25,9 @@ exports.noContentType = () => {
 exports.contentTypeInvalid = (contentType) => {
     console.log(`File of type '${contentType}' is not an image, no processing is required`);
 };
+exports.unsupportedType = (unsupportedTypes, contentType) => {
+    console.log(`image type '${contentType}' is not supported, here are the supported file types: ${unsupportedTypes.join(", ")}`);
```
Salakar, line 29:
```suggestion
    console.log(`Image type '${contentType}' is not supported, here are the supported file types: ${unsupportedTypes.join(", ")}`);
```

storage-resize-images/functions/src/logs.ts
```diff
@@ -30,6 +30,17 @@ export const contentTypeInvalid = (contentType: string) => {
   );
 };
 
+export const unsupportedType = (
+  unsupportedTypes: string[],
+  contentType: string
+) => {
+  console.log(
+    `image type '${contentType}' is not supported, here are the supported file types: ${unsupportedTypes.join(
```
Salakar, line 38:
```suggestion
    `Image type '${contentType}' is not supported, here are the supported file types: ${unsupportedTypes.join(
```

PR #330, opened 2020-05-28
chore(*): TS compiled code update

russellwheatley:
- Ran `npm run build` from root directory, which has changed the TypeScript compiled JS code. It must be a TypeScript version update that has changed the way it compiles to JS.


Ehesp:
What change has caused the complied code output to be different?

russellwheatley:
It'll almost certainly be an updated TypeScript version. We're pointing at version `^3.6.3` in the root `package.json`, but a look in my node modules reveals TypesScript is now on a later version, `3.7.2`.

russellwheatley:
I guess the compiled code for each extension will be updated as we roll out changes so I'm going to go ahead and close this.

Issue #323, opened 2020-05-22
Recycling download tokens is problematic (storage-resize-images)


nicoqh:
### [READ] Step 1: Are you in the right place?

...

### [REQUIRED] Step 2: Describe your configuration

- Extension name: storage-resize-images
- Extension version: 0.1.8

### [REQUIRED] Step 3: Describe the problem

In a previous version, `firebaseStorageDownloadTokens` was deleted from the metadata object before the resized image was uploaded.

#279 removed this deletion, resulting in `firebaseStorageDownloadTokens` being copied from the original object to the new (resized) object.

One issue is that download tokens should be unique per object, ref https://github.com/firebase/extensions/issues/34#issuecomment-552595619:

> I chatted with the Cloud Storage for Firebase team about firebaseStorageDownloadTokens. The field should be unique per object. If it isn't set on an object, it will be set automatically by the Cloud Storage for Firebase api when getDownloadURL() is called.

My use case illustrates the problem: Files are not public, but I display the resized versions using `getDownloadURL()`. The resulting download URL contains the token, which can be used for the _original file_ as well. But the original image should not be accessible! Also, all thumbnails use the same token, which is equally problematic.

It seems to me that copying `firebaseStorageDownloadTokens` is a workaround to make sure the resized images are displayed in the Firebase Console (which won't work without a token), but that tokens are meant to be unique for good reasons.

**Two possible solutions:**

1. Let the extension generate a download token (tokens are UUIDv4) an add it to the resized image's metadata map. **This works perfectly, and images are viewable in the Firebase Console as well.** Example:

```javascript
    // `objectMetadata` is from the original image that we want to resize
    const metadata: any = {
      contentDisposition: objectMetadata.contentDisposition,
      contentEncoding: objectMetadata.contentEncoding,
      contentLanguage: objectMetadata.contentLanguage,
      contentType: contentType,
      metadata: {
        ...(objectMetadata.metadata || {}),
        firebaseStorageDownloadTokens: "123-TEST-TOKEN", // Should be a call to uuid4()
      }
    };
```

`firebaseStorageDownloadTokens` is a Firebase concept, so I don't see any downsides with using it in an official Firebase Extension.

What about people who want to overwrite the original image and keep the same download URL (ref #268)? This could be a config option (if "replace original file", then reuse the old token, otherwise generate download tokens for each resized version).

[...]

PR #322, opened 2020-05-22
docs(storage-resizes-images) Update docs regarding download tokens

nicoqh:
`firebaseStorageDownloadTokens` is no longer deleted, ref #140 and #279

PR #351, opened 2020-06-15
fix(storage-resize-images): generate fresh token for resized image (Issue #323).

Ehesp:
As discussed in https://github.com/firebase/extensions/issues/323, image download tokens are being copied which is problematic. This PR updates the flow to check if a download token exists on the original image, and if so, generates a fresh token on the image being resized.

- [x] Peer testing
- [x] Documentation updates (https://github.com/firebase/extensions/pull/322 impacted)


russellwheatley:
LGTM

russellwheatley:
Able to access images with token created using `uuidv4` library.
Screenshot 2020-06-16 at 09 46 24
Screenshot 2020-06-16 at 09 46 41

PR #357, opened 2020-06-19
fix(firestore-bigquery-export): record document id in schema view and revert timestamp change

### LINKED ISSUE #347: firebase-bigquery-export:schema-view Timestamps filtered out on "latest" view

- Extension name: **firebase-bigquery-export/gen-schema-view** 
- Extension version: **0.1.9**

When generating a schema for the firebase bigquery export extension, Firebase Timestamp values do not show up in the `latest` schema but do in the changelog. 

Looking in the code, there is a message that claims `TIMESTAMP` values can not be grouped, and are thus excluded. I feel like this may be a mistake. Here's the sample code in `firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts`, 

```typescript
   // For "latest" data views, certain types of fields cannot be used in
    // "GROUP BY" clauses. We keep track of them so they can be explicitly
    // transformed into groupable types later.
    if (field.type === "array") {
      arrays.push(qualifyFieldName(prefix, field.name));
    }
    if (field.type === "geopoint") {
      geopoints.push(qualifyFieldName(prefix, field.name));
    }

     if (field.type === "timestamp") {
       timestamps.push(qualifyFieldName(prefix, field.name));
     }
```

I don't understand why this applies to timestamps (it's very possible I simply am ignorant on the subject -- I'm no SQL expert). When I comment out the `timestamp` clause, the schemas and resulting views work exactly as expected, with SQL Timestamp values where I had registered a timestamp in my schema. 

#### Steps to reproduce:
- Create a firestore collection with a timestamp field. 
- Create a schema.json file for this collection and register the timestamp field. 
- Run the generate-schema-view script
- Notice the timestamp does not show up when you execute a select statement on the resulting `_latest` view. 

My "fix"
Simply remove the if statement `schema.ts` that adds timestamp values to the exclusion array.

russellwheatley:
fixes #347 
Reverts the changes mistakenly made in this [commit](https://github.com/firebase/extensions/commit/e0fea7d4b986a54ad1044bcc26764899c0d40474).

Latest view table as it currently stands creates seconds and nanoseconds column for a timestamp type:
Screenshot 2020-06-19 at 12 48 18

Reverted back to it's original implementation where a timestamp type creates a single column with a UTC date string:
Screenshot 2020-06-19 at 12 48 51

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/fixtures/sql/latestConsistentSnapshot.txt
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/fixtures/sql/latestConsistentSnapshot.txt
@@ -5,6 +5,7 @@
 --   data: A raw JSON payload of the current state of the document.
 SELECT
   document_name,
+  document_id,
   timestamp,
   event_id,
   operation,
@@ -13,6 +14,7 @@ FROM
   (
     SELECT
       document_name,
+      document_id,
       FIRST_VALUE(timestamp) OVER(
         PARTITION BY document_name
         ORDER BY
@@ -48,6 +50,7 @@ WHERE
   NOT is_deleted
 GROUP BY
   document_name,
+  document_id,
   timestamp,
   event_id,
   operation,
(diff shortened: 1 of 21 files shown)


russellwheatley:
Not sure why this isn't passing.

russellwheatley:
Do not merge until #377 has been merged into this branch.

PR #381, opened 2020-07-08
chore(firestore-translate-text): use logger

russellwheatley:
- Tested against node version 10 using Firebase Functions logger as per [docs](https://firebase.google.com/docs/functions/writing-and-viewing-logs)
- Tested against node version 8 (see latest comment below)
- Updated `firebase-functions` package to `^3.7.0`

Below is a `logger.log(String, Config)` call: 

Screenshot 2020-07-08 at 12 55 41
The rest of the console output:
Screenshot 2020-07-08 at 12 55 58


firestore-translate-text/functions/lib/logs/index.js
```diff
@@ -16,68 +16,68 @@
...
-    console.error(...messages_1.messages.translateInputToAllLanguagesError(string, err));
+    firebase_functions_1.logger.error(...messages_1.messages.translateInputToAllLanguagesError(string, err));
 };
 exports.updateDocument = (path) => {
-    console.log(...messages_1.messages.updateDocument(path));
+    firebase_functions_1.logger.log(messages_1.messages.updateDocument(path));
 };
 exports.updateDocumentComplete = (path) => {
-    console.log(...messages_1.messages.updateDocumentComplete(path));
+    firebase_functions_1.logger.log(messages_1.messages.updateDocumentComplete(path));
```
dackers86, line 82: Is there a reason we are using a variable name with a suffix of _1.

Also can logger be deconstructed? I think this is available in Node 10?

```js 
const {logger} require("firebase-functions");
```

firestore-translate-text/functions/src/logs/index.ts
```diff
@@ -14,99 +14,99 @@
...
 };
 
 export const inputFieldNameIsOutputPath = () => {
-  console.error(messages.inputFieldNameIsOutputPath());
+  logger.error(messages.inputFieldNameIsOutputPath());
 };
 
 export const start = () => {
-  console.log(...messages.start(config));
+  logger.log(messages.start());
```
jhuleatt, line 69: Looks like we lost the config printout here - is that intentional? If not, it could be added as the second argument to `logger.log`

Salakar: Looks like `config` was moved to the `init` logger

firestore-translate-text/functions/src/logs/messages.ts
```diff
@@ -21,10 +21,7 @@ export const messages = {
   ],
   inputFieldNameIsOutputPath: () =>
     "The `Input` field name must not be the same as an `Output` path for this extension to function correctly",
-  start: (config = {}) => [
-    "Started execution of extension with configuration",
-    config,
-  ],
+  start: () => "Started execution of extension with configuration",
```
Salakar, line 24: Wording here should be adjusted (` with configuration`) if the config has been moved to another fn

firestore-translate-text/functions/src/logs/index.ts
russellwheatley: I updated the code to previous behaviour so the config is printed out for each execution. I don't think it's my call to change it.

russellwheatley:
Just deployed on node 8 again (sanity check). Successful console logs:
Initialisation:
Screenshot 2020-07-21 at 17 51 30
Execution:
Screenshot 2020-07-21 at 17 53 43

Salakar:
Testing webhook with this comment, ignore me 🙈

dackers86:
Closing in favour of https://github.com/firebase/extensions/pull/413

i14h:
@dackers86 do you need us to close this PR?

PR #383, opened 2020-07-09
chore(*): reran npm install at root

jhuleatt:
I had to run `npm install` at root to build the npm scripts before releasing them, but it resulted in this unexpected TypeScript noise

### DIFF
--- a/firestore-bigquery-export/functions/lib/logs.js
+++ b/firestore-bigquery-export/functions/lib/logs.js
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 Object.defineProperty(exports, "__esModule", { value: true });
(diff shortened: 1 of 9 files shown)


jhuleatt:
According to https://github.com/microsoft/TypeScript/issues/38552#issuecomment-628926350 it seems like this may not matter? 

> you should not treat this as an invariant

jhuleatt:
All these `void 0` assignments are later reassigned to actual values, so I don't think this affects functionality

jhuleatt:
this might be caused by an old `package-lock.json` on my machine

jhuleatt:
@tinaliang also ran into this, and solved it by deleting the `node_modules` folder in each extension's directory

PR #354, opened 2020-06-16
feat(firestore-bigquery-export): support collectionGroup queries

russellwheatley:
- Created a new parameter which allows the user to decide if the imported Firestore documents are queried via a `collection` query or a `collectionGroup` query.
- Doesn't fully solve #17 as wildcards aren't allowed in Firestore queries. 
- Updated recently merged `non-interactive` option to also accept `--query-collection-group` parameter. This feature can now be used interactively or non-interactively.

## Testing

Tested the import of a collection with 30 documents using a `collection` query:
Screenshot 2020-06-29 at 12 06 22

Tested the import of a sub-collection with 2 documents using a `collectionGroup` query:
Screenshot 2020-06-29 at 12 06 32

I've also added the `collectionGroup` query as an argument to be used during `--non-interactive` mode: https://github.com/firebase/extensions/pull/354/files#diff-2f74e55e61e11ab08c67d590ea86116bR82

Screenshot 2020-07-14 at 10 52 13

PR #385, opened 2020-07-13
Update index.ts

iampujan:
changing collection to collectionGroup. this will now support for sub collections as well.


iampujan:
@googlebot I signed it!

russellwheatley:
Hi @iampujan, thanks for the PR. There is already a PR proposed for allowing collection group queries as part of the import script: https://github.com/firebase/extensions/pull/354

PR #394, opened 2020-07-22
Add upgrade info to counter extension

kevinthecheung:
You might have already been doing this, but if not, here's a pass at it.


laurenzlong:
@Ehesp I just noticed that the "billingRequired" field is set to false, it should be changed to true when we make this change.

PR #398, opened 2020-07-29
fix(firestore-bigquery-export): Fix document id bug when upgrading BQ extension version

russellwheatley:
## The problem
When a user upgrades the extension, the latest`@firebaseextensions/firestore-bigquery-change-tracker` package is installed which is expecting tables with the `document_id` column. Tables that do not have the `document_id` column are causing the extension to crash.

## Solution
Update the `@firebaseextensions/firestore-bigquery-change-tracker` package to check if existing tables (i.e `posts_raw_changelog` & `posts_raw_latest`) have a `document_id` column and update if they do not. Thus making it backwards compatible.

## How it was tested
1. Checked out version `0.1.5` of `firestore-bigquery-export` extension. 
2. Installed with `@firebaseextensions/firestore-bigquery-change-tracker` locked to version `1.1.6`. Below is a screenshot of a row in the BigQuery changelog table after adding a document  to Firestore. Note that it does not yet contain a `document_id` column:
Screenshot 2020-07-29 at 14 07 18
3. Uninstall extension whilst keeping existing tables `posts_raw_changelog` & `posts_raw_latest` in BigQuery.
4. Installed latest version with PR changes to `@firebaseextensions/firestore-bigquery-change-tracker` package. After adding a new document to Firestore the below now shows the same table with a `document_id` column automatically added to the schema:
Screenshot 2020-07-29 at 14 32 11
5. Added another document with latest `document_id` column in place to check behaviour when the column already exists:
Screenshot 2020-07-29 at 14 34 38
6. A screenshot of the `posts_raw_latest` view also updated with a `document_id` column:
Screenshot 2020-07-29 at 15 36 09
7. The logs of the function. Note that there are 4 console logs prefixed with "qqqqq" which I used to check behaviour which are not part of this PR:
Screenshot 2020-07-29 at 14 34 12


jhuleatt:
I tried to verify this fix by running `@firebaseextensions/firestore-bigquery-change-tracker` 1.1.7 to create some tables without `document_id`, then 1.1.9 to repro the error, and then this PR's fix . However, I found a bug where the *first* entry seems to fail, but then all subsequent entries work. 

The first time it ran with this PR's fix, it logged that it added the `document_id` column successfully:

```
>  BigQuery table with name posts_raw_changelog already exists in dataset firestore_export!
>  Updated 'posts_raw_changelog' table with a 'document_id' column
>  View with id posts_raw_latest already exists in dataset firestore_export.
>  Updated 'posts_raw_latest' table with a 'document_id' column
```

But, when it tried to actually insert the record:

```
>  Inserting 1 row(s) of data into BigQuery
>  Error when mirroring data to BigQuery PartialFailureError
>      at /Users/jhuleatt/git/extensions/firestore-bigquery-export/node_modules/@google-cloud/bigquery/build/src/table.js:1548:23
>      at Util.handleResp (/Users/jhuleatt/git/extensions/firestore-bigquery-export/node_modules/@google-cloud/bigquery/node_modules/@google-cloud/common/build/src/util.js:141:9)
>      at /Users/jhuleatt/git/extensions/firestore-bigquery-export/node_modules/@google-cloud/bigquery/node_modules/@google-cloud/common/build/src/util.js:422:22
>      at onResponse (/Users/jhuleatt/git/extensions/firestore-bigquery-export/node_modules/retry-request/index.js:206:7)
>      at /Users/jhuleatt/git/extensions/firestore-bigquery-export/node_modules/@google-cloud/bigquery/node_modules/teeny-request/build/src/index.js:222:13
>      at processTicksAndRejections (internal/process/task_queues.js:97:5) {
>    errors: [ { errors: [Array], row: [Object] } ],
>    response: {
>      kind: 'bigquery#tableDataInsertAllResponse',
>      insertErrors: [ [Object] ]
>    }
>  }
```

However, for any record I added to Firestore/modified after that, it worked fine:

```
>  BigQuery dataset already exists: firestore_export
>  BigQuery table with name posts_raw_changelog already exists in dataset firestore_export!
>  View with id posts_raw_latest already exists in dataset firestore_export.
>  Inserting 1 row(s) of data into BigQuery
>  Inserted 1 row(s) of data into BigQuery
>  Completed execution of extension
```
 
So it seems like the script in its current form might cause the extension to lose the first changelog it tries to record when updating a table that's missing `document_id`. This appears to be different behavior than what happened in your logs - are you able to reproduce this issue, or might it be something I'm doing wrong when testing?

jhuleatt:
Apparently this is a [known issue](https://issuetracker.google.com/35905247) with BigQuery. The BigQuery team condones the workarounds found in [this Stack Overflow answer](https://stackoverflow.com/a/25292028/4816918). 

I think the "wait 2 minutes" solution would be tricky for us to implement since our updates happen in parallel and one instance of the function could update the schema while another instance may then check the schema within 2 minutes, see that the schema already contains document_id, and do the insert without knowing that it should wait. 

The `ignoreUnknownValues` option might be the best solution. The older rows in the table are missing the `document_id` value anyway, so it wouldn't be too disruptive. The only dangerous thing I can think of is that if we introduce a different bug in the future that we don't catch because `ignoreUnknownValues` may hide it (for example, if we change a column name by accident). Maybe we can place the data insert in a `try`/`catch` block, and in the `catch` block retry with `ignoreUnknownValues` only if it failed on `document_id` not found? Or is that being overcautious?

huangjeff5:
+1 to running `ignoreUnknownValues` only on `document_id` not found.

I feel like this is a tricky case to recognize. If I was a new developer to this repo working on the extension and I needed to add a new field, I might think that it's ok to add another column directly to schema.ts. I would rather have it yell at me while testing, rather than have it silently fail and realize there was an issue later on.

Issue #410, opened 2020-08-05
Add regex `^[^/]+(/[^/]+/[^/]+)*$` to all Firestore collection path params


laurenzlong:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

### [REQUIRED] Step 2: Extension name

All Firestore extensions

### What feature would you like to see?

Firestore collections must be an odd number of segments, otherwise the extension will fail with an installation error like https://github.com/stripe/stripe-firebase-extensions/issues/28#issuecomment-669147387. Should add this regex to all Firestore collection params: https://github.com/firebase/extensions/blob/7469c4a1181e393074fdc477ba67c6145346ac52/firestore-translate-text/extension.yaml#L110. 

Also the validation error message should be something more descriptive like `Firestore collection paths must be an odd number of segments separated by slashes, e.g. "path/to/collection"` 

### How would you use it?

Tell us how you'd use this feature in your app.

PR #418, opened 2020-08-10
feat(firestore-bigquery-export): Add validation regex for collection path parameter.

russellwheatley:
fixes https://github.com/firebase/extensions/issues/410

- `firestore-bigquery-export` was the only extension without a regex for its Firestore collection path param.

Attempts at odd number of path segments:
Screenshot 2020-08-10 at 14 24 32


i14h:
@laurenzlong is this the same as the bug you identified in #437?

laurenzlong:
No this is unrelated.

PR #419, opened 2020-08-10
feat(firestore-send-email): adds support for Handlebars partials

### LINKED ISSUE #372: Trigger Email HandleBar Partials

firestore-send-email

Is there any support for handlebar partials in firestore-send-email? If not, is that something on the roadmap because there's no easy to reuse my code or create shared templates right now.

### How would you use it?

Create shared templates!

mbleigh:
This adds support for arbitrary partials for reusable pieces of templated emails. The mechanism is a new `{partial: true}` field on any document in the `TEMPLATES_COLLECTION`.

The neat (if I do say so myself) feature of this is that partials are unique per-environment, so you can have a single partial that renders differently depending on whether the template is for `html`, `text`, or `amp`.

Partials have the same data context as the template itself, so it is possible to inject custom data into a partial.

Fixes #372


russellwheatley:
@googlebot I consent

dackers86:
### Testing

## Basic Email

image

## Missing template name example

image

## Template Not found example

image

## Working template example

image

image


## AMP example

image

image

dackers86:
Hi @mbleigh.

I've ran through a few test cases above. Have you experienced any issues with AMP? GMail and Hotmail either show an error like above or fail to show any detail at all?

mbleigh:
@dackers86 I haven't really played around with the AMP stuff much, when I originally built the library I just added it because the underlying `nodemailer` library supported it so I passed it through.

Does the testing above have anything to do with this PR (adding support for partials)?

dackers86:
I suppose this mostly covers edge cases.  Happy to edit to cover just partials and potentially raise a separate issue for AMP if the problem persists?

dackers86 (approved):

jhuleatt (approved):

PR #421, opened 2020-08-14
feat(firestore-translate): text multiple fields update

dackers86:
### Testing

Single
image

Multiple

image


russellwheatley (changes requested):
I think this [function](https://github.com/firebase/extensions/blob/961d248fcc15fc6511d8e57f7116b199d30b1a07/firestore-translate-text/functions/src/index.ts#L119) needs to be updated to reflect there could be mutliple fields. For instance, it is checking `const inputHasChanged = inputAfter !== inputBefore;` when they could be two object references, and thus never the same value.

firestore-translate-text/functions/src/index.ts
```diff
@@ -173,13 +194,58 @@ const translateDocument = async (
...
+    if (fn) await fn();
+  }
+
+  return updateTranslations(snapshot, translations);
+};
+
+const translateDocument = async (
+  snapshot: admin.firestore.DocumentSnapshot
+): Promise<void> => {
+  const input: string = extractInput(snapshot);
```
Ehesp, line 239: Not specifically a `string` now.

firestore-translate-text/functions/src/index.ts
```diff
@@ -173,13 +194,58 @@ const translateDocument = async (
...
+
+  if (typeof input === "object") {
+    await translateMultiple(input, snapshot);
+    return Promise.resolve();
```
Ehesp, line 243: Can just `return` here, since it's already a promise, rather than creating a new instance.

Ehesp (changes requested):

firestore-translate-text/functions/src/index.ts
```diff
@@ -118,36 +118,56 @@ const handleUpdateDocument = async (
...
+  if (typeof inputBefore !== typeof inputAfter) {
     logs.documentUpdatedChangedInput();
     await translateDocument(after);
-  } else if (inputBefore) {
-    logs.documentUpdatedDeletedInput();
-    await updateTranslations(after, admin.firestore.FieldValue.delete());
+    return;
+  }
+
+  let beforeValues = [];
```
russellwheatley, line 144: I think you can remove lines 144-157. Stringifying the input & checking for equality should be enough. Update line 159. i.e. `JSON.stringify(inputBefore) === JSON.stringify(inputAfter)`

russellwheatley (changes requested):

firestore-translate-text/functions/src/index.ts
dackers86: I think you're right here. Will update and retest

russellwheatley:
LGTM.

jhuleatt (approved):

KeisukeNagakawa:
Hi! Is this functinoality available in extension?  I've tried like this but didn't work(this caused field name of "content,title" as the target field to translate. not "content" and "title" ). 
スクリーンショット 2022-03-07 15 50 47

PR #424, opened 2020-08-18
August 20, 2020 Release

laurenzlong:


laurenzlong:
The linter appears to be failing, however both `npm run format` and `npm test` passes locally. So it may be something funky with travis.

PR #451, opened 2020-09-15
fix(release): use placeholder changelog entry when changelog file does not exist

Salakar:
Fixes an issue where if an extension does not have a `CHANGELOG.md` file the script would error and exit ([example](https://github.com/firebase/extensions/runs/1097667905#step:3:12)).

image

Additionally made the following changes (see individual commits):

 - formatted the script
 - removed temporary version skipping logic as it's now no longer required as each extension has since had a release.

### Testing

Tested all 3 scenarios (release already exists, changelog could not be found, new release):

image

Testing releases can be seen here: https://github.com/invertase/extensions-release-testing/releases

image

image

### DIFF
--- a/.github/workflows/scripts/release.sh
+++ b/.github/workflows/scripts/release.sh
@@ -167,7 +167,7 @@ for i in $(find . -type f -name 'extension.yaml' -exec dirname {} \; | sort -u);
 
   changelog_contents="No changelog found for this version."
 
-  # Ensure changelog exists exists
+  # Ensure changelog exists
   if [ -f "$i/CHANGELOG.md" ]; then
     # Pluck out change log contents for the latest extension version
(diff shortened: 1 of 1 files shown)


dackers86 (approved):

.github/workflows/scripts/release.sh
```diff
@@ -158,37 +158,26 @@ if [[ -z "$GITHUB_REPOSITORY" ]]; then
...
   extension_version=$(awk '/^version: /' "$i/extension.yaml" | sed "s/version: //")
 
-  # TODO(salakar): Remove once this is no longer outputting logs.
-  case "${SKIP_EXTENSION_VERSIONS[@]}" in  *"$extension_name $extension_version"*)
-    echo "Skipping $extension_name version $extension_version"
-    continue
-  esac
+  changelog_contents="No changelog found for this version."
+
+  # Ensure changelog exists exists
```
rosariopfernandes, line 170: Seems like a minor typo on this comment

joehan, line 170:
```suggestion
  # Ensure changelog exists
```

PR #454, opened 2020-09-22
Sept 24 2020 Release

joehan:
It looks like Author was removed from a bunch of these READMEs - is this expected?


jhuleatt:
> It looks like Author was removed from a bunch of these READMEs - is this expected?

This was added to the Firebase CLI in this PR: https://github.com/firebase/firebase-tools/pull/2393

Any chance you have an old CLI version (or at least an old one that's in your npm deps in your local extensions directory)?

PR #455, opened 2020-09-23
Changelog for sept 24 release

joehan:


kevinthecheung (approved):

jhuleatt (approved):
I was thinking of adding a note about the script update, but think we'll leave that out since it doesn't affect the extension itself.

PR #405, opened 2020-07-31
Add image content type option to convert uploaded images to GCS/Firestore

moog16:
This PR is to add an extension config option to also convert the uploaded image to a different image content-type (eg. jpg, png). Supported types are: jpg, png, webp, tiff, raw.

The file-extension and the content-type will be converted. If a file-extension is not provided, the function will not rename the new modified/generated file with a new extension. 

### Examples:
1. 
filename: `my-image.png`
content-type: `image/png`

convert to `jpg` & 200x200

modified filename: `my-image_200x200.jpg` 
modified content-type: `image/jpeg`

2. 
filename: `my-image`
content-type: `image/png`

convert to `jpg` & 200x200

modified filename: `my-image_200x200` 
modified content-type: `image/jpeg`

PR #465, opened 2020-10-06
Revert "Add image content type option to convert uploaded images to GCS/Firestore"

tinaliang:
Reverts firebase/extensions#405

Although tests pass on Travis, they were failing on my local machine. I believe the root cause is https://github.com/Automattic/node-canvas/issues/1386


jhuleatt (approved):
LGTM. Let's hold off on releasing this until we have a chance to investigate why the tests were flaky.

PR #239, opened 2020-03-24
test(firestore-bigquery-export): initial pass at tests

dackers86:
WIP - will update when ready for review

PR #469, opened 2020-10-16
test(firestore-bigquery-export): Test script 

russellwheatley:
Run `install-test.sh` script which does the following:
1. Links `firestore-bigquery-change-tracker` package with extension.
2. Runs `npm i` from root to install dependencies and compile.
3. Installs latest BigQuery extension from local project. 
5. Inserts data into Firestore collection.
6. Prints out latest BigQuery extension cloud function logs resulting from data insertion. 
  
To test the PR, I emulated the bug with the `document_id` column. I installed  BigQuery extension `0.1.5` (i.e without `document_id` changes) and ran some data into the the tables.

I then ran the `install-test.sh` against BigQuery version `0.1.6` (i.e. with `document_id` changes). Here is the console output from the script;
Screenshot 2020-10-16 at 17 48 20

The above highlighted log is the error from the `document_id` feature update.

## Notes
- You will have to manually input "yes" twice when prompted for billing requirements & permissions [here](https://github.com/firebase/firebase-tools/blob/master/src/commands/ext-install.ts#L61) in Firebase CLI. This is the only other manual step once you run the script. Ideally, we would be able to use a rest API, but there isn't one publicly available to my knowledge. 
- I updated the test coverage for `firestore-bigquery-export` using some of a previous [PR](https://github.com/firebase/extensions/pull/239) courtesy of @dackers86 
- moved the package.json into the functions folder to keep the same directory structure as the other extensions.
- the test extension will always be the default instance id (i.e. `firestore-bigquery-export`)


package.json
```diff
@@ -7,7 +7,7 @@
     "format": "prettier --write \"**/*.{js,md,yml,ts,json,yaml}\"",
     "lint": "prettier --list-different \"**/*.{js,md,yml,ts,json,yaml}\"",
     "clean": "lerna run --parallel clean",
-    "build": "lerna run --parallel build",
+    "build": "lerna run build",
```
dackers86, line 10: Isn't this needed?

dackers86 (approved):

package.json
russellwheatley: Ran in parallel throws an error in one of the big query packages.  It makes builds quicker, but it isn't essential. It's a strange error as it's trying to compile before the `tsc` binary is installed I think. It might be a problem with the structure of the packages (i.e packages within packages).

Salakar (approved):

jhuleatt (approved):

PR #479, opened 2020-10-27
feat(storage-resize-images): delete images on success

### LINKED ISSUE #470: Should storage-resize-images delete the original image if the resize fails?

This feature request is for extension: **`storage-resize-images`**

Right now, if the `DELETE_ORIGINAL_FILE` param is set to `true`, `storage-resize-images` will delete the original image it is resizing *even if the resize operation fails*. We should consider *not* deleting the original image if the resize fails.

#### Pros of this change

If the resize operation fails, at least the original image still exists, so it could potentially be used in place of a resized image.

#### Cons of this change

It's a breaking change - people that use the extension currently may expect and want the original images to be deleted no matter what.

russellwheatley:
fixes #470 

Use enum to check which delete option user selected & deletes appropriately.

Extension executing with always delete enum (0): 
Screenshot 2020-10-27 at 13 16 13
Extension executing with never delete enum (1): 
Screenshot 2020-10-27 at 13 18 17
Extension executing with on success enum (2):
Screenshot 2020-10-27 at 13 21 24


dackers86 (approved):

Salakar (approved):

firestore-bigquery-export/functions/package.json
```diff
@@ -17,7 +17,7 @@
     "@google-cloud/bigquery": "^2.1.0",
     "@types/chai": "^4.1.6",
     "chai": "^4.2.0",
-    "firebase-admin": "^9.1.0",
+    "firebase-admin": "^8.1.0",
```
jhuleatt, line 20: @russellwheatley is this to undo the type issues we talked about in the meeting today?

russellwheatley: Yes, it was causing problems in the BQ tests.

jhuleatt: mind moving this and the firestore-bigquery-change-tracker change into another PR to make it a little easier to track?

russellwheatley: Yeah, that makes sense. Will do 😄

storage-resize-images/functions/src/resize-image.ts
```diff
@@ -0,0 +1,181 @@
...
+          logs.remoteFileDeleting(filePath);
+          await remoteFile.delete();
+          logs.remoteFileDeleted(filePath);
+        } catch (err) {
+          logs.errorDeleting(err);
+        }
+      }
+    }
+
+    return { size, success: true };
```
jhuleatt, line 165: since we pass the status back anyway, what do you think about doing the delete back in index.ts, just so the two deletes (`always` and `onSuccess`) are closer to each other in the code?

russellwheatley:
@jhuleatt I've made those updates as requested. I think we ought to have enough test coverage now for `storage-resize-images`. I think this PR has brought the coverage down as it wasn't being tested [previously](https://github.com/firebase/extensions/pull/479/files#diff-a5e14eef1b6c57b59a8c4e65a4d06ffbd2c9109f0c8ce2a839e3ca0e5a8e0472L11).

To clarify why there's a lot of changes in this PR - I'm trying to break up the functionality so we are able to test them.  I think the next step would be to test the actual `modifyImage` function. 

These tests will fail due to the `firestore-bigquery-export` extension dependency `firebase-admin`. This PR ought to be merged in first: https://github.com/firebase/extensions/pull/492 to fix that issue.

jhuleatt:
CI is only failing on Node 8. Since we're planning on removing the Node 8 tests anyway (https://github.com/firebase/extensions/issues/491), I'm going to merge this.

Issue #331, opened 2020-05-28
chore(*): within all extension.yaml files, change syntax of parameter ref to ${param:PARAMETER_NAME}


rachelsaunders:
Currently, all the official Firebase extensions in their extension.yaml files use the parameter reference syntax of ${PARAMETER_NAME}.  However, with the recent [v8.4.0 release](https://github.com/firebase/firebase-tools/releases/tag/v8.4.0), we can/should use the "universal" syntax of ${param:PARAMETER_NAME} in extension.yaml files.

"universal" = all extension files now support this parameter reference syntax (and it's the preferred syntax :-)

Ehesp:
@rachelsaunders does this also apply to built in parameters (such as `PROJECT_ID`)?

rachelsaunders:
Yep, it's my understanding that it's _any_ parameter, both user-configured and auto-populated.  We can def double-check, though @laurenzlong

karayu:
Let's keep this change unmerged until we get more soak time in the CLI. The CLI supporting this released about a month ago.

karayu:
Assigning to Kevin for FYI/tracking

PR #486, opened 2020-11-05
chore(storage-resize-images): generate release notes

jhuleatt:


storage-resize-images/extension.yaml
```diff
@@ -126,7 +126,7 @@ params:
     example: my-project-12345.appspot.com
     validationRegex: ^([0-9a-z_.-]*)$
     validationErrorMessage: Invalid storage bucket
-    default: ${param:STORAGE_BUCKET}
+    default: ${STORAGE_BUCKET}
```
rosariopfernandes, line 129: Is the new param syntax causing problems when generating release notes? Because #331 states that it should be the preferred syntax.

jhuleatt: Hi @rosariopfernandes! It actually caused a problem in the install flow:

7nQZCVSykEmWFjM

I'm not sure if that's a console bug, or a backend bug, so I'm just reverting back to `${STORAGE_BUCKET}` and will follow up with the rest of the extensions team to fix this going forward.

rosariopfernandes: @jhuleatt Ah, I see. Thanks for clarifying!

storage-resize-images/extension.yaml
```diff
@@ -196,4 +196,5 @@ params:
         value: tiff
       - label: Do not convert
         value: false
+    default: false
```
jhuleatt, line 199: Need this so that the select box isn't empty when installing

kevinthecheung (approved):

PR #488, opened 2020-11-10
feat(firestore-counter): ios swift client

russellwheatley:
iOS Swift client.


russellwheatley:
Hey @kevinthecheung, do you mind taking a look at the `README.md`, please?  😄

firestore-counter/clients/ios/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+class ViewController: UIViewController {
+    var controller = FirestoreShardCounter(docRef: Firestore.firestore().collection("pages").document("hello-world"), field: "visits")
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // event listener which returns total amount
+        controller.onSnapshot { (value, error) in
+          if let error = error {
+            // handle error
```
dackers86, line 21: Is there anything needed for executing on these comments?

firestore-counter/clients/ios/README.md
```diff
@@ -0,0 +1,45 @@
...
+            // handle value
+          }
+      }
+    }
+
+    @IBAction func getLatest(_ sender: Any) {
+      // get current total
+        controller.get(){ (sum, error) in
+          if let error = error{
+            // handle error
```
dackers86, line 32: Same here?

firestore-counter/clients/ios/README.md
russellwheatley: No, just to direct User using package to handle error or value.

firestore-counter/clients/ios/README.md
russellwheatley: Same

Salakar (approved):

jhuleatt:
@peterfriese would you be able to take a look at this? It's a client sample for the firestore-sharded-counter extension, which currently only has a js client sample.

firestore-counter/clients/web/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2018",
```
peterfriese, line 3: Unrelated to the Swift port.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+    for path in shards.keys {
+      firestore.document(path).addSnapshotListener { (snapshot, error) in
+
+        if let error = error {
+          print("FirestoreCounter: Error listening to document changes: \(error)")
+
+          observer(nil, error)
+        } else {
+          if let snapshotValue = snapshot?.get(self.field) {
+            let doubleVal = snapshotValue as! Double
```
peterfriese, line 77: Try to avoid force unwrapping if at all possible.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+            var sum = Double(0)
+            self.shards[path] = doubleVal
+
+            for value in self.shards.values {
```
peterfriese, line 81: This can we written in a more functional way.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+              sum += value
+            }
+
+            observer(sum, nil)
+          }
+        }
+      }
+    }
+  }
```
peterfriese, line 90: Here is a more functional version of the methods which also has slightly better error handling:

```
  public func addSnapshotListener(_ listener: @escaping (Double?, Error?) -> Void) {
    shards.keys.forEach { path in
      firestore.document(path).addSnapshotListener { snapshot, error in
        if let error = error {
          print("FirestoreCounter: Error listening to document changes: \(error)")
          listener(nil, error)
        } else {
          guard let snapshot = snapshot,
                let snapshotValue = snapshot.get(self.field),
                let doubleVal = snapshotValue as? Double else {
            listener(nil, error)
            return
          }
          self.shards[path] = doubleVal
          let sum = self.shards.reduce(0.0) { $0 + $1.value }
          listener(sum, nil)
        }
      }
    }
  }
```

peterfriese: Please note that this version doesn't keep track of the handles produced by the snapshot listeners.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+
+  public func get(result: @escaping (Double?, Error?) -> ()) {
+    var documentIds = [String]()
+    var totalValue = Double(0)
+
+    for key in shards.keys {
+      documentIds.append((key as NSString).lastPathComponent)
+    }
+
+    shardRef.whereField(FieldPath.documentID(), in: documentIds).getDocuments() { (snapshot, error) in
```
peterfriese, line 34: Style guide on the main iOS SDK is to remove the parens around the closure parameter list, i.e.
```
hardRef.whereField(FieldPath.documentID(), in: documentIds).getDocuments() { snapshot, error in
```

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+      if let error = error {
+        print("FirestoreCounter: Error getting documents: \(error)")
+        result(nil, error)
+      } else {
+        for queryDoc in snapshot!.documents {
```
peterfriese, line 39: Try to avoid force unwrapping, if at all possible.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+          let document = queryDoc.data()
+          let value = document[self.field] == nil ? 0 : document[self.field] as! Double;
```
peterfriese, line 41: Try to avoid force unwrapping, if at all possible. Also can get rid of the semicolon.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+    self.shardRef = docRef.collection(collectionId)
+    shards[self.docRef.path] = Double(0)
+    shards[self.shardRef.document(String(self.shardId)).path] = Double(0)
+    shards[self.shardRef.document("\t" + String(self.shardId.prefix(4))).path] = Double(0)
+    shards[self.shardRef.document("\t\t" + String(self.shardId.prefix(3))).path] = Double(0)
+    shards[self.shardRef.document("\t\t\t" + String(self.shardId.prefix(2))).path] = Double(0)
+    shards[self.shardRef.document("\t\t\t\t" + String(self.shardId.prefix(1))).path] = Double(0)
+  }
+
+  public func get(result: @escaping (Double?, Error?) -> ()) {
```
peterfriese, line 26: Use `Void` instead of `()` for return types.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+  private let shardRef: CollectionReference
+  private var docRef: DocumentReference
+  private var field: String
+
+  public init(docRef: DocumentReference, field: String) {
+    self.docRef = docRef
+    self.field = field
+    self.firestore = docRef.firestore
+    self.shardRef = docRef.collection(collectionId)
+    shards[self.docRef.path] = Double(0)
```
peterfriese, line 18: No need to use `Double(0)` - you've already told the compiler about the type in line 7.
Also don't need to use `self` here.

Replace with:
```
shards[docRef.path] = 0
```

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+    shards[self.shardRef.document(String(self.shardId)).path] = Double(0)
```
peterfriese, line 19: Can be simplified to
```
shards[shardRef.document(shardId).path] = 0
```

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+    shards[self.shardRef.document("\t" + String(self.shardId.prefix(4))).path] = Double(0)
```
peterfriese, line 20: Ditto, as well for the following lines.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
...
+  private var shards = [String: Double]()
+  private let collectionId = "_counter_shards_"
+  private let shardRef: CollectionReference
+  private var docRef: DocumentReference
+  private var field: String
+
+  public init(docRef: DocumentReference, field: String) {
+    self.docRef = docRef
+    self.field = field
+    self.firestore = docRef.firestore
```
peterfriese, line 16: No need to use `self` here.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
peterfriese, line 39: There's a more functional way to write this - see my re-implementation of `onSnapshot` for inspiration.

firestore-counter/clients/ios/Sources/FirestoreCounter/FirestoreCounter.swift
```diff
@@ -0,0 +1,91 @@
+import FirebaseFirestore
+import Firebase
+
+public class FirestoreShardCounter {
```
peterfriese, line 4: How about naming this `FirebaseDistributedCounter` - I think it's better to not expose the implementation strategy in the name. Plus, "Distributed Counter" is the extension's official name.

firestore-counter/clients/ios/README.md
```diff
@@ -0,0 +1,45 @@
...
+import FirestoreCounter
+import FirebaseFirestore
+
+class ViewController: UIViewController {
+  var controller = FirestoreShardCounter(docRef: Firestore.firestore().collection("pages").document("hello-world"), field: "visits")
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    // event listener which returns total amount
+    controller.onSnapshot { (value, error) in
```
peterfriese, line 19: Recommend using the same name for the value here and further down in line 30.

firestore-counter/clients/ios/README.md
```diff
@@ -0,0 +1,45 @@
...
+        // handle error
+      } else if let value = value {
+        // handle value
+      }
+    }
+  }
+
+  @IBAction func getLatest(_ sender: Any) {
+    // get current total
+    controller.get() { (sum, error) in
```
peterfriese, line 30: Use `value` instead of `sum`?

peterfriese (changes requested):
I've left a few comments to make this more idiomatic Swift code.

I've also noticed that the implementation of the snapshot listener seems to suffer from race conditions - it's easily possible to end up with a slightly higher value than what's actually stored in the sharded document. Not sure if this is specific to the Swift implementation or if all implementations suffer from this issue.

russellwheatley:
Thanks for the great feedback, @peterfriese. I've updated the code sample as per your recommendations. I haven't updated the `README.md` or the `Package.swift` files. I don't think we're distributing this code, we don't for either the web or the android code samples either.  What do you think, @jhuleatt?

jhuleatt:
> Thanks for the great feedback, @peterfriese. I've updated the code sample as per your recommendations. I haven't updated the `README.md` or the `Package.swift` file as my thoughts are to remove them entirely. I don't think we're distributing this code, we don't for either the web or the android code samples either. What do you think, @jhuleatt?

Distributing the `firestore-counter` client samples as packages for any platform isn't something we do right now. However, if keeping the `Package.swift` file around leaves the door open for us to do that later, it may be worth keeping it. Would it be a lot of work to get it into good enough shape to include in this PR?

I also think it would be worth it to keep the `README`. Currently, the `POSTINSTALL` for `firestore-counter` only mentions web, but we should eventually update it to reflect the fact that we have Android and iOS samples now. I've filed #520 for that.

kevinthecheung (changes requested):

russellwheatley:
i've updated the `README.md` & `swift.package` as per recommendations except instructions to install it as a Swift package. I guess we will update that part once it is distributed.

peterfriese:
Thanks for updating the code! From a user's perspective, I would certainly expect to be able to pull in this client library using SPM (as opposed to having to copy & paste the code).

kevinthecheung (approved):

jhuleatt (approved):

jhuleatt:
Let's merge this as-is, and plan a potential publish to SPM separately.

PR #492, opened 2020-11-11
chore(firestore-bigquery-export): revert firebase-admin dependency back to v8

russellwheatley:
Downgrade `firebase-admin` in `firestore-bigquery-export` extension to stop it from breaking the tests. Some types have changed and may need to be updated, or the big query test script for inserting data will have to be changed, and the `not-in` query removed.


dackers86 (approved):

russellwheatley:
I couldn't figure out how to get the types to work correctly with `firebase-admin` v9, not entirely sure if that was a local environment setup or something else. But it was blocking, so I changed the `install-script.js` for BQ test scripts instead so it was compatible with `firebase-admin` v8.

PR #498, opened 2020-11-13
@moog16 feat(storage-resize-images): add gif to supported resize image-types

moog16:


russellwheatley:
Sharp by itself does not support resizing GIFs, it requires libvips compiled with support for ImageMagick as noted [here](https://github.com/lovell/sharp/issues/2437). I'd also image the ability to change type would not work either.

moog16:
Yup you're totally right. Thanks for taking a look anyways 👍

PR #502, opened 2020-11-17
feat(storage-resize-images): support non-default cloud storage bucket

rosariopfernandes:
I'm not sure if this feature was removed by mistake or if it was never there in the first place.
The extension already prompts the user for the storage bucket they wish to use, but the this config `param` was never actually used when triggering the Cloud Function.


rosariopfernandes:
Nevermind. I just noticed the bucket is specified on the extension.yaml file:

https://github.com/firebase/extensions/blob/53ffd27b4532c429d94d652d7b6d08ccc5e76e0c/storage-resize-images/extension.yaml#L65

PR #504, opened 2020-11-17
feat(rtdb-limit-child-nodes) Adds an option to select an alternative database instance

dackers86:
Database configuration options updated to included selected database instance. Allow an alternative instance to be targeted for the installed extension

## Testing

- [x] Default Instance updated as expected
- [x] Custom Instance as expected
- [x] Only specified db instance is updated as expected

### Default Database Instance

image

image

### Custom Database Instance

image

image

image


Salakar (approved):

rtdb-limit-child-nodes/extension.yaml
```diff
@@ -120,6 +120,18 @@ params:
     validationErrorMessage: Path cannot have spaces.
     required: true
 
+  - param: SELECTED_DATABASE_INSTANCE
+    label: Selected Realtime Database instance
+    description: >
+      What is the selected Realtime Database used for this extension?
```
kevinthecheung, line 126:
```suggestion
    label: Realtime Database instance
    description: >
      Which Realtime Database instance do you want to limit?
```

kevinthecheung (changes requested):

rtdb-limit-child-nodes/extension.yaml
dackers86: Thanks @kevinthecheung. I have updated the label and description. This matches updated added for https://github.com/firebase/extensions/pull/505

PR #505, opened 2020-11-17
feat(delete-user-data) Adds an option to select an alternative database instance

dackers86:
Allows an alternative instance to be targeted for the installed extension.

## Testing

- [x] Default Instance updated as expected
- [x] Custom Instance as expected
- [x] Only specified db instance is updated as expected

### Default Database Instance

image

image

image

### Custom Database Instance

image

image

image


Salakar (approved):

delete-user-data/extension.yaml
```diff
@@ -148,6 +148,17 @@ params:
     default: shallow
     required: true
 
+  - param: SELECTED_DATABASE_INSTANCE
+    label: Selected Realtime Database instance
+    description: >
+      What is the selected Realtime Database used for this extension?
```
kevinthecheung, line 154: I'm not sure I understand the "selected" part. This param is used to select the instance, right?

How about:

```suggestion
  - param: SELECTED_DATABASE_INSTANCE
    label: Realtime Database instance
    description: >
      From which Realtime Database instance do you want to delete user data?
```

kevinthecheung (changes requested):

PR #514, opened 2020-11-24
fix(firestore-bigquery-export): updated to accept a comma separated schema for Cli flags when generating views

dackers86:
Cli for generating schema views incorrectly parses schemas when using flags to input data.

This causes the Cli to ask the user to input manually, whereas this fix will now allow flags to be used again.


dackers86:
Further testing proves this feature is already working. Closing PR

Issue #422, opened 2020-08-16
Templates should handle attachments


somejeff:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

👍 

### [REQUIRED] Step 2: Extension name

This feature request is for extension: firestore-send-email

### What feature would you like to see?

It would be great it `templates` could handle `attachments`.

Currently, the `template.ts` returns a `message` object that only includes:
``` javascript
{
  subject: 'Hello {{username}}',
  html: '<b>Welcome {{name}}</b>',
  text: '...',
  amp: ' ... '
}
```

It could also include an array of attachments (in the form of https://nodemailer.com/message/attachments/):
``` javascript
{
  subject: 'Hello {{username}}',
  ...
  attachments: [
    {
      filename: 'Invoice {{invoice_number}}.pdf',
      content: '{{invoice_base64_content}}',
      encoding: 'base64'
  ]
}
```
Then it would use the `data` attribute as usual:
``` javascript
admin.firestore().collection('${param:MAIL_COLLECTION}').add({
  toUids: ['abc123'],
  template: {
    name: 'invoice',
    data: {
      username: 'ada',
      name: 'Ada Lovelace',
      invoice_number: 'Q2020-0001',
      invoice_base64_content: ' ... base 64 encoded binary here ... '
    }
  }
})
```

The result would be an email sent with substituted subject, email contents **and attachments**

### How would you use it?

This would allow us to create templates that have attachments, such as quotes, invoices, progress reports.   
Tools such as html-pdf could be used to create the PDFs
Or URIs,  pretty much anything from https://nodemailer.com/message/attachments/ except filesystem.

jhuleatt:
Thanks for the suggestion @somejeff! 

This is a bit tricky because base64 encoding violates the [Firestore documents should be small](https://firebase.google.com/docs/firestore/best-practices#realtime_updates) rule. However, it looks like NodeMailer allows the `path` to be a url, so we could change the firestore document to something like this so we support a Firebase Storage path instead:

```js
admin
  .firestore()
  .collection("${param:MAIL_COLLECTION}")
  .add({
    toUids: ["abc123"],
    template: {
      name: "invoice",
      data: {
        username: "ada",
        name: "Ada Lovelace",
[...]

PR #519, opened 2020-11-30
feat(firestore-send-email): template attachments

russellwheatley:
feature request #422.

Allow attachments within template records to be interpolated with data values passed into a new record in the mail collection. 

### Notes
1. Partials are not supported, there is a console log to warn users.
2. I added the [strict](https://handlebarsjs.com/api-reference/compilation.html#handlebars-compile-template-options) option to attachment compiling. My concern is with crucial properties missing for attachments, the email could be sent without the attachment. Not sure if this is good UX. Happy to remove if this is not warranted.

Here are the logs of it working fine:
Screenshot 2020-11-30 at 16 21 52

The template tested with 2 attachments:
Screenshot 2020-11-30 at 16 17 47

The data inserted into `mail` collection:
```js
{
    to: "russell@invertase.io",
    template: {
      name: "test_template",
      data: {
        name: "Russ",
        filename: "data",
        content_var: "World!",
        prop: "cool_music",
        prop2: "http://path-to-mp3/track.mp3",
      },
    },
  }
```

The email received:
Screenshot 2020-11-30 at 16 19 06


dackers86 (approved):

firestore-send-email/POSTINSTALL.md
```diff
@@ -79,13 +79,20 @@ The template document can include any of the following fields:
 - **text:** A template string for the plaintext content of the email.
 - **html:** A template string for the HTML content of the email.
 - **amp:** A template string for the [AMP4EMAIL][amp4email] content of the email.
+- **attachments:** An array containing attachment(s) with template strings as values; [Nodemailer options](https://nodemailer.com/message/attachments/) supported: utf-8 string, custom content type, URL, encoded string, data URI, and pregenerated MIME node (be aware that your email has no access to the cloud server's file system).
```
Salakar, line 82:
```suggestion
- **attachments:** An array of attachments with template strings as values; [Nodemailer options](https://nodemailer.com/message/attachments/) supported: utf-8 string, custom content type, URL, encoded string, data URI, and pre-generated MIME node (be aware that your email has no access to the cloud server's file system).
```

Salakar (approved):

firestore-send-email/functions/src/logs.ts
```diff
@@ -81,6 +81,14 @@ export function missingUids(uids: string[]) {
   );
 }
 
+export function noPartialAttachmentSupport() {
+  logger.log("partial attachments are not handled and will be ignored");
```
jhuleatt, line 85:
```suggestion
  logger.warn("partial attachments are not handled and will be ignored");
```

Let's `warn` here, since it means someone misunderstood how to use the extension.

firestore-send-email/functions/src/logs.ts
```diff
@@ -81,6 +81,14 @@ export function missingUids(uids: string[]) {
   );
 }
 
+export function noPartialAttachmentSupport() {
```
jhuleatt, line 84: Can this function take some additional args so that we can provide more context to the developer? That way, instead of `"partial attachments are not handled and will be ignored"`, we could do something like:

```js
logger.warn(`Ignoring partial attachment for templace "${docId}". The firestore-send-email extension does not support partial attachments.`)
```

jhuleatt: Let's follow up with this in another release

jhuleatt (approved):

PR #527, opened 2020-12-07
test(firestore-send-email): use emulator

russellwheatley:
Initial setup for `firestore-send-email` emulator tests.


russellwheatley:
Hey @jhuleatt, as the mail attachments [PR](https://github.com/firebase/extensions/pull/519) was failing due to a 0.28% decrease in test coverage, I thought I'd implement an emulator test for the `firestore-send-email` extension. The Firestore emulator doesn't appear to suffer from the same issue as the RTDB emulator. 

Understandably, the tests fail as there are no credentials in place to use the `firebase-admin` module to add data:
Screenshot 2020-12-08 at 13 32 34

 I think we might need to set the service account JSON key to the `FIREBASE_CONFIG` environment variable in the github workflow. This means we can then initialise the admin sdk without passing in any params as stated [here](https://firebase.google.com/docs/admin/setup#initialize-without-parameters).

Do you have any thoughts, @Salakar?

jhuleatt:
As discussed in our meeting today - will the emulator tests run & code coverage tool detect the tests if you [import the CLI as a module](https://github.com/firebase/firebase-tools#using-as-a-module)?

russellwheatley:
Coverage still isn't picked up when running emulator programmatically. We have found an issue thread that clears up the mystery, and a somewhat hacky way to fix the [problem](https://github.com/facebook/jest/issues/3190#issuecomment-354758036). What do you think, @jhuleatt? Try and implement?

jhuleatt:
> Coverage still isn't picked up when running emulator programmatically. We have found an issue thread that clears up the mystery, and a somewhat hacky way to fix the [problem](https://github.com/facebook/jest/issues/3190#issuecomment-354758036). What do you think, @jhuleatt? Try and implement?

Yeah, please give that a try. If it ends up taking a long time to implement or seems like it will be flaky, then we can give up and get rid of code coverage.

dackers86:
After trying to implement last week, I feel we may need to remove the coverage if it isn't useful!

Even with the above additions a minimum amount of coverage improved.

image

dackers86:
Additionally, using the emulator provides an issue similar the problem discussed here https://stackoverflow.com/questions/65907608/firebase-emulator-leaking-when-using-with-jest. Has anyone seen a possible solution to this?

russellwheatley:
After Tuesday's discussion, we've decided to remove the codecov check from the CI as it isn't feasible (or requires too much work and hacking) to get the emulator process recognised by our test suite. Please merge in PR #586 first as described in this issue #587.

jhuleatt:
@russellwheatley do we still need to resolve the permissions issue here? I thought we'd be able to get away without setting up admin credentials because the emulator docs [state that they aren't required for RTDB/Firestore triggers](https://firebase.google.com/docs/functions/local-emulator#set_up_admin_credentials_optional).

firestore-send-email/functions/src/index.ts
```diff
@@ -340,7 +363,7 @@ async function processWrite(change) {
 
 export const processQueue = functions.handler.firestore.document.onWrite(
   async (change) => {
-    initialize();
+    await initialize();
```
jhuleatt, line 366: did we have a race condition here previously?

russellwheatley:
@jhuleatt  No permissions are needed for the emulator.

firestore-send-email/functions/src/index.ts
russellwheatley: Not a race condition, but it needs to be async for the integration test to work [here](https://github.com/firebase/extensions/pull/527/files/7f769bd1e0d33dcd18ddf122a609d98f9bfef9b8#diff-f36d0ed1929c306fadc22ba963e485ebcc96248da84b6b0d57cc1df7c0b37f3bR48-R67)

jhuleatt:
> Additionally, using the emulator provides an issue similar the problem discussed here https://stackoverflow.com/questions/65907608/firebase-emulator-leaking-when-using-with-jest. Has anyone seen a possible solution to this?

Yes, I ran into this with ReactFire. I ended up always running Jest with `--detectOpenHandles --forceExit`

dackers86 (approved):

Salakar (approved):

jhuleatt:
@firebase/invertase this PR is still marked as draft - is it ready to merge, or should we hold off?

russellwheatley:
@jhuleatt It should be good to go now.

jhuleatt (approved):

PR #70, opened 2019-10-24
feat(firestore-shorten-urls-dynamic-links): introduce Dynamic Links URL shortening extension

kevinthecheung:

PR #540, opened 2020-12-17
Delete firestore-shorten-urls-dynamic-links

jhuleatt:
This extension was in [open alpha](https://github.com/firebase/extensions/pull/70#issuecomment-678594272). It has now been moved to the experimental extensions repo: https://github.com/FirebaseExtended/experimental-extensions/tree/next/firestore-shorten-urls-dynamic-links


Salakar (approved):
LGTM, could also be removed from here: https://github.com/firebase/extensions/blob/next/lerna.json#L7 but not a blocker.

jhuleatt:
> LGTM, could also be removed from here: https://github.com/firebase/extensions/blob/next/lerna.json#L7 but not a blocker.

Thanks @Salakar! Removed it

i14h (approved):

PR #551, opened 2021-01-01
[firestore-bigquery-export/@fs-bq-import-collection] enhance error outputs when specified `--non-interactive` with invalid arguments

rika-t:


dackers86:
Thanks @rika-t . Added for our review team

dackers86 (approved):

Issue #552, opened 2021-01-02
storage-resize-images: allow to specify a specific size for a specific path


kuamanet:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

### [REQUIRED] Step 2: Extension name

This feature request is for extension: `storage-resize-images`
### What feature would you like to see?

I'd like to specify one (or more) size on a per-path basis

### How would you use it?

I would generate different image resolutions for different content types (ex. artwork content type vs category content type)

russellwheatley:
Hey @kuamanet, thanks for the suggestion. We're wary of increasing the number of parameters for each extension, particularly one that has 7 already. If anyone else who would like to see this feature, please leave a comment below so we can see how popular it is!

In the mean time, I would suggest that what you want is already possible. You could just create a `storage-resize-image` extension for each image size(s), and use the include and exclude path [options](https://github.com/firebase/extensions/blob/next/storage-resize-images/README.md#additional-setup) to specify which images ought to be resized.

chrisjacobs91:
@russellwheatley I would also find this useful. I am about to setup separate instances of the extension for different image paths, but from how I understood the documentation, the include/exclude fiedlds only match specific file paths.

My folder structure is: 
/profile/{userID}/photos/{photoID}/photo.jpg
/recipe/{recipeID/photos/{photoID}/photo.jpg

Can I simply put /profile in my include list to include ANY images uploaded to any children folders underneath the /profile path?

russellwheatley:
@chris Yes, that is correct.

chrisjacobs91:
can confirm this works and feels like a perfectly good solution to this problem. personally I wouldn't want this feature to be prioritised over some of the other requests :)

WesselvanDam:
Hello, I have a similar question so that's why I'm commenting here instead of opening a new issue. In my app users can upload a profile picture, but to prevent the storage from getting too full I want to resize the image twice: once to a 600x600 size and once to a thumbnail size of 64x64. My storage file structure is as follows:

/avatars/original/photo_600x600.jpg
/avatars/thumbnail/photo_64x64.jpg

Since the feature requested in this issue isn't available, I tried to install the extension twice. The sequence I had in mind was:

[...]

PR #553, opened 2021-01-02
feat(storage-resize-images): allow users to specify which size should be applyed to which path

kuamanet:
First attempt to #552


dackers86:
@kuamanet We will take a closer a look at this and will be able to provide feedback shortly!

PR #561, opened 2021-01-10
Improve Firestore path validation

IchordeDionysos:
Currently, only top-level collections are allowed for the firestore collection "path", but NO subcollections.
To work around this issue @russellwheatley made a change to allow collection group queries #354.

However, in some cases, collection group queries won't work well, and manually importing specific subcollections might be required.

This PR tries to relax the restriction on the input validation.

---

Before this change, this would be allowed: `collection`
But this wouldn't be allowed: `collection/doc/subcollection` and `collection/{doc}/subcollection`

After this change, this will be allowed: `collection` and `collection/doc/subcollection`
But his wouldn't be allowed: `collection/{doc}/subcollection`

### DIFF
--- a/firestore-bigquery-export/scripts/import/src/index.ts
+++ b/firestore-bigquery-export/scripts/import/src/index.ts
@@ -35,7 +35,7 @@ const read = util.promisify(fs.readFile);
 const unlink = util.promisify(fs.unlink);
 
 const BIGQUERY_VALID_CHARACTERS = /^[a-zA-Z0-9_]+$/;
-const FIRESTORE_VALID_CHARACTERS = /^[^\/]+$/;
+const FIRESTORE_VALID_CHARACTERS = /^[^({|})]+$/;
 
 const FIRESTORE_COLLECTION_NAME_MAX_CHARS = 6144;
 const BIGQUERY_RESOURCE_NAME_MAX_CHARS = 1024;


IchordeDionysos:
@russellwheatley @Ehesp not sure here if I should update the package version or a changelog.

If I should do something else, add tests, etc. just tell me (just did this small change quickly on Github) :)

dackers86:
HI @IchordeDionysos. Thanks for this, added to our project board  to take a closer look at!

IchordeDionysos:
@dackers86 thanks should be a fairly straightforward change 👍

dackers86:
# Testing

image

dackers86 (approved):

rashad:
Since everything has been approved, any chance we could merge this anytime soon ? Could use this imports 😅

lszyba1:
Hello,
Could one of you explain how I can import the whole collection with all subcollections.
I've run the 
npx @firebaseextensions/fs-bq-import-collection
"users"
but I can't figure our how to get:
```
users/{userid}/cars
users/{userid}/cars/{parts}
```

I understand in docs it talks about collectionGroup, but I can not find any examples on how exactly one goes about using it or how it relates to running this function in google clould shell?

a) Is collectionGroup a :
```
import collectionGroup
const user_cars =  collectionGroup("users/{userid}/cars") 
and somehow passing it to what?
```
or
b) ? am I suppose to create these somewhere? Where?
c) iscollectionGroup a parameter that is passed to the function

In best case scenario can someone tell me "excicitly" how to import above to subcollections:
```
users/{userid}/cars
users/{userid}/cars/{parts}
```

projectid = myprojectabc123

Thank you
Lucas

Issue #476, opened 2020-10-22
storage-resize-images: Resize Images Extension incorrectly names files containing "."


augustkimo:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: (`storage-resize-images`)
- Extension version: 0.1.12

### [REQUIRED] Step 3: Describe the problem
The resize image extension is incorrectly naming files which are named from a timestamp. The "." in the file name is likely the issue. 

Normally files from the extension are automatically named the same with the width/height following.
ex: profilePic  ->  profilePic_148x148

When renaming files which are named from a timestamp, the width/height goes into the middle of the name.
ex: 2020-10-22T15:29:00.736769  ->  2020-10-22T15:29:00_148x148.736769
The outputted file name should be: 2020-10-22T15:29:00.736769_148x148

#### Steps to reproduce:
1. Upload a photo using a time stamp (this bug might also occur when a "." is in the file name).
2. Have Firebase Resize Photo extension enabled to not delete the original file, and save the new resized image under a "thumbnails" directory.
3. New thumbnail image name will be named incorrectly. 

##### Expected result
A file named: 2020-10-22T15:29:00.736769_148x148

##### Actual result
A file named: 2020-10-22T15:29:00_148x148.736769

##### Suggested Fix / Problem
Edit: The problem is that when uploading a file to storage which doesn't contain a file extension in the set name (picture.of.cats versus picture.of.cats.jpg), `fileNameWithoutExtension` is still called. This results in the the text after the furthest period being detected as the extension, when it is really part of the file name.  Maybe there should be a check before `fileNameWithoutExtension` to ensure that a there is a valid extension in the file name first. 

~~I assume the extension looks for the first "." in the file name, and names the file without the file type extension. (picture.png would be named picture_148x148).~~

~~The code which names the new photo should look for the very last "." when renaming, not the first one. 
ex: myPicture.is.named.with.dots.png should -> myPicture.is.named.with.dots_148x148
but instead the extension will name the picture: myPicture_148x148.is.named.with.dots~~

russellwheatley:
Hey @augustkimo, I'm struggling to recreate the problem you're encountering. Here's a screenshot using `storage-resize-images` version `0.1.12` which have the correct file names:
![Screenshot 2020-10-23 at 13 33 45](https://user-images.githubusercontent.com/16018629/97004098-69a78100-1534-11eb-956e-a870150bd4a5.png)

[...]

PR #579, opened 2021-01-29
feat(storage-resize-images): added multiple image formats to res…

### LINKED ISSUE #575: [storage-resize-images] Multiple image output types for resizer

This feature request is for extension: **storage-resize-images**

I'd like to output multiple image formats from one original file, e.g. origin is a JPEG and outputs JPEGs and WEBP.

### How would you use it?

I'd be using it to get the benefits of lower sizes of WebP while maintaining browser compatibility offering JPEGs.

dackers86:
The current behaviour allows you to convert to a single different type.

This PR allows you to select multiple file types for image type conversion (i.e. `tiff`, `jpeg`).

The original file type can also be selected alongside other types which is not possible currently.

fixes #575

### DIFF
--- a/storage-resize-images/extension.yaml
+++ b/storage-resize-images/extension.yaml
@@ -211,21 +211,13 @@ params:
     required: false
 
   - param: IMAGE_TYPE
-    label: Convert image to preferred type
+    label: Convert image to preferred types
     description: >
-      The image type you'd like your source image to convert to. The default for this option will 
+      The image types you'd like your source image to convert to. The default for this option will 
       be to keep the original file type.
-    type: select
-    options:
-      - label: jpg
-        value: jpg
-      - label: png
-        value: png
-      - label: webp
-        value: webp
-      - label: tiff
-        value: tiff
-      - label: Do not convert
-        value: false
-    default: false
-    required: false
+    example: "jpeg,webp,gif,tiff,avif,heif"
+    validationRegex: ^\b(jpeg|webp|gif|tiff|avif|heif|original)\b(,\b(jpeg|webp|gif|tiff|avif|heif|original)\b)*$
+    validationErrorMessage:
+      Invalid image type(s), must be a comma-separated list containing at least one of jpeg,webp,gif,tiff,avif,heif,original.
+    default: original
+    required: true
--- a/storage-resize-images/extension.yaml
+++ b/storage-resize-images/extension.yaml
@@ -210,7 +210,7 @@ params:
     example: max-age=86400
     required: false
 
-  - param: IMAGE_TYPES
+  - param: IMAGE_TYPE
     label: Convert image to preferred types
     description: >
       The image types you'd like your source image to convert to. The default for this option will 
@@ -225,7 +225,7 @@ params:
         value: png
       - label: tiff
         value: tiff
-      - label: raw
-        value: raw
-    default: raw
-    required: true
+      - label: original
+        value: original
+    default: original
+    required: false
--- a/storage-resize-images/extension.yaml
+++ b/storage-resize-images/extension.yaml
@@ -210,14 +210,22 @@ params:
     example: max-age=86400
     required: false
 
-  - param: IMAGE_TYPE
+  - param: IMAGE_TYPES
     label: Convert image to preferred types
     description: >
       The image types you'd like your source image to convert to. The default for this option will 
       be to keep the original file type.
-    example: "jpeg,webp,gif,tiff,avif,heif"
-    validationRegex: ^\b(jpeg|webp|gif|tiff|avif|heif|original)\b(,\b(jpeg|webp|gif|tiff|avif|heif|original)\b)*$
-    validationErrorMessage:
-      Invalid image type(s), must be a comma-separated list containing at least one of jpeg,webp,gif,tiff,avif,heif,original.
-    default: original
+    type: multiSelect
+    options:
+      - label: jpeg
+        value: jpeg
+      - label: webp
+        value: webp
+      - label: png
+        value: png
+      - label: tiff
+        value: tiff
+      - label: raw
+        value: raw
+    default: raw
     required: true
--- a/storage-resize-images/extension.yaml
+++ b/storage-resize-images/extension.yaml
@@ -226,6 +226,6 @@ params:
       - label: tiff
         value: tiff
       - label: original
-        value: original
-    default: original
-    required: false
+        value: false
+    default: false
+    required: true
--- a/storage-resize-images/functions/src/config.ts
+++ b/storage-resize-images/functions/src/config.ts
@@ -43,5 +43,5 @@ export default {
   includePathList: paramToArray(process.env.INCLUDE_PATH_LIST),
   excludePathList: paramToArray(process.env.EXCLUDE_PATH_LIST),
   deleteOriginalFile: deleteOriginalFile(process.env.DELETE_ORIGINAL_FILE),
-  imageTypes: process.env.IMAGE_TYPE,
+  imageTypes: paramToArray(process.env.IMAGE_TYPE),
 };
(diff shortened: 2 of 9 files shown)


storage-resize-images/extension.yaml
```diff
@@ -210,22 +210,22 @@ params:
...
       - label: tiff
         value: tiff
-      - label: Do not convert
-        value: false
-    default: false
-    required: false
\ No newline at end of file
+      - label: raw
+        value: raw
+    default: raw
```
russellwheatley, line 230: This ought to be backwards compatible as implicitly suggested by Jeff's [comment](https://github.com/firebase/extensions/issues/575#issuecomment-767149474). 

Here's how I think that would work:

1. Switch this file back to the way it was except change the type to `multiSelect`. 
2. In the code, maintain the check to see if they want image conversion i.e `const hasImageTypeConfigSet = imageType !== "false";`
3. If the above condition is `true` then split the string like you're doing here: `const imageTypes = new Set(config.imageTypes.split(","));` and loop through each type for resize.

storage-resize-images/functions/src/resize-image.ts
```diff
@@ -96,24 +99,15 @@ export const modifyImage = async ({
...
-    fileExtension && hasImageTypeConfigSet ? `.${imageType}` : fileExtension;
-
-  let modifiedFileName;
-
-  if (supportedExtensions.includes(fileExtension)) {
-    modifiedFileName = `${fileNameWithoutExtension}_${size}${modifiedExtensionName}`;
-  } else {
-    // Fixes https://github.com/firebase/extensions/issues/476
-    modifiedFileName = `${fileNameWithoutExtension}${fileExtension}_${size}`;
-  }
```
russellwheatley, line 115: put this code back in, please. It was implemented for this [fix](https://github.com/firebase/extensions/issues/476)

russellwheatley (changes requested):

storage-resize-images/extension.yaml
dackers86: I'm not sure this would be possible as we had to introduce a wrapper for every image type to then process each image size.

We do currently default to the raw image type that is provided, this should at least create the original file type.

I'm not sure which bit needs to be backwards compatible. I'd assume going forward it would use the newer implementation?

storage-resize-images/functions/src/resize-image.ts
dackers86: No problem. Looks like it was removed in error, adding back in...

dackers86: Code re-added.

storage-resize-images/extension.yaml
dackers86: Updated to multiSelect. 

Original type added as a default to replace 'do not convert'.

dackers86:
Latest update now reverts `IMAGE_TYPES` back to `IMAGE_TYPE`

storage-resize-images/functions/src/resize-image.ts
```diff
@@ -96,14 +103,14 @@ export const modifyImage = async ({
   contentType: string;
   size: string;
   objectMetadata: ObjectMetadata;
+  format: string;
 }): Promise<ResizedImageResult> => {
-  const { imageType } = config;
-  const hasImageTypeConfigSet = imageType !== "false";
-  const imageContentType = hasImageTypeConfigSet
-    ? supportedImageContentTypeMap[imageType]
+  const useOriginalFormat = format !== "original";
```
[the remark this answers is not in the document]
russellwheatley: also, the naming would be something like `notUsingOriginalFormat`

storage-resize-images/functions/src/resize-image.ts
```diff
@@ -152,11 +159,10 @@ export const modifyImage = async ({
     logs.imageResized(modifiedFile);
 
     // Generate a converted image type buffer using Sharp.
-    if (hasImageTypeConfigSet) {
-      logs.imageConverting(fileExtension, config.imageType);
-      modifiedImageBuffer = await convertType(modifiedImageBuffer);
-      logs.imageConverted(config.imageType);
-    }
+
+    logs.imageConverting(fileExtension, format);
```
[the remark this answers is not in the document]
russellwheatley: You're always converting the image to a type even if it is "original". You need to have that if statement wrapped around here as it was previously to check the image format is to be converted.

russellwheatley (changes requested):

storage-resize-images/functions/src/resize-image.ts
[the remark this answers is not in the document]
dackers86: Renamed and changed to `false`.

Can we provide a use case for why this would need to be backwards compatible, the configuration will now include the word `false` which seems a little counter intuitive?

Ideally I think we should continue to keep it as `original` or another named equivalent.

storage-resize-images/functions/src/resize-image.ts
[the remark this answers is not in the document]
dackers86: Added check to only convert if a specified type.

storage-resize-images/functions/src/resize-image.ts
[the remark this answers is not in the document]
russellwheatley: The previous versions of this extension used "false" to stop the extension from converting the type. If those users upgrade to this future version via the UI, that config value will no longer stop conversion if it is changed to "original".

dackers86:
Example logging:

image

russellwheatley (approved):
LGTM

storage-resize-images/functions/src/config.ts
```diff
@@ -43,5 +43,5 @@ export default {
   includePathList: paramToArray(process.env.INCLUDE_PATH_LIST),
   excludePathList: paramToArray(process.env.EXCLUDE_PATH_LIST),
   deleteOriginalFile: deleteOriginalFile(process.env.DELETE_ORIGINAL_FILE),
-  imageType: process.env.IMAGE_TYPE,
+  imageTypes: process.env.IMAGE_TYPE,
```
Salakar, line 46: Can we make this an array here via `paramToArray` like includePathList & excludePathList do above - for consistency.

Salakar (changes requested):
Just the one feedback item re the config imageTypes - lets make that an array like the others.

storage-resize-images/functions/src/config.ts
dackers86: Updated, testing below...

image

Salakar (approved):

storage-resize-images/functions/src/resize-image.ts
```diff
@@ -96,14 +103,14 @@ export const modifyImage = async ({
...
   objectMetadata: ObjectMetadata;
+  format: string;
 }): Promise<ResizedImageResult> => {
-  const { imageType } = config;
-  const hasImageTypeConfigSet = imageType !== "false";
-  const imageContentType = hasImageTypeConfigSet
-    ? supportedImageContentTypeMap[imageType]
+  const shouldFormatImage = format !== "false";
+  const imageContentType = shouldFormatImage
+    ? supportedImageContentTypeMap[format]
```
jhuleatt, line 110: Seems like we could end up in a situation where `imageContentType` is `undefined` if `format` doesn't exist in `supportedImageContentTypeMap`. Let's check if `format` is valid either here, and/or in `config.ts` and/or `convertType`.

(looks like this potential bug isn't new to this PR, but is probably worth fixing while we're here)

jhuleatt: Oh, `format` comes from a multiselect, not a free-form entry. Maybe not a big deal then? Might still help just to check in case we accidentally change the spelling on a format or something in the future

jhuleatt (approved):

Issue #584, opened 2021-02-03
fs-bq-import-collection - ENOENT error with a non-root collection path


gregfenton:
### Describe your configuration

- Extension name: **firestore-bigquery-export**
- Extension version: **0.1.12**
- Configuration values (redact info where appropriate):
  - Cloud Functions location: **us-central1**
  - BigQuery Dataset location: **us**
  - Collection path: **config/recurring/recur_act**
  - Dataset ID: **tick8s**
  - Table ID: **bqfs_tbl_recur_act**

### Describe the problem

Running the command `@firebaseextensions/fs-bq-import-collection` throws an error (see command and output below).  The exact parameters (see "configuration" section above) do work for the extension -- changing a doc in Firestore does have the record sync to BigQuery.  But running this command to import existing docs to BigQuery fails.  The exact same command works fine for a `--source-collection-path` whose collection is a top-level (a "root") path.

This fails:
```
npx @firebaseextensions/fs-bq-import-collection --non-interactive --project tick8s-test --source-collection-path 'config/recurring/recur_act' --query-collection-group false --dataset tick8s --table-name-prefix bqfs_tbl_xxx --dataset-location us --batch-size 300
{"severity":"WARNING","message":"Warning, FIREBASE_CONFIG and GCLOUD_PROJECT environment variables are missing. Initializing firebase-admin will fail"}
Importing data from Cloud Firestore Collection: config/recurring/recur_act, to BigQuery Dataset: tick8s, Table: bqfs_tbl_xxx_raw_changelog
{"severity":"INFO","message":"BigQuery dataset already exists: tick8s"}
{"severity":"INFO","message":"Creating BigQuery table: bqfs_tbl_xxx_raw_changelog"}
{"severity":"INFO","message":"Created BigQuery table: bqfs_tbl_xxx_raw_changelog"}
[...]

PR #589, opened 2021-02-08
fix(firestore-bigquery-export): Updated file naming for bq imports

dackers86:
fixes: https://github.com/firebase/extensions/issues/584

When using collections below the root collection of database, the `/`separating the collections  is interpreted as part of the temporary file path which is created. This has been updated to use `-` instead.

### DIFF
--- a/firestore-bigquery-export/scripts/import/package.json
+++ b/firestore-bigquery-export/scripts/import/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "@firebaseextensions/firestore-bigquery-change-tracker": "^1.1.2",
     "commander": "5.0.0",
+    "filenamify": "^4.2.0",
     "firebase-admin": "^7.1.1",
     "firebase-functions": "^2.2.1",
     "generate-schema": "^2.6.0",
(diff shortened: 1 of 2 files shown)


gregfenton:
Just for sanity-sake, are there other characters that are valid in a Firestore collection path that would lead to troubles when creating a filename on various OSes ?  Backslash, colon, braces/parens/brackets, or other filesystem-evil characters?

dackers86:
@gregfenton I certainly think that is worth looking at.

@jhuleatt Do you think this is something that we could consider?

jhuleatt:
I'm not sure about the `sanitize-filename` package because of its license, but definitely worth trying to catch other potential filename issues.

gregfenton:
I don't have an opinion on the implementation of meeting the requirement.  I pointed out that particular library purely as an example.

My main point is: rather than addressing the issue one use-case at a time, is there some existing algorithm/library that can sanitize the filename to be "safe" across platforms.  Otherwise the team (hi @dackers86 !) might get pulled back to this issue time and again.

dackers86:
There are another examples of libraries with an MIT license. For example https://github.com/sindresorhus/filenamify

Otherwise the current PR does solve the current solution, perhaps we could create another issue / dicscussion for offering cross platform / better support for file paths?

dackers86:
# Testing

### Test cases using format (a/b/c)

- [x] b
- [x] b*
- [x] b*(
- [x] b*£


### Example result

image

russellwheatley (approved):
LGTM

PR #600, opened 2021-02-15
removing error throwing of expected behaviour

FredCoen:
Removing Errors of expected behavior in the following two scenarios:

1) Removing a non-existent member throws an error. Non-existence is expected because members unsubscribe regularly from an audience. Should log the info but not throw an error. The final state is the same.

2) Adding a member who already is part of the audience throws an error. You often have multiple ways users can signup to an audience. Firebase auth is only one of them. Should log the into but not throw an error. The final state is the same.

### DIFF
--- a/auth-mailchimp-sync/functions/src/index.ts
+++ b/auth-mailchimp-sync/functions/src/index.ts
@@ -63,7 +63,7 @@ export const addUserToList = functions.handler.auth.user.onCreate(
       );
       logs.complete();
     } catch (err) {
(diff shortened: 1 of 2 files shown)


FredCoen:
@googlebot I signed it!

i14h:
@FredCoen thanks for the PR. We've migrated this extension to a new `mailchimp-firebase-sync` extension under Mailchimp. @Ehesp has mirrored this PR under the new extension so I'll close this one

PR #655, opened 2021-05-04
chore(delete-user-data): refactor test data into delete user data

dackers86:
Moved test-data from the extension root into the delete user data folder


delete-user-data/test-data/lib/index.js
```diff
@@ -0,0 +1,113 @@
+"use strict";
```
Salakar, line 1: This file shouldn't be committed, the previous one didn't commit this, only the `.ts` file

dackers86:
## Testing

image

image

image

image

image

Salakar (approved):

jhuleatt (approved):

PR #701, opened 2021-07-14
chore: remove package-lock.json from .gitignore

joehan:
Removing package-lock.json from .gitignore, for a few reasons:
 - Not having these committed caused a broken release of the bigquery extension a few weeks back 
 - npms official help docs say that it is intended to be commited: https://github.com/npm/cli/blob/release-6.14.7/docs/content/configuring-npm/package-lock-json.md
 - The Cloud Build code for node14 runs npm ci, which will not generate a new package-lock.json

I didn't go through and generate package-locks for each extension - I figure we can add these alongside normal PRs to each extension.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ build
 *.iml
 local.properties
 coverage
-package-lock.json
 yarn.lock
 
 firestore-debug.log
--- a/firestore-counter/.gitignore
+++ b/firestore-counter/.gitignore
@@ -2,7 +2,6 @@ test-project-key.json
 
 .vscode
 node_modules
-package-lock.json
 *.js.map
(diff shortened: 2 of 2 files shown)


joehan:
Any idea what is happening  with these tests? I don't understand how I could have broken these by changing a .gitignore.

dackers86:
Thanks  @joehan 

I've added this to the backlog for tracking

Salakar:
And now off the backlog :) there was some changes a couple of days back to fix typing issues - syncing the branch has now included the fixes here

Salakar (approved):
LGTM - my only comment would be that the package locks are not committed in this PR so they'd show as changes for the next commit after this - could we include them in this one before merging perhaps?

PR #702, opened 2021-07-20
Add the from field to the example document.

joshpetit:
The from field is required for some email providers, so adding it to the example 
makes it obvious where it should be.

### DIFF
--- a/firestore-send-email/README.md
+++ b/firestore-send-email/README.md
@@ -15,6 +15,7 @@ Here's a basic example document write that would trigger this extension:
 ```js
 admin.firestore().collection('mail').add({
   to: 'someone@example.com',
+  from: 'somebodyelse@example.com',
   message: {
     subject: 'Hello from Firebase!',
     html: 'This is an <code>HTML</code> email body.',


i14h:
thanks for the suggestion. We'll do a little bit of testing on our side before merging this PR.

dackers86:
Hi @joshpetit 

I've completed an early review, my tests don't appear to allow a `from` value to override.

I don't believe is an issue from the PR but may be an existing bug, have you had an example working where you can override the `from` field?

joshpetit:
Thanks @dackers86! That's strange, I couldn't use the extension with sendgrid until I added the from field to the document. I kept receiving this error:
```
 "Error: Message failed: 550 The from address does not match a verified Sender Identity.
 Mail cannot be sent until this error is resolved. 
Visit https://sendgrid.com/docs/for-developers/sending-email/sender-identity/ to see the Sender Identity requirements"
```
Then once I added the from field and matched it to me verified sender identity it worked.

This is the document that failed:
```json
{
  "message": {
    "subject": "sub",
    "text": "This is the text"
  },
  "to": "recipient@email.address"
}

```

The document that passed:
```json
{
  "from": "verified.sender@email.address",
  "message": {
    "subject": "sub",
    "text": "This is the text"
  },
  "to": "recipient@email.address"
}
```
I just infered this was just an undocumented field we could add. Can you confirm or deny this? Thanks a lot :).

dackers86 (approved):

dackers86:
Thanks @joshpetit.

It's my understanding that this field is based on which smtp server is receiving the email.

Google for example, does not allow custom from fields. I have updated the description to reference this. LGTM

joshpetit:
@googlebot I consent.

joshpetit:
@dackers86 Make sense, I'm glad we can document that :)

dackers86:
@googlebot I consent

i14h:
@kevinthecheung can you review the copy update please?

firestore-send-email/README.md
```diff
@@ -15,13 +15,18 @@ Here's a basic example document write that would trigger this extension:
...
     subject: 'Hello from Firebase!',
     html: 'This is an <code>HTML</code> email body.',
   },
 })
 ```
 
+**custom senders**
+
+The `from` field is an optional parameter. If a custom sender is included, please ensure the receiving SMTP server accepts this fields. Typically, a provider will use the email provided by the configured account details if a custom sender is not allowed.
+
```
kevinthecheung, line 29:
```suggestion
**Custom senders**

The `from` field is an optional parameter. If you specify a value for this field, ensure the receiving SMTP server accepts custom senders in this field. Typically, a provider will use the email address provided by the configured account if the provider doesn't allow custom senders.

```

kevinthecheung (approved):

Issue #697, opened 2021-07-13
Email {{FirstName}} or {{Full Name}} parameters for customer


DCSnip3r:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

### [REQUIRED] Step 2: Extension name

Firestore-send-email

### What feature would you like to see?

Able to specify customer name (first name, last name) in templates/emails.

### How would you use it?

I'd like to be able to say "Dear {{FirstName}}," or "Dear {{FullName}}" and have this information directly pulled from the customer record if it is available.

ankit-programmer:
[Handlebars](https://firebase.google.com/products/extensions/firestore-send-email#:~:text=You%20can%20also%20optionally%20configure%20this%20extension%20to%20render%20emails%20using%20Handlebar%20templates.%20Each%20template%20is%20a%20document%20stored%20in%20a%20Cloud%20Firestore%20collection.) are already supported for these kinds of use cases.

i14h:
Thanks for the suggestion @DCSnip3r. @dackers86 can you look if this suggestion is technically possible?

mbleigh:
Since we have to read the email out of the `USERS_COLLECTION` anyway, it seems pretty reasonable to be able to access other fields in that document via something like `{{user.field_name}}`. I like the suggestion!

I'm not sure if anyone on the Firebase team will have time in the near future to implement it, but I think we'd definitely review a PR if someone wanted to add it.

ankit-programmer:
I would like to implement it.

ankit-programmer:
Hi @mbleigh 
Can you please take a look at https://github.com/firebase/extensions/pull/704 and let me know if any modification is required.

dackers86:
After further discussion, we are placing this back under consideration to see if this type of request gains more traction.

It is quite a technical change which can be handled by developers and could benefit with more developer interest before continuing.

cabljac:
Closing this as unplanned for now. We can reopen if this gets renewed demand. Thanks!

Issue #209, opened 2020-02-27
Send Mail: New Template Collection Not Found


MichaelRShelton:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: firestore-send-email
- Extension version: 0.1.3

### [REQUIRED] Step 3: Describe the problem
I added a new template to an already working extension deployment, but it said the template did not exist even though the logs say it was loaded, causing it to error out. I had to redeploy the extension using the same configurations for the new data to be loaded. This is similar to what happened in #92 

#### Steps to reproduce:

1. Have a working templates and mail collection
2. Add a new template after successful sends.
3. Reference new template in a new mail or update an old one to use it and change state to "RETRY"
4. When it fails, redeploy extension
5. Win

8:40:21.836 am | ext-firestore-send-email-processQueue | loaded template 'reset'
8:40:37.340 am | ext-firestore-send-email-processQueue | Function execution started
8:40:37.349 am | ext-firestore-send-email-processQueue | Started execution of extension with configuration { location: 'us-central1', mailCollection: 'mail', smtpConnectionUri: '<omitted>', defaultFrom: 'noreply@murray-staging.bioblend.net', defaultReplyTo: undefined, usersCollection: 'users', templatesCollection: 'templates' }
8:40:37.468 am | ext-firestore-send-email-processQueue | Attempting delivery for message: mail/OagLhtDRXlslmCJJAxNa
8:40:37.468 am | ext-firestore-send-email-processQueue | Error when delivering message=mail/OagLhtDRXlslmCJJAxNa: Error: tried to render non-existent template 'reset'

##### Expected result
I expected the newly loaded template to be used without error

##### Actual result
Receive an error until I redeployed.

mbleigh:
Thanks for the report, this seems like a bug. @Ehesp can you investigate?

Ehesp:
[...]

PR #704, opened 2021-07-21
Added email personalization from user document.

ankit-programmer:
admin.firestore().collection('mail').add({
personalize: true,
toUids : ["singleUserId"],
  message: {
    subject: 'Hello {{first_name}}',
    html: 'This is an <code>HTML</code> email body.',
  },
})

This will read the first_name field from the users collection and replace it in message.

### DIFF
--- a/firestore-send-email/functions/src/index.ts
+++ b/firestore-send-email/functions/src/index.ts
@@ -21,11 +21,11 @@ import * as nodemailer from "nodemailer";
 import * as logs from "./logs";
 import config from "./config";
 import Templates from "./templates";
-import { QueuePayload } from "./types";
+import { QueuePayload, TemplateData } from "./types";
 
 logs.init();
 
-let db;
+let db: FirebaseFirestore.Firestore;
 let transport;
 let templates;
 let initialized = false;
@@ -40,9 +40,7 @@ async function initialize() {
   db = admin.firestore();
   transport = await transportLayer();
   if (config.templatesCollection) {
-    templates = new Templates(
-      admin.firestore().collection(config.templatesCollection)
-    );
+    templates = new Templates({ collection: admin.firestore().collection(config.templatesCollection) });
   }
 }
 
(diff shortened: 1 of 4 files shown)


ankit-programmer:
@googlebot I signed it!

dackers86:
Hi @ankit-programmer.

Can you confirm what functionality this PR is looking to add, from my understanding...

1. For adding personalisation such as first name, could you pass the data to a template. 

For example...

```js
  const { auth, firestore } = admin;

  // create template
    firestore()
      .collection(`templates`)
      .doc("example_template")
      .set({
        subject: "Hello {{first_name}}",
        html: "This is an HTML email body.",
      })
      .then(() => console.log("created template! ", "example_template"));
  
    // find user / use custom collection
    const user = await auth()
      .getUserByEmail("exmaple.com")
      .then(userRecord => userRecord)
      .catch(error => {
        console.log("Error fetching user data:", error);
      });
  
    console.log("Got user >>>", user);
  
    // send email
    firestore()
      .collection("mail")
      .add({
        to: "example.com",
        template: {
          name: "example_template",
          data: {
            first_name: user.displayName,
          },
        },
      });
```


Also I think this PR is also solving the problem that new templates are not  updated post install. Their is a currently a thread with potential solutions and discussion https://github.com/firebase/extensions/issues/209

clouture-admin:
Hi @dackers86 
Some users don't want to use templates for email personalization. According to #697, he wants to use {{variable}} in the email message itself.
I am using the email (subject and body) itself as a template. It will replace the variables with the actual data from the users collection. So, Hi {{first_name}} will become Hi Ankit in actual message.
**I haven't tried to solve #209 in this PR.**
Do let me know if something needs to be changed in this PR. I will be more than happy to do that. Thanks!

ankit-programmer:
Hi @dackers86
Some users don't want to use templates for email personalization. According to #697, he wants to use {{variable}} in the email message itself.
I am using the email (subject and body) itself as a template. It will replace the variables with the actual data from the user's collection. So, Hi {{first_name}} will become Hi Ankit in the actual message.
I haven't tried to solve #209 in this PR.
Do let me know if something needs to be changed in this PR. I will be more than happy to do that. Thanks!

dackers86:
Apologies @ankit-programmer. 

You're correct, thanks for the resources. 

I'm wondering if we could make this a little more generic.

What do you think of an update where we just append the user information to the data object, something like `_userData` for example.

Data could then look something like....

```js
  {
    ...,
    _userData: {
      ...collectionData
    }
  }
```
Rendering a template similar too...

```js
  template: {
   subject: "@{{_userData.forename}} is now following you!",
   html: "Just writing to let you know that <code>@{{_userData.forename}}</code> ({{name}}) is now following you.",
```


This would reduce the amount of updates, namely...

1. No more need to create a template on the users behalf, this information can be used freely and can be customisable. I'm assuming the currently create templated would be updated in the current source code?

2. Documentation could have a new section on accessing `_meta` or `_userData` for personalising emails.

3. We can remove the personalisation flag and custom logic.

Let me know what you think or if there's anything I've missed!

Issue #652, opened 2021-04-28
Support a Retention policy


iislucas:
### [REQUIRED] Step 2: Extension name

This feature request is for extension: **firestore-bigquery-export**

### What feature would you like to see?

It would be super handy to have the extension provide some data retention support. e.g. for a given table it would be great to be able to say that once a month I want to delete any records that were deleted more than 60 days ago. 

### How would you use it?

This would allow BigQuery to much more easily respect GDPR & related retention policies. (without custom coding)

dackers86:
Thanks @iislucas. I've placed this under consideration

i14h:
@karayu fyi this seems like a interesting feature

jhuleatt:
We document how to clean up stale records in a [sample query](https://github.com/firebase/extensions/blob/next/firestore-bigquery-export/guides/EXAMPLE_QUERIES.md#:~:text=Remove%20stale%20data%20from%20your%20changelog%20table), so I think our first step should be to modify that query to filter for only documents that have been deleted in Firestore. Once we have a query that can reliably get stale deleted records in Firestore, we can consider setting up a scheduled function that runs that query and deletes any returned records.

iislucas:
If I understand it, that query will remove non-deleted live records also - that will be bad for applications that use it e.g. to store user data that hasn't been updated in a while. I'd rather like to remove old updated versions of data, and also any records that were deleted older than, say, 30 days.

dackers86:
Document PR added here https://github.com/firebase/extensions/pull/708

iislucas:
Thanks! That's helpful. I suggest making a variant of this: delete all entries that are older than 1 month and not the latest. This will much more directly map to the common use-case of retention mapping.

PR #708, opened 2021-07-27
chore(firestore-bigquery-export) updated example query to include deleted records

dackers86:
Updated document query for  https://github.com/firebase/extensions/issues/652

### DIFF
--- a/firestore-bigquery-export/guides/EXAMPLE_QUERIES.md
+++ b/firestore-bigquery-export/guides/EXAMPLE_QUERIES.md
@@ -116,11 +116,10 @@ If you want to clean up data from your `changelog` table, use the following
 
 ```sql
 /* The first WHERE clause is the start of the time range, and the second WHERE
-   clause is the end of the time range. The query below deletes every row in a
-   24-hour time frame on the 4th September, 2020 that has ben previously deleted. */
+   clause is the end of the time range. The query below deletes any rows below that are over one
+   month old and have been deleted. */
 
 DELETE FROM `[PROJECT ID].[DATASET ID].[CHANGELOG TABLE ID]`
-WHERE '2020-09-04 00:00:00' < timestamp
-  AND '2020-09-05 00:00:00' > timestamp
+  WHERE DATE(timestamp) >= DATE_ADD(CURRENT_DATE(), INTERVAL 1 MONTH)
   AND operation = "DELETE"
 ```
--- a/firestore-bigquery-export/guides/EXAMPLE_QUERIES.md
+++ b/firestore-bigquery-export/guides/EXAMPLE_QUERIES.md
@@ -112,25 +112,23 @@ WHERE favorite_numbers_index = 0
 ### Remove stale data from your changelog table
 
 If you want to clean up data from your `changelog` table, use the following
-`DELETE` query to delete the rows that fall within a range of timestamps.
+`DELETE` query to delete all rows that fall within a certain time period, 
+e.g. greater than 1 month old.
 
 ```sql
-/* The first WHERE clause is the start of the time range, and the second WHERE
-   clause is the end of the time range. The query below deletes any rows below that are over one
-   month old. */
-
+/* The query below deletes any rows below that are over one month old. */
 DELETE FROM `TABLENAME_raw_changelog`
 WHERE (document_name, timestamp) IN
 (
-WITH latest AS (
+  WITH latest AS (
     SELECT MAX(timestamp) as timestamp, document_name
     FROM `TABLENAME_raw_changelog`
     GROUP BY document_name
-)
-SELECT (t.document_name, t.timestamp)
-FROM `TABLENAME_raw_changelog` AS t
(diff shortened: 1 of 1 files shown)


dackers86:
@iislucas example updated!

iislucas:
Thanks! 

Actually I think the new query will delete even the latest timestamped change, so that probably isn't good. Also, I think you should be subtracting time, not adding, right? Perhaps you want something more like this: 

```
DELETE FROM `TABLENAME_raw_changelog`
WHERE (document_name, timestamp) IN
(
WITH latest AS (
    SELECT MAX(timestamp) as timestamp, document_name
    FROM `TABLENAME_raw_changelog`
    GROUP BY document_name
)
SELECT (t.document_name, t.timestamp)
FROM `TABLENAME_raw_changelog` AS t
JOIN latest  ON (t.document_name = latest.document_name )
WHERE t.timestamp != latest.timestamp 
AND t.timestamp < TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 MONTH)
)
```

dackers86:
> Thanks!
> 
> Actually I think the new query will delete even the latest timestamped change, so that probably isn't good. Also, I think you should be subtracting time, not adding, right? Perhaps you want something more like this:
> 
> ```
> DELETE FROM `TABLENAME_raw_changelog`
> WHERE (document_name, timestamp) IN
> (
> WITH latest AS (
>     SELECT MAX(timestamp) as timestamp, document_name
>     FROM `TABLENAME_raw_changelog`
>     GROUP BY document_name
> )
> SELECT (t.document_name, t.timestamp)
> FROM `TABLENAME_raw_changelog` AS t
> JOIN latest  ON (t.document_name = latest.document_name )
> WHERE t.timestamp != latest.timestamp 
> AND t.timestamp < TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 MONTH)
> )
> ```

Apologies, I understand now in terms retaining the latest and removing older records. 

That query looks great, however the last `WHERE` clause would be an issue....

```
AND t.timestamp < TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 MONTH)
```

` TIMESTAMP_SUB` does not support monthly intervals. As much as I would rather use timestamps, do you think the following would be a valid substitution?

```
AND DATETIME(t.timestamp) < DATE_ADD(CURRENT_DATETIME(), INTERVAL 1 MONTH)
```

iislucas:
Small tweak: I would just make it `INTERVAL -1 MONTH` so that it's a date in the past instead of the future?

but otherwise using dates is fine, and DATE_ADD is good; sorry I didn't realise that TIMESTAMP_SUB didn't support month time intervals.

firestore-bigquery-export/guides/EXAMPLE_QUERIES.md
```diff
@@ -116,7 +116,21 @@ If you want to clean up data from your `changelog` table, use the following
 
 ```sql
 /* The first WHERE clause is the start of the time range, and the second WHERE
-   clause is the end of the time range. The query below deletes every row in a
-   24-hour time frame on the 4th September, 2020. */
-DELETE FROM `[PROJECT ID].[DATASET ID].[CHANGELOG TABLE ID]` WHERE '2020-09-04 00:00:00' < timestamp AND '2020-09-05 00:00:00' > timestamp
+   clause is the end of the time range. The query below deletes any rows below that are over one
+   month old and have been deleted. */
```
Salakar, line 120:
```suggestion
   month old. */
```

firestore-bigquery-export/guides/EXAMPLE_QUERIES.md
```diff
@@ -112,11 +112,23 @@ WHERE favorite_numbers_index = 0
...
+DELETE FROM `TABLENAME_raw_changelog`
+WHERE (document_name, timestamp) IN
+(
+  WITH latest AS (
+    SELECT MAX(timestamp) as timestamp, document_name
+    FROM `TABLENAME_raw_changelog`
+    GROUP BY document_name
+  )
+  SELECT (t.document_name, t.timestamp)
+  FROM `TABLENAME_raw_changelog` AS t
```
Salakar, line 129:
```suggestion
DELETE FROM `[PROJECT ID].[DATASET ID].[CHANGELOG TABLE ID]`
WHERE (document_name, timestamp) IN
(
  WITH latest AS (
    SELECT MAX(timestamp) as timestamp, document_name
    FROM `[PROJECT ID].[DATASET ID].[CHANGELOG TABLE ID]`
    GROUP BY document_name
  )
  SELECT (t.document_name, t.timestamp)
  FROM `[PROJECT ID].[DATASET ID].[CHANGELOG TABLE ID]` AS t
```

Salakar (approved):

huangjeff5:
Can you take another look at this one? It fails the validate / formatting check.

dackers86:
@huangjeff5 All checks now passing

PR #709, opened 2021-07-27
fix(firestore-send-email) allow templates to be automatically updated

dackers86:
fixes: https://github.com/firebase/extensions/issues/209


mbleigh:
This won't have the desired effect, I'm afraid. Since it is running in a separate function it will only reset initialized for itself, not for any running email processing functions. I left a comment on the underlying bug with a potential approach to solving (though honestly this whole bug is a "this should never happen" scenario which I find a bit frustrating).

dackers86:
Closing in favour of https://github.com/firebase/extensions/pull/713

PR #713, opened 2021-07-30
fix(firestore-send-email): added fallback for finding email templates

dackers86:
fixes: https://github.com/firebase/extensions/issues/209

### DIFF
--- a/firestore-send-email/functions/src/templates.ts
+++ b/firestore-send-email/functions/src/templates.ts
@@ -22,6 +22,8 @@ import {
   registeredPartial,
   templateLoaded,
   noPartialAttachmentSupport,
+  checkingMissingTemplate,
+  foundMissingTemplate,
 } from "./logs";
 
 const subjHandlebars = create();
@@ -110,6 +112,12 @@ export default class Templates {
     this.waits.forEach((wait) => wait());
   }
 
+  checkTemplateExists = async (name) => {
+    const query = this.collection.where("name", "==", name).get();
+
+    return query.then((t) => !t.empty);
+  };
+
   async render(
     name: string,
     data: any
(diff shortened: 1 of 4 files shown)


firestore-send-email/functions/src/templates.ts
```diff
@@ -110,6 +112,12 @@ export default class Templates {
     this.waits.forEach((wait) => wait());
   }
 
+  checkTemplateExists = async (name) => {
+    const query = this.collection.where("name", "==", name).get();
```
Ehesp, line 116:
```js
return this.collection.where("name", "==", name).get().then(t => !t.empty);
```

firestore-send-email/functions/src/templates.ts
```diff
@@ -110,6 +112,12 @@ export default class Templates {
     this.waits.forEach((wait) => wait());
   }
 
+  checkTemplateExists = async (name) => {
```
Ehesp, line 115: Doesn't need to be `async`.

firestore-send-email/functions/src/templates.ts
```diff
@@ -122,9 +130,16 @@ export default class Templates {
...
-      return Promise.reject(
-        new Error(`tried to render non-existent template '${name}'`)
-      );
+      //fallback, check if template does exist, results may be cached
+      checkingMissingTemplate(name);
+      const templateExists = await this.checkTemplateExists(name);
+
+      if (!templateExists)
+        return Promise.reject(
+          new Error(`tried to render non-existent template '${name}'`)
```
Ehesp, line 139: `tried` -> `Tried`.

Ehesp (changes requested):

Ehesp (approved):

huangjeff5:
Can we remove the .js files for this extension?

Issue #724, opened 2021-08-19
error instead of "delivery.error" ?


LefebvreIlyas:
Hello everyone,

Is it a mistake or is it intentional?

https://github.com/firebase/extensions/blob/2ad23a74ab40d09bc45395fa31dfe596a5c94d4a/firestore-send-email/functions/src/index.ts#L313

dackers86:
@LefebvreIlyas This looks like a potential error, but also maybe a breaking change if users have assumed the original error should be at the root.

Added PR for review.

LefebvreIlyas:
@dackers86 Thanks for the quick response and PR!

i14h:
@LefebvreIlyas thanks for reporting this. Since changing it will introduce a backward incompatible change for now we'll add `delivery.error` as a new field and in the next major upgrade delete the `error` field.

PR #726, opened 2021-08-20
fix(firestore-send-email): moved lease expired error message to delivery document

dackers86:
fixes: https://github.com/firebase/extensions/issues/724

### DIFF
--- a/firestore-send-email/functions/src/index.ts
+++ b/firestore-send-email/functions/src/index.ts
@@ -310,7 +310,7 @@ async function processWrite(change) {
         return admin.firestore().runTransaction((transaction) => {
           transaction.update(change.after.ref, {
             "delivery.state": "ERROR",
-            error: "Message processing lease expired.",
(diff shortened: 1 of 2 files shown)


firestore-send-email/functions/src/index.ts
```diff
@@ -311,6 +311,7 @@ async function processWrite(change) {
           transaction.update(change.after.ref, {
             "delivery.state": "ERROR",
             error: "Message processing lease expired.",
```
russellwheatley, line 313: maybe just leave a note in the code that it is to be removed in the next major release.

russellwheatley (approved):
LGTM

firestore-send-email/functions/src/index.ts
dackers86: @russellwheatley Thanks Russell, agreed. Added error message, open to suggestions for any  / less detail on the comment.

PR #756, opened 2021-10-01
feat(firestore-bigquery-export): add support for transforming data via a callable function

apascal07:


apascal07:
@dackers86 Will need approval to run the workflows. Thanks.

dackers86:
Thanks @apascal07. I've added to the project board.

dackers86:
Thanks @apascal07.

Would it be possible to include a screenshot of this working as expected? 

Alternatively would be possible to mock an example test case in the `chnage-tracker` tests, this would keep the code coverage levels up and provide assurance this hasn't affected any other functionality.

apascal07:
Tested with my own project + transform function that logs and returns the payload: https://ibb.co/q1m29Yg

dackers86 (approved):

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
```diff
@@ -82,7 +84,25 @@ export class FirestoreBigQueryEventHistoryTracker
         },
       };
     });
-    await this.insertData(rows);
+    const transformedRows = await this.transformRows(rows);
+    await this.insertData(transformedRows);
```
huangjeff5, line 88: Should we explicitly handle the case where transformedRows is undefined (i.e. transform function exists but throws an error)? What will happen in that scenario now?

apascal07: Silently not writing to the table is probably not a good idea. Silently writing the original row instead of the supposed transformed row is an option (depending on the intention, do we want any row or just no row and have some way of backfilling later?). Final option is to throw an error so instead of logging it in transformRows(), could just let it propogate downstream and have the whole execution fail. I don't know what the standard behavior is for extensions.

huangjeff5: Hmm, seems like the last option is most in line with what the current behavior is. It makes sense to me to fail there rather than calling insertData() on potentially bad input since insertData()'s behavior could change in the future.

PR #770, opened 2021-10-14
Fix broken test by pinning @types/node to 14 #769

elvisun:
We are on node 14 but some transitive dependencies are going over to 16, so we pin the type to 14 to avoid test breaking.


dackers86:
# Testing

image

image

image

joehan (approved):

Issue #795, opened 2021-11-14
[Delete-User-Data] It's not possible to delete data from a realtime database located in europe-west1


Andree98:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: delete-user-data
- Extension version: 0.1.11 

#### Steps to reproduce:

Configure the extension to delete data from realtime database located in europe-west1

##### Expected result

Should delete data without any problem

##### Actual result

This error

![Screenshot 2021-11-14 124453](https://user-images.githubusercontent.com/34305925/141681815-4a41780d-4c2b-432a-ad92-8170b56f1476.png)

dackers86:
Hi @Andree98.

Can you confirm the region for your `database` and the installed region for the `extension`.

These values will need to match, the extension can be reconfigured in the firebase console should it need to be updated.

Andree98:
Hello @dackers86 

Both are located on europe-west1

Database

![Screenshot 2021-11-15 102853](https://user-images.githubusercontent.com/34305925/141766248-2d059956-4220-4491-8b9a-cf42a8b052b7.png)

Extension

![Screenshot 2021-11-15 102930](https://user-images.githubusercontent.com/34305925/141766253-3524b46e-825c-4668-aad9-2fd17a69338c.png)

i14h:
this seems like its because we omit the region in the initialization url here https://github.com/firebase/extensions/blob/641092c2580d765cdd3f11bb26328b8c85754563/delete-user-data/functions/src/index.ts#L27

PR #799, opened 2021-11-18
fix(delete-user-data): add RTDB locations

posiek07:
fixes: #795

## Checklist

- add Realtime Database location choice to extensions.yaml
- refactor SELECTED_DATABASE_INSTANCE to camelcase letters (selectedDatabaseInstance)
- add selectedDatabaseLocation
- add getDatabaseUrl function with tests
- fix condition by checking if databaseUrl exist (was provided by user)
- update README.md

## Tests
**Update Engine**
Screenshot 2021-11-19 at 12 29 42

**Installed Extension**
Screenshot 2021-11-19 at 12 28 41

**Extension Settings**

Belgium(west-europe1) RTDB
Screenshot 2021-11-19 at 12 28 54

United States(us-central1) RTDB
Screenshot 2021-11-19 at 13 51 59

**Added data**

Belgium(west-europe1) RTDB
Screenshot 2021-11-19 at 12 25 39
Screenshot 2021-11-19 at 12 25 52
Screenshot 2021-11-19 at 12 28 20
Screenshot 2021-11-19 at 12 26 05

United States(us-central1) RTDB
Screenshot 2021-11-19 at 13 54 42
Screenshot 2021-11-19 at 13 54 31
Screenshot 2021-11-19 at 13 53 54
Screenshot 2021-11-19 at 13 54 16

**Extension succesful logs**

Belgium(west-europe1) RTDB
Screenshot 2021-11-19 at 12 30 23

United States(us-central1) RTDB
Screenshot 2021-11-19 at 13 56 40


dackers86:
@googlebot I consent.

posiek07:
@googlebot I consent.

delete-user-data/functions/src/index.ts
```diff
@@ -24,8 +24,9 @@ import * as logs from "./logs";
 // Initialize the Firebase Admin SDK
 admin.initializeApp({
   credential: admin.credential.applicationDefault(),
-  databaseURL: `https://${config.SELECTED_DATABASE_INSTANCE}.firebaseio.com`,
-});
+  databaseURL: `https://${config.selectedDatabaseInstance}${config.selectedDatabaseLocation === "us-central1" ? ".firebaseio.com" : `.${config.selectedDatabaseLocation}.firebasedatabase.app`}`
```
Ehesp, line 27: Isn't it possible that `config.selectedDatabaseLocation` isn't provided (since it's not required), and this would create the URL `undefined.firebasedatabase.app`?

posiek07: Yes that is possible. Would you suggest then not initializing it at all when you won't provide `SELECTED_DATABASE_INSTANCE `as well? We could make the SELECTED_DATABASE_LOCATION required, but if you won't provide an `selectedDatabaseInstance`(SELECTED_DATABASE_INSTANCE) that will also be undefined.

Thats why I provided extra check before making a call from RTDB to accommodate that somehow.
But making `config.selectedDatabaseLocation. ` would be useful in this case to avoid an error for sure.

Salakar: I thought it'd always be there since the param has a default value: https://github.com/firebase/extensions/pull/799/files#diff-e25bb90bce079f1ec5fc20cf7bebf942c7ec97308184bef1a78ebecf19b1cfa8R179 - I might be wrong though on what the behaviour of default does

posiek07: I think I we should use the same as for the functions LOCATION so maybe
`required: true
default: us-central1
immutable: true
`
What do you think ? Happy to hear some possible implementations.

dackers86: One other issue. https://firebase.google.com/docs/projects/locations

us-central1 | Iowa | DB_NAME.firebaseio.com
-- | -- | --
europe-west1 | Belgium | DB_NAME.europe-west1.firebasedatabase.app
asia-southeast1 | Singapore | DB_NAME.asia-southeast1.firebasedatabase.app

Singapore has it's own unique subdomain. Source: https://firebase.google.com/docs/projects/locations

posiek07: @dackers86 I think that is implemented there in the code already. If it not us-central1 the address will be correct.

posiek07: @Mike @Elliot New helper function now is added with unit tests. Screenshots are updated as well. It won't initialise now if both of the parameters won't be provided and it won't make any call neither. I kept it as optional for better readability from firebase console and be consistent with other options . There is also a sidenote saying  `(Only applicable if you provided the Realtime Database instance parameter.)`

dackers86 (approved):

dackers86 (approved):

posiek07:
Known issues with CLA involving co-contributors. Closing in favour of #865

PR #786, opened 2021-10-22
fix(*): refactor emulator setup for testing

dackers86:
fixes: https://github.com/firebase/extensions/issues/689

PR #816, opened 2021-12-03
fix(firestore-send-email): add smtp secret parameter

posiek07:
## Problems
 - password isn't encoded correctly and could be compromised
 - email or password can be deconstructed incorrectly from URI when containing some special characters
 - possible problems with regex validation
 
 ## Solution
 Break the URI address into sections. Create new optional fields for typing the values for setting up an SMTP server with a hidden `secret` password.
 
 ## Checklist
- [x] add 4 optional fields in extensions.yaml. When `SMTP_SERVER_HOST_AND_PORT` is provided the `SMTP_CONNECTION_URI` won't be used that is also changed to optional. (this should provide a smooth update process)
- [x] add new ENV variables in config and jest-config, emulator-param files 
- [x] add a `setSmtpCredentials` helper function for selecting the right mail config
- [x] add logger error when you do NOT provide none of the configuration variables (`SMTP_SERVER_HOST_AND_PORT` or `SMTP_CONNECTION_URI `)
- [x] add unit tests for `setSmtpCredentials` helper function

## Under Consideration
- making `SMTP_SERVER_HOST_AND_PORT` and `SMTP_SERVER_SSL` required to make user reconfigure extension on update and maybe remove previous `SMTP_CONNECTION_URI` field. We cannot make login and password required as we want still a possibility to send unauthed emails (I have tested it locally with port 25 and it worked with omitted user values).  
- combine `SMTP_SERVER_HOST_AND_PORT` with `SMTP_SERVER_SSL` connection and format for e.g. `smtps:smtp.gmail.com:465`

## Tests
**Update Engine**
Screenshot 2021-12-03 at 11 34 02

**Installation and configuration of extension**
Screenshot 2021-12-03 at 11 33 38
Screenshot 2021-12-03 at 11 33 14
Screenshot 2021-12-03 at 11 33 19

**Successful results**
Screenshot 2021-12-03 at 11 35 43
Screenshot 2021-12-03 at 11 34 36

## Extra tests
**Old setup**
Screenshot 2021-12-17 at 08 38 03
Screenshot 2021-12-17 at 09 29 06
Screenshot 2021-12-17 at 09 33 52
Screenshot 2021-12-17 at 09 30 27

**New setup**
Screenshot 2021-12-17 at 09 44 55
Screenshot 2021-12-17 at 10 10 41
Screenshot 2021-12-17 at 09 57 22
Screenshot 2021-12-17 at 09 55 33

**Both setups (choose new recommended setup if SMTP connection and port filled)**
Screenshot 2021-12-17 at 10 07 37
Screenshot 2021-12-17 at 10 07 45
Screenshot 2021-12-17 at 10 10 41
Screenshot 2021-12-17 at 10 09 31


firestore-send-email/functions/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,53 @@
...
+describe("function setSmtpCredential test returning creating nodemailer transport based on ENV config", () => {
+  test("Mail config without smtpConnectionUri(SMTP_CONNECTION_URI env variable), old setting", () => {
+    const config: Config = {
+      smtpServerDomain: "smtp.gmail.com:25",
+      location: "",
+      mailCollection: "",
+      defaultFrom: "",
+    };
+    const credentials = setSmtpCredentials(config);
+    expect(credentials).toBeInstanceOf(Mail);
```
dackers86, line 14: Can we add additional tests on the configuration. For example: checking that the credentials has the correct domain name / port etc.

firestore-send-email/functions/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    expect(credentials.options.port).toBe(25);
+    expect(credentials.options.host).toBe("smtp.address.com");
+  });
+
+  test("Mail config without smtpServerDomain(SMTP_SERVER_HOST_AND_PORT env variable) and smtpConnectionUri(SMTP_CONNECTION_URI env variable). Should return null", () => {
```
dackers86, line 44: Can potentially exclude the `config` names, this will keep the descriptions shorter.

Additionally, the first sentence appears as a `describe` instruction. Could potentially divide these tests into multiple describe functions for each test case?

firestore-send-email/functions/src/helpers.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const setSmtpCredentials = (config) => {
+  const {
+    smtpConnectionUri,
+    smtpServerDomain, // eg.smtp.gmail.com:465
+    smtpServerSSL,
+    smtpEmail,
+    smtpPassword,
+  } = config;
+  let smtpCredentials;
```
dackers86, line 11: We can perhaps return early from this function. Instead of assigning the credentials to a variable.

`if(smtpServerDomaiin) return createTransport(...`
`if(smtpConnectionUri) return createTransport(...`
`return null;`

firestore-send-email/functions/__tests__/emulator/emulator-params.env
```diff
@@ -2,6 +2,10 @@ LOCATION=europe-west2
 TEMPLATES_COLLECTION=templates
 MAIL_COLLECTION=mail
 SMTP_CONNECTION_URI=smtps://fakeemail@gmail.com:secret-password@smtp.gmail.com:465
+SMTP_SERVER_HOST_AND_PORT=smtp.gmail.com:465
+SMTP_SERVER_SSL=true
```
dackers86, line 6: We may need additional `e2e` to ensure that each scenario for the configuration works.

This may help https://github.com/firebase/extensions/pull/786/files, specifically `firebase ext:dev:emulators:exec `. Is it possible to run  test scenarios where there is no username / password

dackers86 (changes requested):

firestore-send-email/functions/__tests__/helpers.test.ts
posiek07: No worries. I will remove config in test description.

Why separate it into smaller describe functions? I cannot find the reason behind it as this is how your group tests in jest.

firestore-send-email/functions/__tests__/helpers.test.ts
posiek07: Sure : )

firestore-send-email/functions/__tests__/emulator/emulator-params.env
posiek07: Yes, this scenario also worked (without login and pass). I have only tested it manually in local `node` because I could not test it online. Port 25 is blocked to use for Firebase Functions as I understand and I needed it to send email without credentials by Gmail. 

How do you see testing different configurations as you can only use one .env file for tests running? We could overwrite them in each test, but I am not sure if this is what are you after.

firestore-send-email/functions/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    };
+    const credentials = setSmtpCredentials(config);
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.port).toBe(25);
+    expect(credentials.options.host).toBe("smtp.mail.com");
+    expect(credentials.options.auth.user).toBe(config.smtpEmail);
+    expect(credentials.options.secure).toBe(false);
+  });
+
+  test("return null", () => {
```
dackers86, line 63:
```suggestion
  test("it provides null credentials with default configuration", () => {
```

dackers86: Additional description to describe what the test is showing?

dackers86:
@googlebot I consent.

posiek07:
closing in favor of: #840

PR #832, opened 2021-12-13
fix(firestore-bigquery-export): documentation fix and version increment

apascal07:


firestore-bigquery-export/PREINSTALL.md
```diff
@@ -33,6 +34,7 @@ Prior to sending the document change to BigQuery, you have an opportunity to tra
     },
   }]
 }
+```
```
huangjeff5, line 37: I believe we also want to add this formatting to the README too

Issue #692, opened 2021-07-07
URL Encode SMTP user & passowrd


alfredsgenkins:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **firestore-send-email**
- Extension version: **0.1.10**

#### Steps to reproduce:

Consider the following SMTPS URI: `smtps://XXX:XXX/X+XXX@email-smtp.eu-west-1.amazonaws.com:465`. 

If passed into `nodemailer` / `firestore-send-email` configuration will result in following error when attempting to send email: `getaddrinfo ENOTFOUND XXX`.

To resolve an issue the password and username have to be URL encoded, which is not mentioned anywhere in the configuration description.

##### Expected result

There is a not-to URL encode the password and username.

##### Actual result

There is no such note, which results in confusing error message.

dackers86:
Thanks @alfredsgenkins.

I have updated this for consideration.

jhuleatt:
We should be able to update the extension itself to use [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) so that someone installing the extension doesn't need to figure out encoding

alfredsgenkins:
I think the problem is that there are no password/username fields. It is required to parse the URL with regex, then encode the result, then use it as a new URI. The biggest problem here would be the case of a password or username containing the `:`. This could lead to the wrong detection of username and password.

Maybe instead, we should switch to declarative transport declaration schema. The port, host, protocol, credentials should be separate fields. In this case, no URL encoding is needed.

Alternatively, the simple notice could be shown in the configuration.

dackers86:
@alfredsgenkins @jhuleatt  After some initial testing, I feel this may be difficult to implement.

**Decoding the uri**
[...]

Issue #648, opened 2021-04-26
firestore-send-email SMTPS field accept only alphanumeric characters


app-condo:
### Describe your configuration

- Extension name: firestore-send-email
- Extension version: 0.1.9
- Configuration values (redact info where appropriate):
  - smtps://xxxx@gmail.com:password@smtp.gmail.com:465

### Describe the problem

firestore-send-email SMTPS accept only alphanumeric characters [A-Za-z0-9]
escaping # character to html & # 3 5 ; don't work

#### Steps to reproduce:

insert a SMTPS password with special character

dackers86:
Hi @app-condo 

I have tested this with some basic special characters, is there a particular character that is causing this issue?

The regular expression for this URI is `^smtps?://[^:]+:[^:]+@.*$` if that helps?

antoniosap:
for example i use # inside password.

dackers86:
Potentially related to https://github.com/firebase/extensions/issues/692

dackers86:
Closing in favour of https://github.com/firebase/extensions/pull/816

PR #840, opened 2022-01-06
fix(firestore-send-emial): add SMTP secret password

posiek07:
fixes: #648
fixes: #692 

## Problems
 - password isn't hidden and could be compromised
 - password containing some special characters can make an address decode incorrectly
 - possible problems with regex validation
 
 ## Solution
 Break the URI address into sections. Create a new optional password field for setting up an SMTP server with a hidden `secret` password.
 
 ## Checklist
- [x] add optional field in extensions.yaml. When `SMTP_PASSWORD` is provided the `SMTP_CONNECTION_URI` password won't be used. (this should provide a smooth update process)
- [x] add new ENV variables in config and jest-config, emulator-param files 
- [x] add a `setSmtpCredentials` helper function for selecting the right mail config
- [x] add unit tests for `setSmtpCredentials` helper function
- [x] fix URI regex validation

## Tests
**Update Engine**
Screenshot 2022-01-06 at 11 31 45

**Installation and configuration of extension**
Screenshot 2022-01-06 at 11 28 47

**Successful results (both setups)**
Screenshot 2022-01-06 at 11 30 11
Screenshot 2022-01-06 at 11 25 50

## Extra tests
**Old setup**
Screenshot 2022-01-06 at 11 44 02
Screenshot 2022-01-06 at 11 45 22
Screenshot 2022-01-06 at 11 46 28

**New setup**
Screenshot 2022-01-06 at 12 07 07
Screenshot 2022-01-06 at 12 08 04
Screenshot 2022-01-06 at 12 07 53

### DIFF
--- a/firestore-send-email/functions/__tests__/helpers.test.ts
+++ b/firestore-send-email/functions/__tests__/helpers.test.ts
@@ -1,0 +1,71 @@
+import Mail = require("nodemailer/lib/mailer");
+import { setSmtpCredentials } from "../src/helpers";
+import { Config } from "../src/types";
+
+describe("set server credentials helper function", () => {
+  test(" return smtpServerDomain credentials with new password", () => {
+    const config: Config = {
+      smtpConnectionUri:
+        "smtps://fakeemail@gmail.com:secret-password@smtp.gmail.com:465",
+      smtpPassword: "fakepassword",
+      location: "",
+      mailCollection: "",
+      defaultFrom: "",
+    };
+    const credentials = setSmtpCredentials(config);
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.port).toBe(465);
+    expect(credentials.options.host).toBe("smtp.gmail.com");
+    expect(credentials.options.auth.pass).toBe(config.smtpPassword);
+    expect(credentials.options.secure).toBe(true);
+  });
+
+  test(" return smtpServerDomain credentials with new password (old deleted)", () => {
+    const config: Config = {
+      smtpConnectionUri: "smtps://fakeemail@gmail.com@smtp.gmail.com:465",
+      smtpPassword: "sec#:@ret-password",
+      location: "",
+      mailCollection: "",
+      defaultFrom: "",
+    };
+    const credentials = setSmtpCredentials(config);
+    expect(credentials).toBeInstanceOf(Mail);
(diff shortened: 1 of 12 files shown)


firestore-send-email/functions/src/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+  let url = new URL(smtpConnectionUri);
+  let smtpCredentials;
+  if (!url) {
+    return (smtpCredentials = null);
+  }
+  if (smtpConnectionUri) {
+    smtpCredentials = createTransport({
+      host: decodeURIComponent(url.hostname),
+      port: parseInt(url.port),
+      secure: url.protocol.includes("smtps") ? true : false, // true for 465, false for other ports
```
dackers86, line 16: I'm not too sure if this defines what secure connection is https://nodemailer.com/smtp/.

May be worth looking to see if this field is needed. Or could even be another config option (in the future?).

firestore-send-email/functions/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      location: "",
+      mailCollection: "",
+      defaultFrom: "",
+    };
+    const credentials = setSmtpCredentials(config);
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.port).toBe(465);
+    expect(credentials.options.host).toBe("smtp.gmail.com");
+    expect(credentials.options.auth.user).toBe("fakeemail@gmail.com");
+    expect(credentials.options.auth.pass).toBe("");
```
dackers86, line 68: Should this return a null value.

Will node-mailer interpret this as a valid password instead of non provided?

firestore-send-email/functions/src/helpers.ts
```diff
@@ -0,0 +1,21 @@
+import { createTransport } from "nodemailer";
+import { URL } from "url";
+import { Config } from "./types";
+
+export const setSmtpCredentials = (config: Config) => {
+  const { smtpConnectionUri, smtpPassword } = config;
+  let url = new URL(smtpConnectionUri);
+  let transport;
+
+  if (!url) {
```
dackers86, line 10: Can we log to the console here to say `invalid url, please reconfigure with a valid connection` and return early.

dackers86 (approved):

Issue #627, opened 2021-03-30
fs-bq-schema-views script: Cannot read property 'map' of undefined


slickorange:
### [READ] Step 1: Are you in the right place?
I have an issue with the `fs-bq-schema-views` script for the firestore-bigquery-export extension.
### [REQUIRED] Step 2: Describe your configuration

- Extension name: firestore-bigquery-export
- Extension version: 0.1.13

### [REQUIRED] Step 3: Describe the problem
I get the following error when trying to run the fs-bq-schema-views script.

```
{"severity":"WARNING","message":"Warning, FIREBASE_CONFIG and GCLOUD_PROJECT environment variables are missing. Initializing firebase-admin will fail"}
{}
Cannot read property 'map' of undefined
```

#### Steps to reproduce:

I am running the command as follows:

```
npx @firebaseextensions/fs-bq-schema-views \
  --non-interactive \
  --project=my-project\
  --dataset=firestore_export_staging \
  --table-name-prefix=orders \
  --schema-files=orders.json 
```

I have made my json file very simple to make sure it is not an issue with the json:

```
{
  "fields": [
    {
      "name": "id",
      "type": "string"
    },
    {
      "name": "systemId",
      "type": "number"
    }
]
}
```
##### Expected result
The relevant view should be generated in BigQuery.

##### Actual result
I get an error in the command line `Cannot read property 'map' of undefined`; I can see the "firestore type" functions (eg firestoreArray) but not the generated view.

jhuleatt:
Hey @slickorange, I'm wondering if the first log line is the issue here:

```
"severity":"WARNING","message":"Warning, FIREBASE_CONFIG and GCLOUD_PROJECT environment variables are missing. Initializing firebase-admin will fail"
```

Please follow the [set up credentials](https://github.com/firebase/extensions/blob/next/firestore-bigquery-export/guides/GENERATE_SCHEMA_VIEWS.md#step-2-set-up-credentials) instructions before running the script, and see if that fixes it for you.

slickorange:
@jhuleatt Thanks.. I did follow the instructions to set up the credentials and I did it again now just to make sure. Still having the same issue. Any other ideas? Any ideas on how I can start debugging this?

slickorange:
I have copied the script and I am running it locally.. from what I can tell it seems like `field.fields` schema.js:267 is undefined.. resulting in the failure of the map call on line 125. Still trying to figure out why it is undefined.

slickorange:
[...]

PR #845, opened 2022-01-12
fix: added fields null check when generating schemas

dackers86:
fixes: https://github.com/firebase/extensions/issues/627


firestore-bigquery-export/scripts/gen-schema-view/src/schema.ts
```diff
@@ -382,7 +382,8 @@ function processFirestoreSchemaHelper(
   bigQueryFields: { [property: string]: string }[],
   extractPrefix: string[]
 ) {
-  const { fields } = schema;
+  const { fields = [] } = schema;
+  if (!fields.length) return null;
```
dackers86, line 386: Remove this line. It is not needed.

Salakar (approved):
LGTM pending null return removal as per comment

PR #849, opened 2022-01-13
feat(firestore-bigquery-export): add support for transforming data via a callable function

apascal07:


apascal07:
Reattempt of #756 + #832 after looking into why NPM publish didn't work last time.

huangjeff5 (approved):

dackers86:
@apascal07 My only concern here is that the `change-tracker` will need to be released and `published` as part of the release process, otherwise the updated package json reference:

`"@firebaseextensions/firestore-bigquery-change-tracker": "^1.1.14"`

will be unavailable when upgrading the extension.

apascal07:
Publishing dependencies to the NPM registry is part of our release process so this shouldn't be a concern.

PR #874, opened 2022-02-04
chore(*) added TODOs for billing updates in all extensions pre-install files

dackers86:
@kevinthecheung The text `a small amount` has been highlighted for removal in all extensions.

### DIFF
--- a/auth-mailchimp-sync/PREINSTALL.md
+++ b/auth-mailchimp-sync/PREINSTALL.md
@@ -13,7 +13,7 @@ Make sure that you've set up [Firebase Authentication](https://firebase.google.c
 You must also have a Mailchimp account before installing this extension.
 
 #### Billing
- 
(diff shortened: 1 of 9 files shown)


karayu:
Hey! We should remove this line altogether when @pavelgj turns AR live -- that's planned for this week

dackers86:
Hi @karayu 

I can make this update, just to confirm are we removing the entire line from the preinstalls? For example:

```
You will be charged a small amount (typically around $0.01/month) for the Firebase resources required by this extension (even if it is not used).
```

karayu:
Yep! Though actually we should confirm with @pavelgj that it's rolled out

dackers86:
The AR feature is to be rolled back. Moving to blocked for now.

pr-Mais (approved):

Issue #620, opened 2021-03-14
[storage-resize-images] Output options


japettyjohn:
### [REQUIRED] Step 2: Extension name

This feature request is for extension: **storage-resize-images**

### What feature would you like to see?

I'd like to set a way of setting output options for the sharp image processor.

### How would you use it?

Change the settings like options.quality for webp and the various configuration options for jpeg.

dackers86:
Hi @japettyjohn.

Thanks for the request, this is something that was recently discussed as part of https://github.com/firebase/extensions/issues/605.

I'll put this up for discussion.

PR #878, opened 2022-02-08
feat(storage-resize-images): add output options

posiek07:
fixes: #620 

## Problem
Users would like to set up custom Output Options like `quality` and others for the converted files.

## Proposed Solution
Create optional param for feeding JSON object with settings for converted files.

## Jobs
- [x] add OPTIONS_OUTPUT param with validation
- [x] update docs with example
- [x] parse JSON and provide correct format options into buffer creation

### Installed configuration
Screenshot 2022-02-08 at 13 51 57
Screenshot 2022-02-08 at 13 51 45
Screenshot 2022-02-08 at 14 02 10

### Succesful results
Screenshot 2022-02-08 at 14 00 07
Screenshot 2022-02-08 at 14 03 37


storage-resize-images/extension.yaml
```diff
@@ -243,3 +243,14 @@ params:
...
+
+
+  - param: OUTPUT_OPTIONS
+    label: Output options for selected formats
+    description: >
+      Provide a optional output option stringified object containing Sharp Output Options for selected image types conversion.
+      eg. `{"jpeg": { "quality": 5, "chromaSubsampling": "4:4:4" }, "png": { "pallete": true }}`
+    type: string
+    validationRegex: ^({(.*?)})$
+    validationErrorMessage: Please provide a valid json object. Expected to start with `{`.
```
dackers86, line 255: `expected to start with` presumes this is only invalid way for writing JSON according to this validation.

Are there other cases for invalid regex and if so can we remove that sentence from the `validationErrorMessage`?

storage-resize-images/functions/lib/resize-image.js
```diff
@@ -29,24 +29,32 @@ function resize(file, size) {
 }
 exports.resize = resize;
 function convertType(buffer, format) {
+    let outputOptions = { jpeg: {}, png: {}, webp: {}, tiff: {} };
```
dackers86, line 32: Do we need `tif` and `jpg` as alternative spellings? Or documentation to explain that if a user includes `tif` for example that they would need to use `tiff` for output options?

dackers86 (approved):

joehan (approved):

PR #894, opened 2022-03-02
fix(firestore-bigquery-export): updated failing BQ tests and removed lib files

dackers86:
A current test failed for an invalid check on partitioning view table. As partitioning can be added to a created table with a view this has been updated.

Following the addition of a `prepare` script, the lib folder is no longer required to be stored in the repository. This is now autogenerated on `install`,  two remaining files still existed in this folder and have now been removed.


russellwheatley (approved):
My only nit would be to have separate PRs for each, but LGTM.

PR #900, opened 2022-03-03
fix(firestore-bigquery-export): updated the bq projectId to be required and to default as the current firebase project

dackers86:
Originally a BigQuery Project Id was allowed to be left as an optional field and would default to the current firebase project that the extension is running on.

With this field blank, documentation would show this as an `unspecified parameter`. This update ensures that a `project name` is always provided for the extension to synchronise data with and to always provide a `BigQuery Project Name` for the extensions documentation.

### Default Id

image

image

### Alternaitve id

image

### No Id

image

### DIFF
--- a/firestore-bigquery-export/extension.yaml
+++ b/firestore-bigquery-export/extension.yaml
@@ -192,10 +192,11 @@ params:
   - param: BIGQUERY_PROJECT_ID
     label: Project Id
     description: >-
-      Override the default project bigquery instance. 
+      Override the default project bigquery instance.
       This can allow updates to be directed to a bigquery instance on another project.
     type: string
-    required: false
+    default: ${param:PROJECT_ID}
+    required: true
 
   - param: COLLECTION_PATH
     label: Collection path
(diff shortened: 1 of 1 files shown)


firestore-bigquery-export/extension.yaml
```diff
@@ -192,10 +192,11 @@ params:
   - param: BIGQUERY_PROJECT_ID
     label: Project Id
     description: >-
-      Override the default project bigquery instance. 
+      Override the default project bigquery instance.
       This can allow updates to be directed to a bigquery instance on another project.
     type: string
-    required: false
+    default: ${param:PROJECT_ID}
```
pavelgj, line 198: are you sure this will works?

pavelgj (approved):

firestore-bigquery-export/extension.yaml
dackers86: Confirming change updated to `default: ${PROJECT_ID}`

Issue #907, opened 2022-03-09
Cannnot upgrade from 0.1.13 to 0.1.15 of firestore-send-email


Bandit:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: `firestore-send-email`
- Extension version: `0.1.15`
- Configuration values (redact info where appropriate):
  - smtp://`[username]`:`[password]`@smtp.postmarkapp.com:587

#### Steps to reproduce:

Try to upgrade.

##### Expected result

Can upgrade as existing configuration works fine (in `0.1.13`).

##### Actual result

> Invalid SMTP connection URI. Must be in the form `smtp(s)://username:password@hostname:port` or `smtp(s)://username@hostname:port`

dackers86:
Thanks for the feedback @Bandit.

Added for investigation.

dackers86:
@Bandit Please see https://github.com/firebase/extensions/issues/912#issuecomment-1066862681 as a potential solution until the installation patch update can be applied.

PR #909, opened 2022-03-10
fix(firestore-send-email): enhanced regular expression validation

dackers86:
fixes: https://github.com/firebase/extensions/issues/907


apascal07 (approved):

ottob:
This regexp does not support the connection uri that I need to use, and that works with firestore-send-email@0.1.11: `smtps://myuser@mydomain.com@smtp-relay.gmail.com:465?name=mydomain.com`

PR #918, opened 2022-03-15
Remove `password` from connection-uri's err msg

tjlav5:
We don't support passwords here, they need to be in the smtp_password field


tjlav5:
Properly fixed w/ https://github.com/firebase/extensions/pull/909/files

PR #919, opened 2022-03-16
feat(firestore-bigquery-export): added JSON data type to the default bigquery schema

### LINKED ISSUE #1775: 🐛 [firestore-bigquery-export] change datatype in changelog table from STRING to JSON
Currently the `data`-field in the `xxx_raw_changelog` is of type STRING, but afaik this field always contains JSON data. Since BQ natively supports JSON datatypes it would be great to ship this plugin with a standard datatype JSON.

The `old_data` field is also of type STRING, so I guess this should also be shipped by default as JSON. 

Did some tests with this new datatype and the actual changelog table and it works seamlessly 

### Why? 
Because it saves us a lot of money (don't have to query the full blob of data) and we want to make use of all native `JSON_XX` functions.

dackers86:
The BigQuery extension currently uses a `stringified` JSON object to store data when synching from Firestore. This PR prepares for when a JSON type becomes available.

A new JSON datatype is currently in preview and can be tracked here https://cloud.google.com/bigquery/docs/reference/standard-sql/json-data

- [x] Update the default schema to use a JSON datatype
- [ ] Ensure new tables are generated with the new datatype

fixes: https://github.com/firebase/extensions/issues/1775

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
@@ -19,10 +19,10 @@ import * as firebase from "firebase-admin";
 import * as traverse from "traverse";
 import fetch from "node-fetch";
 import {
-  RawChangelogSchema,
-  RawChangelogViewSchema,
   documentIdField,
   documentPathParams,
+  SelectRawChangelogSchema,
+  SelectRawChangelogViewSchema,
 } from "./schema";
 import { latestConsistentSnapshotView } from "./snapshot";
 import handleFailedTransactions from "./handleFailedTransactions";
@@ -55,6 +55,7 @@ export interface FirestoreBigQueryEventHistoryTrackerConfig {
   clustering: string[] | null;
   wildcardIds?: boolean;
   bqProjectId: string | undefined;
+  data_format?: string | undefined;
   backupTableId?: string | undefined;
 }
 
@@ -332,7 +333,9 @@ export class FirestoreBigQueryEventHistoryTracker
       }
     } else {
       logs.bigQueryTableCreating(changelogName);
-      const schema = { fields: [...RawChangelogSchema.fields] };
+
+      /* select default schema type based on configuration */
+      const schema = SelectRawChangelogSchema(this.config.data_format);
 
       if (this.config.wildcardIds) {
         schema.fields.push(documentPathParams);
@@ -364,7 +367,9 @@ export class FirestoreBigQueryEventHistoryTracker
     const dataset = this.bigqueryDataset();
     const view = dataset.table(this.rawLatestView());
     const [viewExists] = await view.exists();
-    const schema = RawChangelogViewSchema;
+
+    /* select default schema type based on configuration */
+    const schema = SelectRawChangelogViewSchema(this.config.data_format);
 
     const partitioning = new Partitioning(this.config, view);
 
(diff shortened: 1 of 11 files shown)


IchordeDionysos:
@dackers86 when is it planned to merge this PR? Sounds like a helpful addition 😍

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/schema.ts
```diff
@@ -76,44 +77,104 @@ export const documentPathParams = {
...
     },
+    documentIdField,
+  ],
+};
+
+export const RawChangelogJSONViewSchema = {
+  fields: [
+    ...defaultViewSchemaFields,
     {
       name: "data",
```
IchordeDionysos, line 175: The field `path_params` should also be made a JSON object depending on the users selection.

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
```diff
@@ -423,6 +430,7 @@ export class FirestoreBigQueryEventHistoryTracker
       if (this.config.wildcardIds) {
         schema.fields.push(documentPathParams);
       }
+
       const latestSnapshot = latestConsistentSnapshotView(
```
IchordeDionysos, line 434: The `latest` BigQuery view also needs to be adjusted as it's not possible to group a JSON column.

This is what comes up when you have `data` as a `JSON` field, with the auto-created`collection_raw_latest` view.
image

IchordeDionysos (changes requested):
Nice work, but there are some missing pieces, especially the latest view won't work anymore when using the `JSON` data type

dackers86:
Thanks [IchordeDionysos](https://github.com/IchordeDionysos). Now that the JSON updated is out of it's original `preview` status we can resume development on this. The next steps will be to address the issues you have pointed out in terms of the `views` and `path_params`.

nfarina:
I was able to make this work by using `SELECT …, ARRAY_AGG(data)[offset(0)] as data,` instead of attempting to `GROUP BY data` since grouping by native JSON type is still unsupported.

However, this removes most of the advantages of the JSON type! When querying the resulting View, the entire `data` column is loaded (and billed) instead of a subset. For instance `SELECT data.age FROM latest` will read the same number of bytes as `SELECT data FROM latest`.

This is as opposed to `SELECT data.age FROM changelog" which only bills for the `age` column (since native JSON types are spread out into native subcolumns at ingestion time).

I don't know enough about BigQuery to suggest an alternative strategy, but I really hope there is a way to have our cake and eat it too here.

dackers86:
Moving this back to `under consideration` for planning.  This PR was an initial start but we need to consider and plan how this fits into the overall BQ schema and development.

basvandorst:
We manually changed the datatype of the changelog-table to JSON and had indeed to deal with the same group by issue. 

We fixed it in this way: 
```
WITH RankedData AS (
  SELECT
    document_name,
    document_id,
    timestamp,
    event_id,
    operation,
    data,
    path_params,
    ROW_NUMBER() OVER(PARTITION BY document_name ORDER BY timestamp DESC) AS row_num
  FROM `collection`
)
SELECT
  document_name,
  document_id,
  timestamp,
  event_id,
  operation,
  data,
  path_params
FROM RankedData
WHERE row_num = 1 
```

Don't know enough about the inner workings /performance impact of BQ but for us it works fine!

tiagosilveiransj:
Any updates on this? It would be very useful.

cabljac:
closing as stale

PR #920, opened 2022-03-16
chore(*): updated documentation to refer to no-cost tier

dackers86:
Copy changes across all extensions.

This update changes all extensions to now use `no-cost tier` instead of `free tier`.


dackers86:
Note: file changes have disappeared. Moving back to review.

PR #928, opened 2022-03-26
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/firestore-counter/clients/node/index.js
+++ b/firestore-counter/clients/node/index.js
@@ -38,10 +38,10 @@ module.exports = class Counter {
     const shardsRef = doc.collection(SHARD_COLLECTION_ID);
     this.shards[doc.path] = 0;
     this.shards[shardsRef.doc(this.shardId).path] = 0;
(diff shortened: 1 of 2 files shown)


dackers86 (approved):

dackers86:
Holding off on any releases of this extension for now.

Waiting until unit tests and reliable way of e2e testing can be completed. Current unit tests provide typescript errors using node 14/16.

cabljac:
closed in favour of https://github.com/firebase/extensions/pull/1376 (for testing purposes)

Issue #838, opened 2021-12-31
Add ability to replace original file, including name.


MrAndrew:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

### [REQUIRED] Step 2: Extension name

This feature request is for extension: `storage-resize-images`

### What feature would you like to see?

I would like the ability to configure the function to simply replace the original file, including the file name so that the appending of the ratio isn't done so automatically. For example, `myPhoto.jpg` -> `myPhoto.webp` instead of `myPhoto_1080x1080.webp`. 

I see the modified file name on line 122 of resize-image.ts includes the file size by default. I'm sure this is due to the variable amount of file resizes the function can create, so modifying more than this line is necessary. 

### How would you use it?

I'd like to use this feature like I would normally use the extension function, just with an extra configuration option. For some use cases only one is desired or the original needs to be kept intact by name.

dackers86:
Duplicate of https://github.com/firebase/extensions/issues/653.

Issue #653, opened 2021-05-01
Is there a way stop renaming the image with  200x200 at the end? This would be a lot helpful


vishnusait:
### I installed this plugin in my project for resizing the images
This works great except for the renaming of the original file name. I wish there is a checkbox to stop renaming the file if I want to delete the original and keep the resized one.
Please help! 
Thanks!

dackers86:
Thanks @vishnusait. Updated for consideration

kalmicka:
I second that idea! Alternatively it would be great to be able to set the name ourself. Using two different sizes someone could e.g. call 100x100 = small and 1024x1024 = large. Afterwards the developer could still edit the image sizes and still have a consistent result (name of the image).

zaidkazi:
#658 I believe as part of this feature 658 should be considered at the same time.

If we remove the 200x200 from a file name by nature many applications will want to treat the original image and the new optimized image as the same. Having the deletion **immediately**  after does not promote good app behaviour in many cases as thats likely the time apps will need the image the most and there is a slight chance of the gap in between where the image being overwritten that an application may request it.

Would having a delay field of letting us choose when to delete  and resize images be beneficial?

i14h:
@kalmicka @zaidkazi @vishnusait  can you explain what's the user need for this feature and what problem would it solve?

zaidkazi:
@jhuleatt Hello i think thats a good alternative but it would still be easier to not have to worry about what the file name will be in general. If i upload a picture and my application wishes to consume it there is a chance my image will be

image.png
or 
image_200x200.png 

(Depending on if its been processed and deleted in time)

i would rather my application does not have to guess

zaidkazi:
@i14h  sorry forgot to tag you aswell

i14h:
@zaidkazi if we resize in place keeping the original image name it might mess up caching not only in GCS but on all the other potential proxies in between the client and the image. Can your application make a second request in case the original request for the image returns a 404?

kalmicka:
> @kalmicka @zaidkazi @vishnusait can you explain what's the user need for this feature and what problem would it solve?

[...]

PR #934, opened 2022-04-04
feat(storage-resize-images): added a new param to specify resized images naming convention

yamankatby:
fixes: #653
fixes: #838 (duplicate issue)

### DIFF
--- a/storage-resize-images/extension.yaml
+++ b/storage-resize-images/extension.yaml
@@ -148,6 +148,24 @@ params:
     default: "200x200"
     required: true
 
+  - param: NAMING_CONVENTION
+    label: Naming convention for resized images
+    description:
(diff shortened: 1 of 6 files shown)


storage-resize-images/extension.yaml
```diff
@@ -148,6 +148,24 @@ params:
     default: "200x200"
     required: true
 
+  - param: NAMING_CONVENTION
```
Salakar, line 151: If we make changes to the .yaml file then the extensions README.md also needs regenerating (see https://github.com/firebase/extensions/blob/next/storage-resize-images/functions/package.json#L13) and changes committed as part of this PR, thanks!

yamankatby:
Closing this PR without merging as the solution provided is not generic enough.

PR #939, opened 2022-04-07
fix(storage-resize-images): check if the outputOptions param is undefined to prevent unnecessary error log

yamankatby:
Currently, if no OUTPUT_OPTIONS is provided then an error log is printed, even though this param is optional: 
image

This PR updates to first check if a value has actually been set.


Salakar (approved):

joehan (approved):
Good catch!

Issue #945, opened 2022-04-21
[FS-BQ-Export] Cannot export whole rows, missing most of rows.


jamestalano:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

[firestore-bigquery-export] 0.1.22

#### Steps to reproduce:

I have updated the collections on firestore from iOS device, swift app by batch writing. It was working well to get the right result when I use normal writing to firestore, but only for batch writing, I cannot see some of rows. It seems like only 2 of 10 rows are exported.

##### Actual result
Only display only 90 rows.
<img width="521" alt="Screen Shot 2022-04-21 at 11 34 21 AM" src="https://user-images.githubusercontent.com/43224497/164508501-9ed38a18-132f-49f6-8910-8f80c92c6113.png">
<img width="744" alt="Screen Shot 2022-04-21 at 11 34 38 AM" src="https://user-images.githubusercontent.com/43224497/164508508-18a02294-62ee-417b-ba68-8c56d877c61c.png">
<img width="954" alt="Screen Shot 2022-04-21 at 11 35 55 AM" src="https://user-images.githubusercontent.com/43224497/164508705-b4d8316b-e80c-4f64-b04c-a4b425efb726.png">

dackers86:
Hi @jamestalano 

Could you provide an example of some output from the extension logs?

jamestalano:
Hi @dackers86 
Thanks for your reply. I already attached the screenshots. What kind of example you need please?

dackers86:
The above screenshot only appears to show timestamps and no output messages.

We have a known issue reporting logs files state that an `IF_MATCH` condition has not been met I just wanted to confirm if this issue was related or if a different issue has been encountered. Thanks!

jamestalano:
[...]

Issue #923, opened 2022-03-18
fsexportbigquery: Error when mirroring data to BigQuery ApiError: Precondition check failed.


jeromesth:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: `firestore-bigquery-export
- Extension version: `0.1.21`
- Configuration values (redact info where appropriate):
  - Cloud Functions location: us-central1
  - BigQuery Dataset location: us
  - Collection path: ----
  - Wildcard Column field with Parent Firestore Document IDs (Optional): false
  - Dataset ID: ----
  - Table ID: ---
  - BigQuery SQL table Time Partitioning option type: MONTH
  - BigQuery Time Partitioning column name (Optional): Parameter not set
  - Firestore Document field name for BigQuery SQL Time Partitioning field option (Optional): createdAt
  - BigQuery SQL Time Partitioning table schema field(column) type (Optional): TIMESTAMP

### [REQUIRED] Step 3: Describe the problem
Looking at the logs of the extension, it has been throwing these errors multiple times per day. See fulls stack trace below

```
Error when mirroring data to BigQuery ApiError: Precondition check failed.
    at new ApiError (/workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/node_modules/@google-cloud/common/build/src/util.js:58:15)
    at Util.parseHttpRespBody (/workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/node_modules/@google-cloud/common/build/src/util.js:193:38)
    at Util.handleResp (/workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/node_modules/@google-cloud/common/build/src/util.js:134:117)
    at /workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/node_modules/@google-cloud/common/build/src/util.js:432:22
    at onResponse (/workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/node_modules/retry-request/index.js:228:7)
    at /workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/node_modules/teeny-request/build/src/index.js:233:13
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  code: 412,
  errors: [
    {
      message: 'Precondition check failed.',
      domain: 'global',
      reason: 'failedPrecondition'
    }
  ],
  response: Gunzip {
    _writeState: Uint32Array(2) [ 16110, 0 ],
[...]

Issue #910, opened 2022-03-11
[firestore-bigquery-export] Table not updated (while raw changelog is updated)


tgpsantos:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: firestore-bigquery-export@0.1.21 
- Extension version: 0.1.21 

### [REQUIRED] Step 3: Describe the problem

I added a new sync export collection to bigquery using firebase export extension. 

I started by:
- creating export function to my_project.groups.all_groups
- after it completed, added all the docs from the original collection to the table (followed [these steps](https://cloud.google.com/bigquery/docs/loading-data-cloud-firestore))

The problem is that whenever a new change occurs, my all_groups table is not updated. Strangely, the all_groups_raw_changelog is updated correctly.

Then I went to the function to see the logs:
![imagem](https://user-images.githubusercontent.com/48353485/157901572-fc258323-560b-4a06-bf6b-585bea274cf6.png)

I have no idea if the problem is associated to this partition warning or not, but I created a new function without the TIME partitioning set. Same problem occurred afterwards.

I have no idea what I messed up in the process, I'd be grateful if you could help me. Happy to provide more info!

dackers86:
HI @tgpsantos.

Can you confirm that your BigQuery datasets/tables and extension configuration are pointing to the same location?

Additionally, on your extension configuration we now have a pre-filled option for the project you would like to sync with, does this have the correct name?

tgpsantos:
> HI @tgpsantos.
> 
> Can you confirm that your BigQuery datasets/tables and extension configuration are pointing to the same location?
> 
> Additionally, on your extension configuration there is have a pre-filled option for the project you would like to sync with, does this have the correct name?

I've checked the extension config and even compared it with a different export I have for another collection. 
I checked again and those params seem to be defined correctly.
Would the raw_changelog table be updated if any of those params was incorrect?

dackers86:
> I've checked the extension config and even compared it with a different export I have for another collection. I checked again and those params seem to be defined correctly. Would the raw_changelog table be updated if any of those params was incorrect?

That seems ok, the location of the function or database would only be an issue if they are both showing incorrect results.

Apologies if breaking this down step by step, the logs suggest the data has been written - I just want to make sure we are looking in the right place.

Can you ensure that you are checking the `all_groups_raw_latest` view to see any updated information. Additionally, can you add a `new document` to the Firestore collection to produce a newly inserted records? Thanks!

tgpsantos:
[...]

Issue #606, opened 2021-02-21
Error when mirroring data to BigQuery: Resource did not meet condition IF_MATCH


guilhermeagostinelli:
- Extension name: `firestore-bigquery-export`
- Extension version: v0.1.12

It was running perfectly fine on v0.1.5 but after upgrading to v0.1.12 the error below has already appeared twice in the logs. There have been executions before and after the error which executed successfully.

```
Error when mirroring data to BigQuery { Error: Resource positiv-app:firestore_export_conclusions.conclusions_raw_changelog did not meet condition IF_MATCH
    at new ApiError (/workspace/node_modules/@google-cloud/common/build/src/util.js:58:15)
    at Util.parseHttpRespBody (/workspace/node_modules/@google-cloud/common/build/src/util.js:193:38)
    at Util.handleResp (/workspace/node_modules/@google-cloud/common/build/src/util.js:134:117)
    at retryRequest (/workspace/node_modules/@google-cloud/common/build/src/util.js:432:22)
    at onResponse (/workspace/node_modules/retry-request/index.js:214:7)
    at /workspace/node_modules/teeny-request/build/src/index.js:233:13
    at process._tickCallback (internal/process/next_tick.js:68:7)
  code: 412,
 ```

dackers86:
Hi @guilhermeagostinelli 

A history of this issue can be found here https://github.com/firebase/extensions/issues/235

The original fix is here https://github.com/firebase/extensions/pull/398. 

Can you confirm that you have a `document_id ` column in your BQ tables?

dackers86:
Also could you confirm this is a recurring issue?

There seems to be a known issue when first implemented https://github.com/firebase/extensions/pull/398#issuecomment-666447313

guilhermeagostinelli:
Hi @dackers86 , thanks for the reply.

> Can you confirm that you have a document_id column in your BQ tables?

Yes, the column is there and, by the way, all documents before the migration date have a NULL value.

> Also could you confirm this is a recurring issue?

It was first seen on 2/19/21 and last seen on 2/21/21
There have been 3 occurrences until now.

dackers86:
Thanks @guilhermeagostinelli.

Although not ideal, could clearing your BQ tables and re-syncing solves this particular case?

Additionally, do you have any errors in the extensions logs? Hopefully they could provide a hint on any issues that are occurring?

SuryaSrikanthsml:
Here is the error in extension's cloud function logs

[...]

PR #946, opened 2022-04-22
fix(firestore-bigquery-export): added updates and tests for improving change tracker initialisation

dackers86:
fixes: https://github.com/firebase/extensions/issues/923
fixes: https://github.com/firebase/extensions/issues/910
fixes: https://github.com/firebase/extensions/issues/945
fixes: https://github.com/firebase/extensions/issues/965
fixes: https://github.com/firebase/extensions/issues/975
fixes: https://github.com/firebase/extensions/issues/606


firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
```diff
@@ -191,17 +192,29 @@ export class FirestoreBigQueryEventHistoryTracker
...
 
-        if (datasetExists && tableExists) {
+        catch (ex) {
           clearInterval(handle);
-          return resolve(table);
+          logs.failedToInitializeWait(ex.message);
         }
-      }, 500);
+
+      }, 5000);
```
Ehesp, line 217: Small nit; before it was 500, now 5000 - is there any side effect we need to mention here? e.g. documentation

dackers86: The original `5000` was introduced to give enough time should the `tables` need enough to create.

This seemed quite lengthy and was reduced with testing for performance (time to taken to receive writes to the database). There is no side effect, but now the PR has been completed. Perhaps `5000` should be reintroduced to limit the chance of writing before a table has been created.

`5000` certainly seems safer than `500`.

meyerovb: does this mean any write firestore will take a minimum of 5 seconds to propagate to bigquery in the future?

dackers86: @meyerovb This would only occur if an error was found with the BigQuery api when writing the record. 

The errors typically appear when a large amount of writes happen simultaneously.

i14h (approved):
LGTM - as discussed offline, lets change the table check approach to a try/catch style where we remove the table check per each call.

jeromesth:
Hello! What is the timeline to have this merged and deployed so we can upgrade our plugins? Thanks!

patwhite:
Is there any way to pass along feedback to the firestore product team, reading this issue and seeing how literally > 50% of replications to bigquery just fail with no real error messages (unless you dig into the cloud function logs), there REALLY needs to be a native CDC log from firestore or native integration to BQ, I don't see how this cloud function method is robust enough for any serious projects.

dreinon:
Hi, is this already released?

dackers86:
Not yet. The release can be tracked here https://github.com/firebase/extensions/pull/1005

PR #983, opened 2022-05-30
chore(*): updated ci node matrix to node 14 and 16

dackers86:
The latest Firebase cli version no longer supports node 12. https://www.npmjs.com/package/firebase-tools/v/11.0.1

Work carried out last year now means no extension is below node 14, so we can now safely upgrade the ci to represent node  versions `14` and `16`


dackers86:
Closing, updates have already been added through https://github.com/firebase/extensions/pull/995

PR #986, opened 2022-06-01
Refactored tests to work with emulator suite

joehan:


firestore-send-email/functions/package.json
```diff
@@ -18,7 +18,7 @@
     "@types/node": "^12.6.9",
     "@types/nodemailer": "^6.2.1",
     "firebase-admin": "^9.3.0",
-    "firebase-functions": "^3.16",
+    "firebase-functions": "^3.20",
```
joehan, line 21: This might not actually be necessary in retrospect - the outdated version was causing problems when the Emulator Suite was trying to treat this like CF3 functions, but I suspect it won't matter now that this is actually being treated as an extension.

firestore-send-email/integration-test/firebase.json
```diff
@@ -1,24 +1,23 @@
...
       "port": 9199
     },
-    "functions": {
-      "port": 9002
-    },
     "ui": {
       "enabled": true
     }
+  },
+  "firestore": {
```
joehan, line 19: This extension uses firestore, so I needed to add this block + rules files so that the firestore emulator would run.

firestore-send-email/integration-test/firestore.rules
```diff
@@ -0,0 +1,198 @@
+rules_version = '2';
```
joehan, line 1: Looks like these were exported from the extensions-testing project by firebase init. They can probably be replaced with default rules files for the purposes of this test.

firestore-send-email/integration-test/firestore.indexes.json
```diff
@@ -0,0 +1,58 @@
+{
```
joehan, line 1: Looks like these were exported from the extensions-testing project by firebase init. They can probably be replaced with default rules files for the purposes of this test.

dackers86:
> 

Thanks! This was for debugging purposes and needed to be changed back.

dackers86:
> Note that I also changed functions.yaml back to extension.yaml - not sure why this was switched over in the first place. Functions.yaml is an internal implementation detail for CF3 functions.


Thanks! This was for debugging purposes and needed to be changed back.

.github/workflows/test.yml
```diff
@@ -30,14 +30,10 @@ jobs:
...
-        run: firebase --open-sesame extdev
-      - name: Run 'firestore-send-email' extension emulator
+      - name: Run 'firestore-send-email' tests on the Extension emulator
         run:
-          cd ./firestore-send-email/functions/__tests__/emulator && sh
-          ./start.sh
-      - name: Run tests with coverage
-        run: npm run test-coverage
+          cd ./firestore-send-email/integration-test/ && firebase emulators:exec
+          "cd ../../ && npm run test-coverage" -P demo-test
```
dackers86, line 36: We could potentially remove this from the CI config and keep it local the extension. If the tests are ran locally, currently there is no emulator and the tests will fail.

PR #1003, opened 2022-06-15
feat(storage-resize-images): add an option to make the resized images public

yamankatby:
Add an opt-in param to make the generated resized images public automatically, so they can be accessed by URL like: https://storage.googleapis.com/{bucket}/{path}

fixes: #140
fixes: https://github.com/firebase/extensions/issues/1029


dackers86 (approved):

yamankatby:
Merged to #1067

Issue #1055, opened 2022-07-20
`delete-user-data`: Paths aren't recognized when there is a space after a comma


rafoolin:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **delete-user-data**
- Extension version: **0.1.13**
- Configuration values (redact info where appropriate):
  - **Cloud Storage paths**: {DEFAULT}/files/{UID},{DEFAULT}/profile_images/{UID}.png

### [REQUIRED] Step 3: Describe the problem

I checked [this issue](https://github.com/firebase/extensions/issues/531), I had the same problem and changed `{DEFAULT}` to `project-id.appspot.com`. This didn't help.

The problem is that folder with `UID` at `{DEFAULT}/files/{UID}` will be deleted, but `UID.png` at `{DEFAULT}/profile_images/{UID}.png` won't be recognized and in the log it says the path is not found.

#### Steps to reproduce:

What happened? How can we make the problem occur?
I noticed if we have space after `,` between paths it fails. For example:

```text
{DEFAULT}/a/{UID}, {DEFAULT}/b/{UID}.png, {DEFAULT}/c/{UID}
```

`a` will be OK, but `b` and `c` will not be found by the extension.

When I removed the space after `,` it is OK and `a`, `b` and `c` are recognized by the extension.

##### Expected result
Maybe it would be better if the spaces were ignored or at least mentioned in the description.

##### Actual result
Comma-separated paths will not be recognized correctly if there is a space after them.

ankit-programmer:
@dackers86 I am working on it. I am planning to trim the paths before processing.
Should I go ahead with the above solution or you have something else in your mind?

dackers86:
@ankit-programmer Thanks!

You're correct that there shouldn't be any spaces. A trim solution would be great 👍🏻

ankit-programmer:
[...]

PR #1056, opened 2022-07-20
Remove space from start and end of path.

ankit-programmer:
fixes: https://github.com/firebase/extensions/issues/1055


dackers86:
Thanks @ankit-programmer.

We could potentially use some test cases adding to ensure this works as expected. Happy to support on this as it could require updates from the new emulator.

dackers86:
Due to risks and a potential `richer params` update in the future mentioned [here](https://github.com/firebase/extensions/issues/1055#issuecomment-1315703290). 

We will close this PR for now, but may be used for future reference.

PR #1133, opened 2022-08-08
feat(storage-resize-images): include original image data in events

### LINKED ISSUE #1019: resizing image extension feature: Include original image data in events
### Step 1: Are you in the right place?
Yes, 

### Step 2: Extension name

This feature request is for extension: `storage-resize-images`

I would like the `firebase.extensions.storage-resize-images.v1.complete` events to contain information about the original uploaded image as well as the resized ones. At the moment the event contains an `outputs` list with the resized images, but it doesn't give developers anything about the image that triggered the resize. I am imagining there could be an `input` object alongside the `outputs` list which contains the path to the original image and its size.

### How would you use it?

My way of using it is a bit niche, but I believe there are more common usecases.
I want to know how much the image was scaled down by and store this for the user. In my app users will be drawing on an image and it will show on the thumbnail by scaling down the drawing by the same factor that the image has been scaled down.
At the moment I have to infer the storage location from the elements in the `outputs` list and read the image again.

Another usecase would just be to move or delete the original image if it is above a certain size.

yamankatby:
Include the original image metadata object in the event.
Metadata object shape: https://firebase.google.com/docs/reference/functions/providers_storage.objectmetadata
fixes #1019

### DIFF
--- a/storage-resize-images/functions/src/index.ts
+++ b/storage-resize-images/functions/src/index.ts
@@ -156,6 +156,7 @@ export const generateResizedImage = functions.storage.object().onFinalize(
           type: "firebase.extensions.storage-resize-images.v1.complete",
           subject: filePath,
           data: {
+            input: object,
             outputs: results,
           },
         }));


dackers86 (approved):

stevebread:
This feature has been undone by 
https://github.com/firebase/extensions/commit/82a1163eead749e246f4c5c89ad7c1bc5d55760a#diff-4a4b11d7d4a93db4e95b9213c754d7b8fb225812bceec8d7f8d1c38c83bd195fR121

dackers86:
Hi @stevebread 

Thanks for posting this issue, we'll look straight into it.

dackers86:
PR tracking here https://github.com/firebase/extensions/pull/1369

PR #1147, opened 2022-08-19
Minor tweaks to the publish script

pavelgj:
* changed `npm install` to `npm ci`
 * introduced  extra flag to pass additional flag for the `firebase ext:dev:publish` call
 * fixed error code handling

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -29,7 +29,7 @@ fi
 REPO_ROOT="`( cd \`dirname \"$0\"\` && cd .. && pwd )`"
 
 cd "$REPO_ROOT"
-npm install
+npm ci
 if [ ! -z "$IGNORE_TEST_FAILURES" ]
 then
   echo "Ignoring failing tests."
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -47,9 +47,10 @@ for i in ${EXTENSIONS[@]}; do
 
   cd "$REPO_ROOT/$i"
   set +e
-  firebase ext:dev:publish $PUBLISHER_ID/$i --non-interactive --force
-  set -e
+  firebase ext:dev:publish $PUBLISHER_ID/$i $FIREX_PUBLISH_EXTRA_FLAGS --non-interactive --force
```
huangjeff5, line 50: how is FIREX_PUBLISH_EXTRA_FLAGS set exactly?

pavelgj: the caller of this script can optionally define this env var and if defined it will be passed along.

huangjeff5 (approved):

Issue #1157, opened 2022-08-23
Migrate storage-resize-images `IMG_BUCKET` param to type `selectResource` 


pavelgj:
Docs: https://firebase.google.com/docs/extensions/alpha/parameters#resourcetype-field

yamankatby:
Hi @pavelgj,
I couldn't view the link provided, it's 404, can you please check it again or provide a code example?

PR #1160, opened 2022-08-26
feat(storage-resize-images): update IMG_BUCKET param type to selectResource

yamankatby:
fixes: #1157

### DIFF
--- a/storage-resize-images/extension.yaml
+++ b/storage-resize-images/extension.yaml
@@ -129,7 +129,8 @@ params:
       original images are either kept or deleted.
       It is recommended to create a separate bucket for this extension. For more information,
       refer to the [pre-installation guide](https://firebase.google.com/products/extensions/storage-resize-images).
-    type: string
+    type: selectResource
+    resourceType: storage.googleapis.com/Bucket
     example: my-project-12345.appspot.com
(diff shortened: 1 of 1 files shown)


pavelgj:
LGTM! That's really all that's needed. The update flow will handle change of param type from `string` to `selectResource`.

pavelgj (approved):

pavelgj:
Can you please merge this?

TangrisJones:
Hello @yamankatby & @pavelgj 

I have some questions about this PR. I believe this has cause a non-backwards compatible change in the extension. Can you perhaps shed some light? 

Previous to this version, I was able to specify any bucket in my GCP storage (v0.1.29). Now however, if I go to upgrade my extension, the "configure extension" section is actually forcing me to change the already configured bucket to the standard firebase appspot bucket (i.e. my-project-12345.appspot.com). Obviously, I can't do this because it will just break the configuration with my app... 😓 

Is this an intentional side effect? Or am I just incredibly stupid and missing something obvious haha? 

If this is indeed how it should be, I must say that this is very rough user experience and IMO should be reverted if possible.... 🤔 

Would love to hear your thoughts! Thank you :)

Issue #876, opened 2022-02-05
[firestore-counter] Counter extension run locally: Transaction to aggregate shards failed. TypeError: Cannot read properties of undefined (reading 'length')


polarby:
### Configuration

- Extension name: Distributed counter
- Extension version: 0.2.4 (19 Oct 2021)
- Configuration values (redact info where appropriate):
LOCATION="europe-west3"
INTERNAL_STATE_PATH= "_firebase_ext_/sharded_counter"
SCHEDULE_FREQUENCY= 1

### Problem

*This is a follow-up bug report to [this](https://stackoverflow.com/questions/70836970/implementation-of-distributed-counter-extension-for-local-emulator) question*

During the Implementation of Distributed-Counter-Extension for local emulator I was able to run the scheduled function (`controllerCore();`) via the [firebase shell](https://podzemski.com/2020/10/14/how-to-trigger-and-run-firebase-scheduled-functions-from-localhost/) and it will only work for the first aggregation. The next one will throw the following error: 

```
Transaction to aggregate shards failed. 
TypeError: Cannot read properties of undefined (reading 'length')\n    at /Users/**/Documents/**/**-Backend/functions/internal/counter/controller.js:118:44\n . 
```

What I have encountered is, that when removing  `_firebase_ext_` from the local firestore it will work again for one aggregation again (and the directory will be created again). 
When running the controllerCore again after the error I will get a timeout warning of my worker function.

```
functions: Your function timed out after ~60s. To configure this timeout, see
      https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation.
>  /usr/local/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:618
>              throw new Error("Function timed out.");

```

#### Steps to reproduce:

Implement the Counter extension to the local emulator
with the following export changes in the index file as described [here](https://stackoverflow.com/a/70837986/11003497):

```js
exports.controllerCore = functions.pubsub.schedule(`every ${Config.counter.SCHEDULE_FREQUENCY} minutes`).onRun(async () => {}
```

```js
exports.worker = functions.firestore.document(`${Config.counter.INTERNAL_STATE_PATH}/workers/{workerId}`).onWrite(async (change, context) => {}
```

```js
exports.onWrite = functions.firestore.document(`${Config.counter.INTERNAL_STATE_PATH}/workers/{workerId}`).onWrite(async (change, context) => {}
```

PaulLVoelker:
[...]

PR #1163, opened 2022-08-29
Resolve local emulator error "Transaction to aggregate shards failed"

polarby:
This is a proposed pull request for [this](https://github.com/firebase/extensions/issues/876#issue-1124877924) issue. Note that this was only tested on the js version, and was simply rewritten for ts purposes.


polarby:
Updated GitHub username in CLA agreement. Should work now!

PR #1195, opened 2022-09-13
ci(*): sync package-lock.json files

yamankatby:
For some reason, some of the package-lock.json files are not synced.

Screen Shot 2022-09-13 at 12 28 07

Just run `npm i`


dackers86:
This could be down to the local node version when generated previously for those packages.

Is this version `node/14.19.1`?

yamankatby:
It's node `v14.20.0`

yamankatby:
Hey @dackers86,
seems the same issue using the node version mentioned (`v14.19.1`)

dackers86:
That's ok. We can keep this as a PR for the release...

Although we do regenerate the packages before each release, this could still be useful.

dackers86:
Closing. This will be re-ran before the next release.

PR #1206, opened 2022-09-14
fix(send-email): read and write delivery status from inside a single transaction

Salakar:
TLDR; This helps avoid duplicate emails being sent due to OOO function calls and duplicate/retried function calls.


firestore-send-email/functions/src/index.ts
```diff
@@ -229,9 +214,23 @@ async function preparePayload(payload: QueuePayload): Promise<QueuePayload> {
...
+    return;
+  }
+  let payload = snapshot.data();
+  // Only attempt delivery if the payload is still in a valid delivery state.
+  if (
+    !payload.delivery ||
+    !(
+      payload.delivery.state === "PENDING" || payload.delivery.state === "RETRY"
+    )
+  ) {
```
Salakar, line 230:
```suggestion
  if (!payload.delivery ||  payload.delivery.state !== "PROCESSING") {
```

firestore-send-email/functions/src/index.ts
```diff
@@ -278,76 +272,146 @@ async function deliver(
 
   // Wrapping in transaction to allow for automatic retries (#48)
   return admin.firestore().runTransaction((transaction) => {
+    // We could check state here is still PENDING or RETRY, but we don't
```
Salakar, line 275:
```suggestion
    // We could check state here is still PROCESSING, but we don't
```

firestore-send-email/functions/src/index.ts
```diff
@@ -278,76 +272,146 @@ async function deliver(
...
-      return deliver(payload, change.after.ref);
+        return true;
+      }
+
+      // We don't know what the state is, so we can't do anything. This should never happen.
+      return false;
+    });
+
+  if (shouldAttemptDelivery) {
+    await deliver(ref);
```
huangjeff5, line 397: If deliver fails, should we try/catch that and move delivery.state to RETRY?

firestore-send-email/functions/src/index.ts
```diff
@@ -278,76 +272,146 @@ async function deliver(
...
+): Promise<void> {
+  const ref = change.after.ref;
+
+  // A quick check to avoid doing unnecessary transaction work.
+  // If the record state is SUCCESS or ERROR we don't need to do anything
+  // transactionally here since these are the 'final' delivery states.
+  // Note: we still check these again inside the transaction in case the state has
+  // changed while the transaction was inflight.
+  if (change.after.exists) {
+    const payloadAfter = change.after.data();
```
huangjeff5, line 295: Let's say F1 and F2 are instances of the same event handler that are created at almost the same time (handler was triggered more than once for a single event). F1 and F2's change.after.data() will both show delivery.state=PENDING, even if F1 already updated delivery.state=PROCESSING.

Perhaps one way to address that is for both F1 and F2 to ignore change.after.data() and read the latest document directly from Firestore. There's still a race condition if F2 reads the latest document before F1 has committed the transaction, but the current implementation AFAICT guarantees that a duplicate function call will deliver twice.

Salakar: We only use change.after here in this one check and only to check for an already finalized status (status that should never change) - it's only here to avoid unnecessary transaction work, since a status of "SUCCESS" or "ERROR" should never change (and if by chance it should change externally by the user or something then it'd trigger another function anyway which would skip past this check)

firestore-send-email/functions/src/index.ts
Salakar: The deliver method has its own error handling and sets either SUCCESS or ERROR states. Setting it to RETRY on failure would be a extension behaviour change, since currently to RETRY a user must explicitly set the state to RETRY themselves external to the extension.

firestore-send-email/functions/src/index.ts
huangjeff5: Gotcha. Just making sure I understand how the duplicate logic works with an example, call it F1 and F2 which are duplicate instances of the same Firestore trigger.

F1 will put the document in PENDING state. When F2 calls deliver(), it will see that the document is in PENDING state, this it exits early. Am I understanding this correctly?

dackers86 (approved):

PR #1217, opened 2022-09-21
fix(firestore-bigquery-export): updating table metadata too often

yamankatby:
Add check to prevent updating the view metadata if nothing changed, same as the row-changelog table:
https://github.com/firebase/extensions/blob/58d8a5f0d43a61612df4a6ebdd87559d6570e563/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts#L343-L345


dackers86 (approved):

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/checkUpdates.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export async function tableRequiresUpdate(
+  table: Table,
+  config: FirestoreBigQueryEventHistoryTrackerConfig,
+  schemaFields: any,
+  documentIdColExists: boolean,
+  pathParamsColExists: boolean
+): Promise<boolean> {
+  // If the table doesn't exist, return early.
+  if (!table) {
```
Ehesp, line 14: The type here is `table: Table` which suggests this will never be called

dackers86: Agreed! Will remove.

cabljac (approved):
lgtm

Issue #1209, opened 2022-09-16
[storage-resize-images] Feature request: AVIF codec support


jimmyff:
WebP is great, but AVIF is even better, please could you add support? 🙏

Thanks

yamankatby:
Hey @jimmyff,
Thank you for raising this issue, good idea. I think this is possible as the Sharp library support it. I'll add it to the project tracker to be considered.

https://sharp.pixelplumbing.com/#formats

PR #1220, opened 2022-09-23

yamankatby:
fixes: #1209 
> **Note** test cases not included as the [image-type](https://www.npmjs.com/package/image-type) library used to check the file type while testing doesn't support AVIF codec.


Ehesp:
There's a section in the POSTINSTALL outline supported formats which needs adding too.

dackers86 (approved):

cabljac:
closed to reopen from a branch instead

Issue #1245, opened 2022-09-28
(Documentation): Path examples for image resizer are unclear


Rafer45:
### [READ] Step 1: Are you in the right place?
I believe so.

### [REQUIRED] Step 2: Describe your configuration
N/A

### [REQUIRED] Step 3: Describe the problem
The image resizer readme includes this bullet point:

> Paths that contain images you want to resize: Restrict storage-resize-images to only resize images in specific locations in your Storage bucket by supplying a comma-separated list of absolute paths. For example, to only resize the images stored in the `/users/pictures` and `/restaurants/menuItems` directories, specify the paths `/users/pictures,/restaurants/menuItems`. You may also use wildcard notation for directories in the path. For example, `/users/*/pictures` would match `/users/profile/pictures/image.png` as well as `/users/profile/pictures/any/sub/directory/image.png`. If you prefer to resize every image uploaded to your Storage bucket, leave this field empty.

The wildcard examples are a bit unclear:
1. "Paths that contain images you want to resize" is ambiguous. Is the "contains" relation transitive? i.e. will a match for `/users/foo` resize `users/foo/pirates/luffy.png`, since the latter is (in a way) "contained" in the former? 
1. The documentation seems to suggest that only wildcard notated paths match subdirectories, but I'd be surprised if that was the case. If it's not the case, could the docs be corrected? If it is the case, could that be made explicit?
1. The wildcard might be slightly clearer if the two examples we were contrasting *only differed in what is matched by the wildcard*. Instead of using `/users/profile/pictures/pic.png` and `/users/profile/pictures/foo/bar/pic.png`, (which introduces a difference of subdirectories that the match probably has nothing to do with), it would be clearer to show the wildcard behavior by showing that it matches `/users/foo/pictures.png` and `users/bar/pictures.png`.

Thanks for your work on extensions!

PR #1251, opened 2022-10-03
docs(storage-resize-images): clear paths examples

yamankatby:
fixes: #1245


dackers86 (changes requested):
This could be considered too confusing, perhaps this could be broken down a little more.

Also the foo bar examples, they do not necessarily need to be added compared to original real world examples.

cabljac:
closing in favour of https://github.com/firebase/extensions/pull/1373

PR #1265, opened 2022-10-14
Simplify parsed path passing

Rafer45:
The previous code involved many assignments at the caller's level and a helper function; this PR removes the helper function and moves the assignments into the function body using a destructuring assignment.

I'm not sure about the variable names I chose, so please be opinionated if you'd like. I kept the old variable names where possible to minimize the diff size.

Looking forward to your thoughts.


cabljac (approved):
lgtm!

Rafer45:
Should I squash and merge, or something else?

PR #1274, opened 2022-10-21
feat(firestore-send-email): add lifecycle events

yamankatby:


dackers86:
What do you think of adding some of the events around the

 `switch (payload.delivery.state) {`

This would contain an updated document based on the snapshot listener.

OnSuccess for example could contain the email payload based on the updated document.

Failure also returns null so would not be caught in the try catch.

dackers86 (approved):

PR #1279, opened 2022-10-26

cabljac:
Opening this PR with the workflow.yaml changed, to see if it's why formatting is breaking.


cabljac:
(it wasn't the reason. Formatting is passing locally, which is strange)

PR #1280, opened 2022-10-26
chore: update validate workflow

cabljac:
This PR updates the lint/format action:
- to use recommended versions of `actions/checkout` and `actions/setup-node` (see attached screenshot)
- to skip the postinstall step (which builds everything) when running `npm lint`

Screenshot 2022-10-26 at 10 39 00


cabljac:
ah you can't do what i wanted to do with npm version `6`...

Issue #1247, opened 2022-09-29
[firestore-bigquery-export] `path_params` is not included when importing firestore docs to BigQuery


jorroll:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **firestore-bigquery-export**
- Extension version: **0.1.25**
- Configuration values (redact info where appropriate):
  - **BIGQUERY_PROJECT_ID=${param:PROJECT_ID}**
  - **COLLECTION_PATH=organizations/{organizationId}/workspaces**
  - **DATASET_ID=comms_sandbox_firestore_export**
  - **DATASET_LOCATION=us-west3**
  - **LOCATION=us-west3**
  - **TABLE_ID=workspaces**
  - **TABLE_PARTITIONING=NONE**
  - **TIME_PARTITIONING_FIELD_TYPE=omit**
  - **WILDCARD_IDS=true**

Additional configuration information

- Script name: **`@firebaseextensions/fs-bq-import-collection`**
- Script version: **0.1.14**

### [REQUIRED] Step 3: Describe the problem

When importing a collection path containing wildcard params using `npx @firebaseextensions/fs-bq-import-collection`, I expect the imported records to contain a `path_params` column containing the wildcard path params. Instead, the imported records have a `path_params` column equal to `{}` (or an error is thrown, depending on the input). This is the same problem as reported in #947 (which was erroneously closed by #982 even though that PR does not fix this issue).

Note: this only affects the `@firebaseextensions/fs-bq-import-collection` import script. Incremental changes synced to BigQuery via the `firestore-bigquery-export` extension have the correct `path_params` values.

#### Steps to reproduce:

[...]

PR #1282, opened 2022-10-31
fix(firestore-bigquery-export): pass in final path of collection path to collectionGroup

cabljac:
This implements the fix suggested in https://github.com/firebase/extensions/issues/1247 

To close that issue we should also update documentation/CLI output.


package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "format": "prettier --write \"**/*.{js,md,yml,ts,json,yaml}\"",
-    "lint": "prettier --list-different \"**/*.{js,md,yml,ts,json,yaml}\"",
+    "lint": "prettier --check \"**/*.{js,md,yml,ts,json,yaml}\"",
```
cabljac, line 8:
```suggestion
    "lint": "prettier --list-different \"**/*.{js,md,yml,ts,json,yaml}\"",
```

cabljac:
Would like to add tests for this before it is approved, [since we have that capability now.](https://github.com/firebase/extensions/pull/1307) i have a concern about the solution, but i might be wrong.

cabljac:
Tests still need some tidying, and could add a couple more.

ciriousjoker:
I just tried out 1473d9bde73a3b50313fe96c6e187b993a19f089 and it works, but only after I manually remove the "sourceCollectionPath" question and put the value (`users/{uid}/workout`) directly into the variable by editing `config.js`.

Mby I'm just stupid, but trying to provide the value when getting asked about it via inquirer always resulted in "The collection path does not match the regular expression provided".

cabljac:
@ciriousjoker thanks, i've made a change to fix that in this PR.

cabljac:
closing in favour of updated PR https://github.com/firebase/extensions/pull/1414

cabljac:
closing in favour of https://github.com/firebase/extensions/pull/1414

PR #1285, opened 2022-11-01

cabljac:
fixes: https://github.com/firebase/extensions/issues/1209 (reopening https://github.com/firebase/extensions/pull/1220)


_emulator/pubsub-debug.log
```diff
@@ -0,0 +1,16 @@
+This is the Google Pub/Sub fake.
```
Lyokone, line 1: Shouldn't we delete this?

Lyokone (approved):

_emulator/pubsub-debug.log
cabljac: yeah not sure why this wasnt gitignored, will remove. thanks

Issue #757, opened 2021-10-01
[firestore-bigquery-export] Resources exceeded during query execution error


meyerovb:
- Extension name: firestore-bigquery-export
- Extension version: 0.1.16
- Configuration values (redact info where appropriate):
  -Cloud Functions location: us-central1
  -BigQuery Dataset location: us

Getting an error selecting from one of my ported collections.  all other much larger collections work fine.

Running: select * from firestore_export.mycollection_raw_latest limit 10
Produces: Resources exceeded during query execution: The query could not be executed in the allotted memory. Peak usage: 122% of limit. Top memory consumer(s): sort operations used for analytic OVER() clauses: 98% other/unattributed: 2%

No other query against anything in my db, other firestore ported collections, materialized views, scheduled generated tables from firestore collection ports, is throwing that error.

Workaround:
Added my own version of _latest view that I believe is functionally equivalent without the memory pressure:

```
create view myschema.mycollection_raw_latest as
SELECT document_name, document_id, timestamp, event_id, operation, data
FROM
  (
    SELECT *, ROW_NUMBER() OVER(PARTITION BY document_name ORDER BY timestamp DESC) rn
    FROM firestore_export.mycollection_raw_changelog
  )
WHERE operation<>'DELETE' and rn=1
```

i14h:
@meyerovb can you also provide the original `latest` view that the extension generated?

yamankatby:
> The first query creates a fake changelog table that will reproduce the error, pretend its just a really big document that gets modified very often. The second is your view that goes kaboom. I tried making my own version but you don't give imports event_ids so nope not happening...
> 
> create or replace table changelog_table as with a as ( select DATE_FROM_UNIX_DATE(CAST(RAND() * 10000 AS INT64)) as timestamp, GENERATE_UUID() as event_id, 'projects/myproject/databases/(default)/documents/mycollection/mydocument' as document_name, 'UPDATE' as operation, string_agg(concat(word, word), '') as data, 'mydocument' as document_id from `publicdata.samples.shakespeare` )select a.* from a, `publicdata.samples.shakespeare` limit 20000;
> 
[...]

PR #703, opened 2021-07-21
Update the SQL script used for generating latest snapshot.

sophie4869:
The old script uses FIRST_VALUE and OVER, which sorts the entire changelog and finds the first record for each document. It can result in a memory issue when running BigQuery reading from the latest snapshot. (Resources exceeded during query execution: The query could not be executed in the allotted memory. Peak usage: 110% of limit. Top memory consumer(s):  sort operations used for analytic OVER() clauses: 96%)

The updated script selects the maximum timestamp for each document_id, and joins back with the table by the latest timestamp instead.

PR #1289, opened 2022-11-08
fix(firestore-bigquery-export): update snapshot script

cabljac:
Reopened https://github.com/firebase/extensions/pull/703 (rebased on next).

- added relevant stress tests
- added tests for the case `Will this cover if there are null time stamps or the latest entry for a document has two entries with the same timestamp?` (mentioned [in this comment](https://github.com/firebase/extensions/pull/703#issuecomment-1302438030))

- added script to generate bq fixtures

fixes https://github.com/firebase/extensions/issues/757


firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/stresstest.test.ts
```diff
@@ -52,8 +54,68 @@ describe("Stress testing", () => {
...
+      );
+
+      const [job] = await bq.createQueryJob({
+        query,
+        useLegacySql: false,
+      });
+
+      const [rows] = await job.getQueryResults();
+
+      expect(rows.length).toEqual(1);
```
dackers86, line 82: Could we assert that the correct columns have been included in the result?

firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/stresstest.test.ts
```diff
@@ -52,8 +54,68 @@ describe("Stress testing", () => {
...
+    }, 240000);
   });
```
dackers86, line 120: As we are adopting a new SQL query for this task. Could we test the migration from the previous query, comparing the results to ensure that nothing has changed in the result set?

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/snapshot.ts
```diff
@@ -60,30 +60,22 @@ export function buildLatestSnapshotViewQuery(
     --   event_id: The id of the event that triggered the cloud function mirrored the event.
     --   data: A raw JSON payload of the current state of the document.
     --   document_id: The document id as defined in the Firestore database
-    SELECT
```
dackers86, line 63: For early adoption, would it be worth considering a configuration option to allow users to opt-in to the new Query. This could then be removed at a later date if the new format is widely adopted/successful?

dackers86 (changes requested):

firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/stresstest.test.ts
cabljac: ya can tie this into the other suggestion

dackers86 (approved):

PR #1291, opened 2022-11-10
docs: add further resources

pr-Mais:
Added a new section "**Further reading & resources**" to the following extensions:
- [storage-resize-images](https://github.com/firebase/extensions/tree/next/storage-resize-images)
- [firestore-send-email](https://github.com/firebase/extensions/tree/next/firestore-send-email)

Section added to both `POSTINSTALL` and `PREINSTALL`.


cabljac:
Actually this will probably need `npm run generate-readmes`  running since pre and post are edited

i14h (approved):

dackers86:
> Actually this will probably need `npm run generate-readmes` running since pre and post are edited
> 
> EDIT: or perhaps not, do we only need to do that in release @dackers86 ?

Yes, don't worry this will be taken care of as part of the release process 👍🏻

PR #1301, opened 2022-11-16
[firestore-translate-text] Adding lifecycle events

joehan:
Adds lifecycle events to firestore-translate-text. When enabled, this will translate existing, untranslated documents in the Firestore collection.

TODO
- [ ] Docs review from @kevinthecheung  - done in firebaseprivate
- [ ] Update package-lock.json once firebase-admin-node@11.3.0 is published
- [ ] Code review
- [ ] Republish and test to ensure that nothing broke while switching to published deps


firestore-translate-text/functions/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2020",
-    "lib": ["ES2020"],
+    "lib": ["ES2020", "dom"],
```
joehan, line 4: Not certain that this is correct - however, I needed to add it to resolve some build issues that i was experiencing when trying to npm i on this repo.

firestore-translate-text/extension.yaml
```diff
@@ -156,3 +166,26 @@ params:
...
+    options:
+      - label: Yes
+        value: true
+      - label: No
+        value: false
+
+
+lifecycleEvents:
+  onInstall:
+    function: fstranslatebackfill
```
dackers86, line 184: I'm not sure if we have the required params at this stage?

Display shows `Translating existing documents in ${param:COLLECTION_PATH}`

Attempted with:

 ${COLLECTION_PATH}
 ${param:COLLECTION_PATH}

This could be a deploy from local issue

joehan: Ahhh, I forgot that we show these preinstall as well. Fixing

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,8 +82,69 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
     }
-  }
-);
+  });
+
+export const fstranslatebackfill = functions.tasks
+  .taskQueue()
+  .onDispatch(async (data: any) => {
+    const runtime = getExtensions().runtime();
+    if (!process.env.DO_BACKFILL) {
```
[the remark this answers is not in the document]
joehan: Ahhh, turns out i tested that the lifecycle event doesn't fire when DO_BACKFILL is unset, but not when its set to 'no'. Fixing here and on the other PRs as well.

firestore-translate-text/extension.yaml
joehan: Confirmed that we aren't subbing this correctly at the moment. @ifielker is working on a backend fix to do so

firestore-translate-text/extension.yaml
```diff
@@ -60,6 +60,16 @@ resources:
         eventType: providers/cloud.firestore/eventTypes/document.write
         resource: projects/${param:PROJECT_ID}/databases/(default)/documents/${param:COLLECTION_PATH}/{messageId}
 
+  - name: fstranslatebackfill
+    type: firebaseextensions.v1beta.function
+    description:
+     Search your specified Cloud Firestore collection for existing, untranslated documents, translates the strings,
```
ifielker, line 66: Searches...

firestore-translate-text/extension.yaml
ifielker: Fix should land on Monday. Sorry about that.

firestore-translate-text/functions/src/index.ts
```diff
@@ -32,16 +34,17 @@ enum ChangeType {
   DELETE,
   UPDATE,
 }
-
+const DOCS_PER_BACKFILL = 200;
```
ifielker, line 37: Did we measure how long it takes to translate 200 docs?

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,8 +82,69 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+  });
+
+export const fstranslatebackfill = functions.tasks
+  .taskQueue()
+  .onDispatch(async (data: any) => {
+    const runtime = getExtensions().runtime();
+    if (!config.doBackfill) {
+      await runtime.setProcessingState(
+        "PROCESSING_COMPLETE",
+        "Existing documents were not backfilled."
```
ifielker, line 94: Might be better to say that "Existing documents were not translated because backfill was not requested". Maybe even add "If you would like to translate existing documents, reconfigure this extension and choose 'true' for 'Should existing, untranslated documents in the Firestore collection be translated as well?'"

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,8 +82,69 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+      logs.backfillComplete(newSucessCount, newErrorCount);
+      if (newErrorCount == 0) {
+        runtime.setProcessingState(
+          "PROCESSING_COMPLETE",
+          `Successfully backfilled ${newSucessCount} documents.`
+        );
+      } else if (newErrorCount > 0 && newSucessCount > 0) {
+        runtime.setProcessingState(
+          "PROCESSING_WARNING",
+          `Successfully backfilled ${newSucessCount} documents, failed to translate ${newErrorCount} documents.`
```
ifielker, line 137: Maybe add a "See function logs for translation errors"

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,8 +82,69 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+        );
+      }
+      if (newErrorCount > 0 && newSucessCount == 0) {
+        runtime.setProcessingState(
+          "PROCESSING_FAILED",
+          `Successfully backfilled ${newSucessCount} documents, failed to translate ${newErrorCount} documents.`
```
ifielker, line 143: "See function logs for translation errors" here too.

firestore-translate-text/functions/src/index.ts
```diff
@@ -98,6 +162,18 @@ const getChangeType = (
   return ChangeType.UPDATE;
 };
 
+const handleExistingDocument = async (
```
ifielker, line 165: We need to skip existing translations so users don't pay for duplicate translations.

firestore-translate-text/functions/src/logs/messages.ts
```diff
@@ -1,8 +1,13 @@
 export const messages = {
+  backfillComplete: (successCount: number, errorCount: number) =>
+    `Finshed backfilling translations. ${successCount} documents checked, ${errorCount} errors.`,
   complete: () => "Completed execution of extension",
   documentCreatedNoInput: () =>
     "Document was created without an input string, no processing is required",
   documentCreatedWithInput: () => "Document was created with an input string",
+  documentFoundWithInput: () => "Backfill fround document with an input string",
```
ifielker, line 8: fround -> found

firestore-translate-text/functions/src/logs/messages.ts
```diff
@@ -1,8 +1,13 @@
 export const messages = {
+  backfillComplete: (successCount: number, errorCount: number) =>
+    `Finshed backfilling translations. ${successCount} documents checked, ${errorCount} errors.`,
```
ifielker, line 3: Is this accurate? `$successCount documents checked` shouldn't it be "$successCount documents translated, ${errorCount} errors" Or does successCount count input -> target language for each target language? Maybe we should have an "already translated" too once we skip existing translations.

firestore-translate-text/functions/tsconfig.json
ifielker: how can we figure out if it is correct?

joehan: Some more context - without this, `npm run build` returns the following error: 
```
node_modules/@google-cloud/firestore/node_modules/google-gax/build/src/streamArrayParser.d.ts(27,23): error TS2304: Cannot find name 'AbortController'.
node_modules/@google-cloud/firestore/node_modules/google-gax/build/src/streamArrayParser.d.ts(28,19): error TS2304: Cannot find name 'AbortSignal'.
```

firestore-translate-text/functions/src/logs/messages.ts
joehan: Touched up the wording here. Now, this is showing the total number of docs translated sucessfully (into all languages) and the total number of docs that had some sort of error (ie successfully translated into 3 of 4 languages).

firestore-translate-text/functions/src/index.ts
joehan: Done and tested: 
Screen Shot 2022-11-18 at 1 46 30 PM

firestore-translate-text/functions/tsconfig.json
dackers86: An alternative solution would be to upgrade the `@types/node` package to a later version, for example:

 `"@types/node": "^18.11.9"`. 

It seems this is out of date since upgrading typescript [earlier this year](https://github.com/firebase/extensions/actions/runs/1918622451/jobs/2720983387).

firestore-translate-text/functions/src/index.ts
dackers86: I ran some testing on this amount of records, which only took 3 seconds.

There seems to be a delay on the initial processing status, with task queueing.

image

It will fail once, retry and then process the documents.

ifielker: Yes, it's normal for the task to fail the first few times. There's nothing we can do about this. It's because we've just created the queue, and the function, and granted function runner permissions, and those permissions take a bit to propagate. As long as the queue has default or better retries, the function will run as soon as the permissions have propagated.

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,13 +82,78 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+  });
+
+export const fstranslatebackfill = functions.tasks
+  .taskQueue()
+  .onDispatch(async (data: any) => {
+    const runtime = getExtensions().runtime();
+    if (!config.doBackfill) {
+      await runtime.setProcessingState(
+        "PROCESSING_COMPLETE",
+        "Existing documents were not translated because backfill was not requested."
```
ifielker, line 94: We can add "If you want to translate existing documents, reconfigure your extension and choose 'yes' for 'Translate existing documents?' Only missing translations will be requested." or something like that.

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,13 +82,78 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+    );
+    const newSucessCount =
+      pastSuccessCount +
+      translations.filter((p) => p.status === "fulfilled").length;
+    const newErrorCount =
+      pastErrorCount +
+      translations.filter((p) => p.status === "rejected").length;
+
+    if (snapshot.size == DOCS_PER_BACKFILL) {
+      const queue = getFunctions().taskQueue(
```
ifielker, line 118: Since this is a sample, we should consider adding some comments like "We are not done the backfill yet, enqueue the next chunk of work" or whatever.

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,13 +82,78 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+        "fstranslatebackfill",
+        process.env.EXT_INSTANCE_ID
+      );
+      await queue.enqueue({
```
ifielker, line 122: We might also want to log something here. If something goes horribly wrong and the task/function never executes for whatever reason, it will be good to know the last thing we enqueued. (and the values of the data we enqueued)

firestore-translate-text/functions/src/index.ts
```diff
@@ -79,13 +82,78 @@ export const fstranslate = functions.handler.firestore.document.onWrite(
...
+          "PROCESSING_COMPLETE",
+          `Successfully translated ${newSucessCount} documents.`
+        );
+      } else if (newErrorCount > 0 && newSucessCount > 0) {
+        runtime.setProcessingState(
+          "PROCESSING_WARNING",
+          `Successfully translated ${newSucessCount} documents, ${newErrorCount} errors. See function logs for specific error messages.`
+        );
+      }
+      if (newErrorCount > 0 && newSucessCount == 0) {
```
ifielker, line 140: Maybe we don't even need a condition around this one if we _return_ out of the other 2 setProcessingState ones.
Then it would also function as a catchall if something weird happened with DOC_PER_BACKFILL and we were counting down instead of up. It would also guarantee that we not leave the processing status as 'processing' forever. I mean that shouldn't happen, but if it does, we have a reasonable fallback that reports the error.

firestore-translate-text/functions/src/index.ts
ifielker: That makes sense. Anything that isn't backfill will be a change, and if the input changes, we need new translations for everything.

firestore-translate-text/functions/src/index.ts
```diff
@@ -144,11 +224,25 @@ const handleUpdateDocument = async (
...
+  snapshot: admin.firestore.DocumentSnapshot,
+  keepExistingTranslations: boolean = false
 ): Promise<void> => {
-  logs.translateInputStringToAllLanguages(input, config.languages);
+  const existingTranslations = extractOutput(snapshot);
+  const languages = config.languages.filter(
+    (targetLanguage: string): boolean => {
+      if (
+        !keepExistingTranslations ||
+        existingTranslations[targetLanguage] != undefined
```
ifielker, line 235: This if condition surprises me. Maybe I'm missing something, but I would have expected:

If {keepExistingTranslations && existingTranslations[targetLanguage] != undefined} then skip language.

firestore-translate-text/functions/src/index.ts
```diff
@@ -179,25 +273,32 @@ const translateSingle = async (
...
   let translations = {};
   let promises = [];
+  const existingTranslations = extractOutput(snapshot);
 
-  Object.entries(input).forEach(([input, value]) => {
-    config.languages.forEach((language) => {
+  Object.entries(input).forEach(([entry, value]) => {
+    const languages = config.languages.filter(
+      (targetLanguage: string): boolean =>
+        !keepExistingTranslations ||
```
ifielker, line 286: Same surprising if condition here

firestore-translate-text/functions/tsconfig.json
ifielker: That seems like a better option...

firestore-translate-text/functions/src/index.ts
joehan: That's reasonable. We should stay pretty conservative tho, because real documents can have many fields, each of which need to be translated individually into many languages. I'm going to increase this to 1500, which should be very safe

firestore-translate-text/functions/src/index.ts
joehan: Good catch - i got this mixed up while i was adding logging.

firestore-translate-text/functions/src/index.ts
joehan: Fixed to match the behavior of translateSingle (and include logging)

firestore-translate-text/functions/src/logs/index.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import { logger } from "firebase-functions";
+import { off } from "process";
```
ifielker, line 18: What is this for?

firestore-translate-text/functions/src/index.ts
joehan: Done

firestore-translate-text/extension.yaml
dackers86: Seems to be working now. Thanks!

ifielker (approved):

joehan:
Ran a larger scale test and am not happy with the results:
Screen Shot 2022-11-22 at 2 57 28 PM
I need to go offline for a bit, but i will dig into this further and make some improvements to our retry logic/error handling later.

joehan:
Refactored this to make 2 main changes:
- Instead of an individual transaction for every write, use firestore.BulkWriter
- Limit the number of parallel calls to the Google translate API, in order to avoid socket hangups.

With these changes, this is performing much better - was able to translate a 10k document collection with 0 errors.
Screen Shot 2022-11-28 at 4 06 05 PM

Going to spend a little more time to see if i can improve the speed, but I am comfortable releasing this in its current state.

joehan:
Parallelized this a little more, and successfully backfilled 60k docs with 0 errors
Screen Shot 2022-11-28 at 5 44 22 PM

firestore-translate-text/extension.yaml
```diff
@@ -156,3 +168,28 @@ params:
     example: translated
     default: translated
     required: true
+
+  - param: DO_BACKFILL
+    label: Translate existing documents
```
ifielker, line 173: Should we add a question mark at the end of the label?

firestore-translate-text/functions/tsconfig.json
ifielker: I saw you updated the @types/node, can we revert this change now?

joehan:
Another round of PR fixes, and tested some more error cases too:
Backfilling with a collection that doesn't exist:
Screen Shot 2022-11-29 at 11 17 16 AM
Adding an invalid language to the list:
Screen Shot 2022-11-29 at 11 19 16 AM

joehan:
@dackers86 Could you take another pass through this? We've made quite a few improvements, and I'd like to potentially release this later this week.

ifielker (approved):

dackers86 (approved):

PR #1302, opened 2022-11-16
[auth-mailchimp-sync] Add lifecycle events

joehan:
Add lifecycle events to add existing users to the specified Mailchimp audience.

TODO

 - [x] Docs review from @kevinthecheung - done in firebaseprivate
 - [x] Update package-lock.json once firebase-admin-node@11.3.0 is published
 - [ ] Code review
 - [x] Republish and test to ensure that nothing broke while switching to published deps


auth-mailchimp-sync/extension.yaml
```diff
@@ -155,3 +164,24 @@ params:
         value: pending
     default: subscribed
     required: true
+
+  - param: DO_BACKFILL
+    label: Import existing users into Mailchimp audience
+    description: >-
+      Do you want to add existing users to the Mailchimp audience?
```
kevinthecheung, line 171:
```suggestion
      Do you want to add existing users to the Mailchimp audience when you install or update this extension?
```

auth-mailchimp-sync/functions/src/index.ts
```diff
@@ -95,5 +99,83 @@ export const removeUserFromList = functions.handler.auth.user.onDelete(
...
+  });
+
+export const addExistingUsersToList = functions.tasks
+  .taskQueue()
+  .onDispatch(async (data) => {
+    const runtime = getExtensions().runtime();
+    if (!config.doBackfill) {
+      await runtime.setProcessingState(
+        "PROCESSING_COMPLETE",
+        "No existing documents imported into BigQuery."
```
ifielker, line 111: Optional: I think "No processing requested" would be better. So it's not confused for the case where there are no existing users.

auth-mailchimp-sync/functions/src/index.ts
```diff
@@ -95,5 +99,83 @@ export const removeUserFromList = functions.handler.auth.user.onDelete(
...
+      );
+      return;
+    }
+    const nextPageToken = data.nextPageToken;
+    const pastSuccessCount = (data["successCount"] as number) ?? 0;
+    const pastErrorCount = (data["errorCount"] as number) ?? 0;
+
+    const res = await getAuth().listUsers(100, nextPageToken);
```
ifielker, line 119: Since this is going to be a sample, let's add a comment above this explaining the choice of 100. (i.e. "We are sure we can complete 100 users within the 9 minute limit" or something like that).

auth-mailchimp-sync/functions/src/index.ts
```diff
@@ -95,5 +99,83 @@ export const removeUserFromList = functions.handler.auth.user.onDelete(
...
+      }
+    });
+    const results = await Promise.allSettled(mailchimpPromises);
+    const newSucessCount =
+      pastSuccessCount + results.filter((p) => p.status === "fulfilled").length;
+    const newErrorCount =
+      pastErrorCount + results.filter((p) => p.status === "rejected").length;
+    if (res.pageToken) {
+      const queue = getFunctions().taskQueue(
+        "fsimportexistingdocs",
```
ifielker, line 153: Has this been tested? Because the name of this function is not fsimportexistingdocs... so I don't think you'll find the taskQueue...

auth-mailchimp-sync/functions/src/index.ts
```diff
@@ -95,5 +99,83 @@ export const removeUserFromList = functions.handler.auth.user.onDelete(
...
+        process.env.EXT_INSTANCE_ID
+      );
+      await queue.enqueue({
+        nextPageToken: res.pageToken,
+        successCount: newSucessCount,
+        errorCount: newErrorCount,
+      });
+    } else {
+      logs.backfillComplete(newSucessCount, newErrorCount);
```
ifielker, line 162: Since this is a sample, we probably want to add a comment here that says something like "The backfill is now complete, set the processing status" or something like that to make it clear that this only happens at the end of the whole backfill, and not after every page. etc.

auth-mailchimp-sync/functions/src/index.ts
joehan: Will do - note that the real reason I chose 100 is 'I arbitrarily made it up. Need to actually verify that it is a reasonable choice.

auth-mailchimp-sync/functions/src/index.ts
joehan: This extension is not tested yet, since i did not realize mailchimp had a free plan.

joehan: Tested it out, and it is working after a few changes:
Screen Shot 2022-11-18 at 3 27 33 PM

auth-mailchimp-sync/functions/src/logs.ts
```diff
@@ -93,3 +93,9 @@ export const userRemoving = (
     `Removing user: ${userId} with hashed email: ${hashedEmail} from Mailchimp audience: ${audienceId}`
   );
 };
+
+export const backfillComplete = (successCount: number, errorCount: number) => {
+  logger.log(
+    `Finshed adding existing users to Mailchimp audience. ${successCount} users added, ${errorCount} errors.`
```
cabljac, line 99:
```suggestion
    `Finished adding existing users to Mailchimp audience. ${successCount} users added, ${errorCount} errors.`
```

cabljac (commented):
Looks good, going to pull the branch down and do a lil testing

joehan:
Added some improved error handling for users already in the audience: 
Screen Shot 2022-11-22 at 12 00 14 PM

cabljac (approved):
looks good to me

PR #1304, opened 2022-11-16
[storage-resize-images] Add lifecycle event

joehan:
Adds lifecycle event to resize existing images in the Storage bucket.

TODO

 - [x] Docs review from @kevinthecheung - done in firebaseprivate
 - [x] Update package-lock.json once firebase-admin-node@11.3.0 is published
 - [x] Code review
 - [ ] Republish and test to ensure that nothing broke while switching to published deps


storage-resize-images/extension.yaml
```diff
@@ -304,6 +314,28 @@ params:
...
+        value: false
+
 events:
   - type: firebase.extensions.storage-resize-images.v1.complete
     description: Occurs when image resizing completes. The event will contain further details about specific formats and sizes.
+
+lifecycleEvents:
+  onInstall:
+    function: backfillResizedImages
+    processingMessage: Resizing existing images in ${param:IMG_BUCKET}
```
ifielker, line 335: Has this been tested? I don't remember if we look for variables in the processing message... (but we should)

storage-resize-images/functions/src/filters.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  if (
+    config.excludePathList &&
+    startsWithArray(config.excludePathList, tmpFilePath)
+  ) {
+    logs.imageInsideOfExcludedPaths(config.excludePathList, tmpFilePath);
+    return false;
+  }
+
+  if (object.metadata && object.metadata.resizedImage === "true") {
```
ifielker, line 50: So if you originally configure your images to be resized to 100x100 and backfill, and then reconfigure your images to be resized to 200x200, won't this return the wrong answer? Because, yes it was resized, but it's not the size that you just requested in the configuration. After fixing this, we should test that it works with config of a different size.

ifielker (commented):
not fully reviewed, but I'm going to work on the issue we found for the backend, then I'll come back and review the rest...

storage-resize-images/functions/src/filters.ts
joehan: I was concerned about this too, but i tested and it behaves appropriately. object.metadata.resizedImage is only set on the resized images (not on the original). On the next pass through, we'll see the original image without object.metadata.resizedImage, and resize it again.

The one case that is a bit wonky here is when 'delete original image' is set to true. In that case, the extension won't see any un-resized images & won't trigger correctly here. Need to think though a solution that handle this case.

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
-      remoteFile = bucket.file(filePath);
-      logs.imageDownloading(filePath);
-      await remoteFile.download({ destination: originalFile });
-      logs.imageDownloaded(filePath, originalFile);
-
-      // Get a unique list of image types
-      const imageTypes = new Set(config.imageTypes);
-
-      // Convert to a set to remove any duplicate sizes
-      const imageSizes = new Set(config.imageSizes);
```
dackers86, line 127: The extension appears to be stuck on processing 3 images. Debugging has shown:

image

This will repeat on each new attempt at processing.

dackers86: The config aims to process 3 images in the query, so this could be a permanent loop of trying to process the first 3 , even when resized.

Could this be because of the initial boot up, when calling `enqueue`? 

`Deleting the original image` setting will also not work as the fn will attempt to cycle through the first 3 images, even if they are resized.

storage-resize-images/functions/src/filters.ts
ifielker: I think the correct solution is to document that only original images will be resized (i.e. we will not resize the already resized version of the original image). And add that if you choose to delete the originals after resizing, you would be required to upload them again if you decide you want a different size. Then they can make their choice between re-resizability and storage costs. Kevin will probably be helpful in how to phrase this well.

storage-resize-images/functions/src/index.ts
ifielker: So usually it takes a while for the function kicked off by our backend to run. That's because it does 7-10 (ish) retries lasting about 30 seconds - 1 minute, while it's waiting for the permissions we just granted (to run the function) to propagate. But if you're seeing the function actually running, then that permission has propagated. 

But the error looks like a permissions error. What is the rest of the error? Lacks ... ? I think this will help us determine if it's just waiting for a different permission to propagate or if that permission would not be granted based on the current configuration.

storage-resize-images/extension.yaml
ifielker: This should theoretically work now, but double check to make sure.

storage-resize-images/functions/src/index.ts
dackers86: First error:

image


Seconds error:

image

Strangely, the fn then does try to init, but then seems to stall...

image

joehan: Hmmm, that Cloud tasks enqueuer permission should be granted by default by the backend... Can you take a look at the service account in Cloud console and see what roles it has?

joehan: FWIW - i just successfully backfilled a bucket with 37 images (3 per task):
Screen Shot 2022-11-22 at 10 13 27 AM
Screen Shot 2022-11-22 at 10 14 37 AM


This seems like a possible backend issue? Either permissions not being granted correctly, or not propagating in time?

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
-              size,
-              objectMetadata: objectMetadata,
-              format,
-            })
-          );
-        });
+    const bucket = admin.storage().bucket(process.env.IMG_BUCKET);
+    const query = data.nextPageQuery || {
+      autoPaginate: false,
+      maxResults: 3,
```
ifielker, line 187: Why maximum of 3? That seems low? We should do as many as we can per chunk of work while still being guaranteed to finish resizing them in 9 minutes. The free tier for task executions is generous, but not infinite :)

storage-resize-images/functions/src/index.ts
dackers86: Can confirm his is now working after upgrading to the latest version of the CLI.

Tested on installation and configuration lifecycles.

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
-              originalFile,
-              parsedPath,
-              contentType,
-              size,
-              objectMetadata: objectMetadata,
-              format,
-            })
-          );
-        });
+    const bucket = admin.storage().bucket(process.env.IMG_BUCKET);
```
ifielker, line 184: Can we verify what happens when the bucket entered by the user is mis-spelled etc.?

storage-resize-images/extension.yaml
dackers86: Tested and working ok 👍🏻

dackers86 (approved):

storage-resize-images/functions/src/index.ts
joehan: This is to avoid out of memory errors when images are larger. Even when resizing single images, we've gotten reports of OOM errors (hence the param to control memory).

Also, when thinking about price concerns, i think its important to consider what the biggest costs will be. In this case, each task/function execution is going to do a lot of data egress/ingress in and out of Cloud Storage, and will consume a lot of compute time. If a user is exceeding the task execution free tier (1m executions) for even very small images (ie 1MB per image), they would incur the following charges:
- 3 million object inserts (Class A operation, $.05 per 10000) = $15
- 3 million object gets (Class B operation, $.004 per 10000) = $1.2
- 3 million MB storage (cost depends on region, but cheapest is ~$.02 per gb per month) = $60 per month
- as well as charge for the compute time of the functions

Tasks only cost $.4 per million after the first million.  With our current set up, this extension will incur at very least $75 of other costs for every $.4 spent on tasks.

storage-resize-images/functions/src/index.ts
joehan: The install fails, which is acceptable behavior IMO: 
Screen Shot 2022-11-29 at 2 02 11 PM

Its also worth noting that in console, this is a rich param, and we only let users select real buckets: 
Screen Shot 2022-11-29 at 2 04 08 PM

storage-resize-images/functions/src/filters.ts
kevinthecheung: I suggest adding this to the DELETE_ORIGINAL_FILE description: "If you enable this option, the extension won't be able to generate new resized images when you add image dimensions to your configuration."

or something like that

joehan: Done, thanks for the guidance!

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
+      logs.backfillComplete(successCount, errorsCount);
+      if (errorsCount == 0) {
+        await runtime.setProcessingState(
+          "PROCESSING_COMPLETE",
+          `Successfully resized ${successCount} images.`
+        );
+      } else if (errorsCount > 0 && successCount > 0) {
+        await runtime.setProcessingState(
+          "PROCESSING_WARNING",
+          `Successfully resized ${successCount} images, failed to resize ${errorsCount} images.`
```
ifielker, line 230: Are the failures in the logs? If so, we should probably mention that here. If not, we should probably put them in the logs and then mention that here :)

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
-            await remoteFile.delete();
-            logs.remoteFileDeleted(filePath);
-          } catch (err) {
-            logs.errorDeleting(err);
-          }
-        }
+      if (errorsCount > 0 && successCount == 0) {
+        await runtime.setProcessingState(
+          "PROCESSING_FAILED",
+          `Successfully resized ${successCount} images, failed to resize ${errorsCount} images.`
```
ifielker, line 236: Same here - mention logs

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
+        "PROCESSING_COMPLETE",
+        "Existing images were not resized."
+      );
       return;
     }
-
-    if (object.metadata && object.metadata.resizedImage === "true") {
-      logs.imageAlreadyResized();
-      return;
+    if (data == undefined) {
```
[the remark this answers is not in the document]
joehan: You're right - i missed that this wasn't showing up in the logs when testing.

storage-resize-images/functions/src/index.ts
```diff
@@ -49,149 +47,194 @@ logs.init();
...
+      } catch (err) {
+        logs.errorDeleting(err);
+      }
+      !verbose || logs.tempOriginalFileDeleted(filePath);
     }
-
-    if (!supportedContentTypes.includes(contentType)) {
-      logs.unsupportedType(supportedContentTypes, contentType);
-      return;
+    if (config.deleteOriginalFile === deleteImage.always) {
```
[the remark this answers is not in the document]
joehan: FWIW - this logic is copied over exactly from the existing code - I just moved the handler for the storage trigger into its own function so I could reuse it. 

Renamed these to be a bit clearer anyway - both are the original file, one is just the local copy and the other is the storage object.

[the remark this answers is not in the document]
ifielker: Thanks for making the code clearer :)

ifielker (approved):

PR #1310, opened 2022-11-23
fix: add optional chaining to forEach method calls, to prevent undefined errors

cabljac:
fixes: https://github.com/firebase/extensions/pull/1310

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/logs.ts
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/src/logs.ts
@@ -217,7 +217,7 @@ export const bigQueryTableInsertErrors = (
 ) => {
   logger.warn(`Error when inserting data to table.`);
 
-  insertErrors.forEach((error) => {
+  insertErrors?.forEach((error) => {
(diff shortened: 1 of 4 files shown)


firestore-bigquery-export/firestore-bigquery-change-tracker/src/logs.ts
```diff
@@ -217,12 +217,12 @@ export const bigQueryTableInsertErrors = (
...
   logger.warn(`Error when inserting data to table.`);
 
-  insertErrors.forEach((error) => {
+  insertErrors?.forEach((error) => {
     logger.warn("ROW DATA JSON:");
     logger.warn(error.row);
 
     if (error && error.errors) {
-      error.errors.forEach((error) =>
+      error.errors?.forEach((error) =>
```
dackers86, line 225: Is the `If` check needed now that we have optional chaining?

cabljac: you're right, this can be simplified

dackers86 (approved):

PR #1321, opened 2022-11-29
Solves the problem of forcing resized images to be placed in the root…

u310i:
… of the bucket when emulated on windows.

This is a case of an emulator on windows.
The modifiedFilePath variable is used as the upload destination for the resized images. In this case, bucket.upload method will not result in an error, but the image will be uploaded to the root of the bucket. Therefore, we use path.posix to convert the path to unix format.


dackers86:
Closing this as the posix fix has been added in version [storage-resize-images-v0.1.34](https://github.com/firebase/extensions/releases/tag/storage-resize-images-v0.1.34)

Issue #1389, opened 2023-01-05
storage-resize-images, backfill: Unhandled error FirebaseFunctionsError: Queue does not exist.


bhr:
### Describe your configuration

- Extension name: `storage-resize-images`
- Extension version: 0.1.32
- Configuration values (redact info where appropriate):
  - Cloud Functions location: europe-west1
  - Cloud Storage bucket for images: <project-id>.appspot.com
  - Sizes of resized images: 240x240,960x960,1920x1920
  - Deletion of original file: false
  - Make resized images public (Optional): false
  - Cloud Storage path for resized images (Optional): thumbs
  - Paths that contain images you want to resize (Optional): Parameter not set
  - List of absolute paths not included for resized images (Optional): <redacted>
  - Cache-Control header for resized images (Optional): Parameter not set
  - Convert image to preferred types: jpeg, webp, false
  - Output options for selected formats (Optional): Parameter not set
  - GIF and WEBP animated option (Optional):  true
  - Cloud Function memory: 1024
  - Backfill existing images: true

### Describe the problem

The cloud function throws the following error:
```
Unhandled error FirebaseFunctionsError: Queue does not exist.
    at FunctionsApiClient.toFirebaseError (/workspace/node_modules/firebase-admin/lib/functions/functions-api-client-internal.js:227:16)
    at /workspace/node_modules/firebase-admin/lib/functions/functions-api-client-internal.js:88:24
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async /workspace/lib/index.js:187:9
    at async /workspace/node_modules/firebase-functions/lib/common/providers/tasks.js:53:17 {
  errorInfo: {
    code: 'functions/failed-precondition',
    message: 'Queue does not exist.'
  },
  codePrefix: 'functions'
```

The issue might arise because the Cloud Tasks Queue is located in region `europe-west1`. The firebase extension might expect region `us-central1` though.

#### Steps to reproduce:

- Update to latest version of `storage-resize-images`
- Change configuration to enable new backfill feature

##### Expected result

- Backfill to work without errors

##### Actual result

- Backfill fails and cloud task fails too
<img width="1644" alt="Screenshot 2023-01-05 at 6 28 53 PM" src="https://user-images.githubusercontent.com/557660/210803561-5c2264c6-b6c5-432f-9bc1-9b2da1e96bb2.png">

TangrisJones:
[...]

PR #1395, opened 2023-01-10
fix(storage-resize-images): support backfill task in custom region

hipyhop:
fixes: #1389

Consider location (region) the extension is installed when loading task queue.

Previously regardless of which region the extension is installed, it will look for a the Task Queue in `us-central1`. 
See documentation for information: https://firebase.google.com/docs/reference/admin/node/firebase-admin.functions.functions.md#functionstaskqueue


dackers86:
Thanks @hipyhop.

This looks great, we'll just need to add an additional updates - I think the new location param in the config has affected one of the configuration tests.

dackers86 (approved):

PR #1412, opened 2023-01-17
[firestore-counter]: e2e testing

cabljac:
This PR adds an e2e test to `firestore-counter`, using the emulator.


firestore-counter/functions/package.json
```diff
@@ -8,26 +8,30 @@
...
     "firebase-functions": "^3.13.2",
-    "uuid": "^3.3.2",
     "rimraf": "^2.6.3",
-    "typescript": "^3.5.2",
-    "@types/chai": "^4.1.6",
-    "@types/mocha": "^5.2.5"
+    "typescript": "^3.5.2"
   },
   "devDependencies": {
+    "uuid": "^3.3.2",
```
cabljac, line 15: this needs to be moved back into dependencies, i've realized.

firestore-counter/functions/__tests__/util.ts
```diff
@@ -0,0 +1,137 @@
+/*
```
dackers86, line 1: As this is essentially taken from the node client. Was this recreated as we cannot reference the current client in the `clients` folder?

Maybe add a comment to reference the original source? Would this filename be better as `client.ts` or `test_client.ts` as it does not include any other utils?

dackers86 (approved):

PR #1416, opened 2023-01-23
[storage-resize-images] Always use posix paths when writing to storage

joehan:
Switch storage-resize-images to always use POSIX paths when writing to Storage.

This addresses an issue (https://github.com/firebase/firebase-tools/issues/5430) where, when emulating this extension on a Windows machine, resized images would be written to the root of the bucket with the Windows path as the filename.

Fixes https://github.com/firebase/firebase-tools/issues/5430.


simenness:
👀

simenness:
# 👋🏻😅
It's been a week.

dackers86 (approved):

Issue #1418, opened 2023-01-24
firestore-send-email: Warn message "message 'undefined' is not a valid object - please add as an object or firestore map, otherwise you may experience unexpected results." logged when using handlebar templates


sceee:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **firestore-send-email**
- Extension version: **0.1.20**
- Configuration values (redact info where appropriate):

### [REQUIRED] Step 3: Describe the problem

When using the trigger email extension with handlebar templates (and no explicit `message` field), the following warning message is logged each time an email is scheduled:
```
message 'undefined' is not a valid object - please add as an object or firestore map, otherwise you may experience unexpected results.
```

#### Steps to reproduce:

Schedule emails by adding an email as follows to the outbound queue:
```
await firestore()
  .collection("mail")
  .add({
    to: ["someone@example.com"],
    template: {
      name: "templatename",
      data: {
        some: "Data",
      },
    },
  });
```

Check the Logs of the `ext-firestore-send-email-processQueue` Cloud function. They emit the following warning message each time an email is sent:
```
message 'undefined' is not a valid object - please add as an object or firestore map, otherwise you may experience unexpected results.
```

This does not make sense as the message field is not required to be there when using handlebar templates.

##### Expected result

No warning is logged that the message field is undefined when using the trigger email extension with handlebar templates.

##### Actual result
See above.

dackers86:
HI @sceee 

Would it be possible to share the template `firestore document` you are using for this example?

sceee:
Hi @dackers86 ,

the used templates are all structured like this:
```
{
html: "<div>...</div>"
subject: "Some subject"
text: "Some text"
}
```

These are individual fields in the firestore document. Emails are also sent correctly, it's just the annoying warning in the logs.

cabljac:
just reproduced this. I'll need to look into when exactly we need the message field (e.g whether we can have templates but still need a message) and then implement a fix for the next release. Thanks!

gitdast:
The same issue ...log full of this. Luckily - emails were sent.

dmvvilela:
Same thing here. E-mails not being sent. Good thing is that I did not upgrade it on prod, only on dev project. Is there a way to downgrade a firebase extension??

sceee:
@dmvvilela in your case it then seems to be something different, probably you need to open another issue for your problem.
This particular issue was "just" about the false warning when using handlebar templates, emails are still being sent as written above.

[...]

PR #1422, opened 2023-01-30
fix(firestore-send-email): Move warning until after checking for templates

cabljac:
This PR moves the warning about missing messages until after the payload has been checked for template references. As mentioned [here ](https://github.com/firebase/extensions/issues/1418) this warning isn't necessary if a template is provided, as the message will be constructed from the template.

fixes: https://github.com/firebase/extensions/issues/1418

### DIFF
--- a/firestore-send-email/functions/src/index.ts
+++ b/firestore-send-email/functions/src/index.ts
@@ -103,6 +103,10 @@ function getExpireAt(startTime: admin.firestore.Timestamp) {
 async function preparePayload(payload: QueuePayload): Promise<QueuePayload> {
   const { template } = payload;
 
+  if (typeof payload.message !== "object" && !template) {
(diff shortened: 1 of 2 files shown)


firestore-send-email/functions/src/index.ts
```diff
@@ -351,17 +355,8 @@ async function processWrite(
...
-      // We expect the payload to contain a message object describing the email
-      // to be sent. If it doesn't, we can't do anything.
-      if (typeof payload.message !== "object") {
-        logs.invalidMessage(payload.message);
-        return false;
-      }
-
-      // The record has most likely just been created by a client, so we need to
-      // initialize the delivery state.
-
```
dackers86, line 363: Do we need to keep the comments here? As in they should follow the code and to remain for the code below?

dackers86 (approved):

Issue #1368, opened 2022-12-20
[firestore-bigquery-export] Better error logging on bigquery extension when partition is set to none


elvisun:
### Describe the problem
We have a firebase customer trying to set up the bigquery extension and is confused because they have partition option set to `none`, but still getting warnings: `Cannot partition an existing table ${table.dataset.id}_${table.id}`.

Looking at the code it might be because this line logs a warning regardless of the partitioning settings.
https://github.com/firebase/extensions/blob/fdbf5004c8d5e7b9749bbab3b2e16950ad812007/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/partitioning.ts#L131

The customer is asking if they can avoid these warning logs, so I wonder if we should have better logging/messaging here?

dackers86:
This is definitely something we should look into, this can be done through:

 - Improved messages
 - Verbose messaging

Or a combination of both, maybe.

adamkeys:
It would seem that there is a deeper issue here. Firstly, per the definition of `bigquery.Table`, exists()` returns a `Promise`, thus will always evaluate as being truthy here:

https://github.com/firebase/extensions/blob/41d9a3b5e46487d725c6bdd8c2f6ad1f0d438d7a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/partitioning.ts#L132-L137

In fact, to emphasize, the same method is called again in the following statement with `await`:

https://github.com/firebase/extensions/blob/41d9a3b5e46487d725c6bdd8c2f6ad1f0d438d7a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/partitioning.ts#L139

Additionally, the comment "No table provided, cannot evaluate" does not seem to correspond with the logic that follows. It appears to actually be in reference to the previous line.

Lastly, the content of the log message indicates that the table cannot be partitioned, but the scope of the method is only to check if the table has already been partitioned. These are conflicting concerns.

I suspect that said block of code was mistakenly included here and that the errant logging brought forward in this issue is a result of that, not due to the logging of it in general.

samatcolumn:
We're also experiencing a ton of logs related to partitioning even though we do not have partitioning configured.

PR #1461, opened 2023-02-13
fix(firestore-bigquery-export): Remove broken partition table check

adamkeys:
This commit removes a misplaced guard found in the isTablePartitioned method of the Partitioning class. The removed conditional was erroneously checking the truthfulness of a result of a method which returns a promise, thereby always evaluating to being true. Additionally, the removed logging operation is unrelated to the purpose of the method. This errant code likely ended up here as a result of a failed merge. This addresses an issue where firestore-bigquery-export needlessly fills the logs with errors about not being able to partition an existing table even when partitioning is disabled.

The log message called upon by the removed guard is also removed as it is no longer used. It may want to be reintroduced as part of the partitioning table creation phase in a future update, but that is outside of the scope of this update.

Additional discussion found in #1368

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/partitioning.ts
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/partitioning.ts
@@ -128,13 +128,9 @@ export class Partitioning {
   }
 
   private async isTablePartitioned() {
-    if (!this.table) return Promise.resolve(false);
     // No table provided, cannot evaluate
-    if (this.table.exists()) {
-      logs.cannotPartitionExistingTable(this.table);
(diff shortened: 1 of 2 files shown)


cabljac:
Hi there @adamkeys , thanks for your contribution! Could you sign the CLA, so I can run some CI checks?

adamkeys:
All set @cabljac

dackers86:
Hi @adamkeys 

Are we certain this not needed. Table is optional in the partition constructor.

Is this guard not necessary to ensure that we do not attempt to partition an existing table?

cabljac:
I think (if it is necessary and stays) it also requires an `await` in the check, as discussed in the linked issue

PR #1463, opened 2023-02-14
fix(firestore-send-email): added template check when no message exists

dackers86:
fixes: https://github.com/firebase/extensions/issues/1418


cabljac (approved):
lgtm!

JovannyCO (approved):
Resolve: "message 'undefined' is not a valid object

PR #1464, opened 2023-02-14
fix(firestore-bigquery-export): support path params

cabljac:
This PR fixes https://github.com/firebase/extensions/issues/1247 and adds support for wildcards when collection group queries are enabled.

Updated documentation to mention this.


cabljac:
Note that in the extension.yaml of the extension proper, we have a different name for the parameter (more appropriate imo) for wildcard exporting. This should be updated here too to match.

johnnybib:
I was trying this out and it seems to still throw an error stating "Collection Ids must not contain '/'"
Unless I was using the command incorrectly? Here's what I entered for my args (removed project id)

--non-interactive --project <project-id> -q true -s user_media_metadata/{userId}/media -d user_media_metadata -t user_media_metadata -l us -m true -u true -e false

The regex for validating the source-collection-path also needs to be updated in the interactive mode. It also rejects any path that contains a '/' character

cabljac:
thanks @johnnybib! will update the PR when i get a chance

nicolasdeory:
The problem is that the multithreaded implementation doesn't have the new path params yet. So running it single threaded is already working for me!

dackers86 (commented):
Looks good! Noticed the `config.test.ts` file is empty, does this need to be included?

Also needs validation updates.

cabljac:
closing this in favour of https://github.com/firebase/extensions/pull/1776 , which includes this update

PR #1465, opened 2023-02-14
chore(delete-user-data): remove firebase-tools dependency

cabljac:
This dependency is no longer needed, however pubsub seems to be missing.


delete-user-data/functions/package.json
```diff
@@ -17,19 +17,19 @@
   "author": "Lauren Long <longlauren@google.com>",
   "license": "Apache-2.0",
   "dependencies": {
+    "@google-cloud/pubsub": "^3.3.0",
```
dackers86, line 20: Is this dependency required, it doesn't appear to have existed previously?

`"@google-cloud/pubsub"`

dackers86: Interesting, ok sounds good!

PR #1473, opened 2023-02-21
fix(delete-user-data): improve search deletions

dackers86:


firestore-send-email/functions/src/index.ts
```diff
@@ -77,32 +76,28 @@ function validateFieldArray(field: string, array?: string[]) {
   }
 }
 
-function getExpireAt(startTime: admin.firestore.Timestamp) {
```
cabljac, line 80: is there a reason this stuff is being removed in this PR? Seems like some mix up of commits?

delete-user-data/functions/src/helpers.ts
```diff
@@ -25,3 +28,26 @@ export const getDatabaseUrl = (
...
   return `https://${selectedDatabaseInstance}.${selectedDatabaseLocation}.firebasedatabase.app`;
 };
+
+export const hasValidUserPath = async (
+  ref: DocumentReference,
+  path: string,
+  uid: string
+): Promise<boolean> => {
+  /** Check path for valid user id */
+  if (path.includes(uid)) return Promise.resolve(true);
```
cabljac, line 38: I think you can just `return true` here, the fact that it's an async function should pick up the promise

cabljac: similarly on line 46 and 52

delete-user-data/functions/src/helpers.ts
```diff
@@ -25,3 +28,26 @@ export const getDatabaseUrl = (
...
+
+  /** Check to find valid field */
+  const snapshot = await ref.get();
+
+  if (snapshot.exists) {
+    for (const field of config.searchFields.split(",")) {
+      if (snapshot.get(new FieldPath(field)) === uid) {
```
cabljac, line 45: Would a includes check here be better too?

something like
`
if (snapshot.get(new FieldPath(field)).includes(uid)) {
..
`

cabljac (changes requested):
Looks good, gave minor suggestions

cabljac (approved):
lgtm!

PR #1474, opened 2023-02-21
fix(firestore-bigquery-export): Added support for callable transform functions 

DeveloperEnvY:
Callable functions return the http response in the response.result field. To support setting the transform function to callable functions we can return the response.result.data field if it is defined, otherwise result the response.data field for http functions. Supporting callable transform functions can be useful for users that want to configure transform functions defined in other extensions, for example the Pangea Redact Text extension.


cabljac (approved):
lgtm!

Issue #1516, opened 2023-03-22
[firestore-send-email] change `missingDeliveryField` logging from `error` to `warn`


luke-rogers:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: `firestore-send-email`
- Extension version: 0.1.24
- Configuration values (redact info where appropriate):

#### Steps to reproduce:

What happened? How can we make the problem occur?

Add a document to the mail collection without the `delivery` field.

```
Error: message=mail/EZtCFfesAL9oVoyg7xMF is missing 'delivery' field
    at entryFromArgs (/workspace/node_modules/firebase-functions/lib/logger/index.js:130:19)
    at Object.error (/workspace/node_modules/firebase-functions/lib/logger/index.js:116:11)
    at Object.missingDeliveryField (/workspace/lib/logs.js:54:33)
    at /workspace/lib/index.js:299:18
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Transaction.runTransaction (/workspace/node_modules/@google-cloud/firestore/build/src/transaction.js:362:26)
    at async processWrite (/workspace/lib/index.js:281:35)
    at async /workspace/lib/index.js:384:9
```

##### Expected result
On missing `delivery` field in the mail document an `warn` level log occurs, as to avoid triggering any `error` level alerting users of the extension might have set up.

##### Actual result
On missing `delivery` field in the mail document an `error` level log occurs

ottob:
I started getting a lot of these errors after I updated the extension today. Why should this even be a warning? It seems to happen for all our emails with no actual consequence.

PR #1518, opened 2023-03-23
fix(firestore-send-email): downgrade error to warn

cabljac:
should fix https://github.com/firebase/extensions/issues/1516


cabljac:
closed as duplicate

PR #1520, opened 2023-03-27
Removing LOCATION param from all official extensions

joehan:
Removing the LOCATION param from all official extensions.

Now that system params is launched, this will be provided by the extensions backend and no longer needs to be in extension.yaml. The backend will also still provide process.env.LOCATION, so we don't need to make any code changes.

Docs for the new feature: https://firebase.google.com/docs/extensions/alpha/parameters#system_parameters


dackers86 (approved):

jschnurr:
How does the user supply the LOCATION using the CLI `firebase deploy` now? Previous to this change the location was picked up from the .env file in /extensions, but now the value is ignored and the deploy defaults to us-central1.

The [docs](https://firebase.google.com/docs/extensions/publishers/parameters#system_parameters) say `users have the opportunity to set custom values when they install your extension.` but they don't say **how**.

jschnurr:
Answering my own question, it was solved for another extension [here.](https://github.com/firebase/extensions/issues/1750#issuecomment-1760669335)

tl;dr - add
```bash
firebaseextensions.v1beta.function/location=asia-northeast1
```
to your .ENV (the variable name is case sensitive).

Issue #1533, opened 2023-04-06
[firestore-bigquery-export] TypeError: Cannot read properties of undefined (reading 'name')  (firebase.firestore.DocumentReference is undefined)


yutak23:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **firestore-bigquery-export**
- Extension version: 0.1.31
- Configuration values (redact info where appropriate):

#### Steps to reproduce:

My env is below.

- Node.js 16.19.1
- firebase-tools 11.25.1

I was using "firestore-bigquery-export" Extensions in my local development environment using an emulator([Firebase Local Emulator Suite](https://firebase.google.com/docs/emulator-suite)). In doing so, I got the error "Cannot read properties of undefined (reading 'name')"(my log is below).

From the error log, it seems that `firebase.firestore.DocumentReference` is undefined.

```console
i  functions: Beginning execution of "asia-northeast1-ext-firestore-bigquery-export-fsexportbigquery"
>  {"bqProjectId":"{projectId}","collectionPath":"users","datasetId":"firestore_export","tableId":"users","location":"asia-northeast1","initialized":false,"datasetLocation":"asia-northeast1","backupCollectionId":"users_bigquery_failed","timePartitioning":null,"clustering":null,"wildcardIds":false,"useNewSnapshotQuerySyntax":true,"severity":"INFO","message":"Started execution of extension with configuration"}
⚠  Google API requested!
   - URL: "https://oauth2.googleapis.com/token"
   - Be careful, this may be a production service.
⚠  Google API requested!
   - URL: "https://bigquery.googleapis.com/bigquery/v2/projects/{projectId}/datasets/firestore_export?"
   - Be careful, this may be a production service.
>  {"severity":"INFO","message":"BigQuery dataset already exists: firestore_export"}
⚠  Google API requested!
[...]

PR #1561, opened 2023-05-24
fix(firestore-bigquery-export): use module instead namespace

cabljac:
Reopening https://github.com/firebase/extensions/pull/1534 as CI wasn't running


yuta-katayama-23:
It seems that firebase-tools is no longer supporting Node.js 14.x due to v12(https://github.com/firebase/firebase-tools/releases/tag/v12.0.0).
For Node.js14, I think we need to change the `npm i -g firebase-tools` in CI to `npm i -g firebase-tools@11.30.0`, should I raise the PR for this fix?

https://github.com/firebase/extensions/blob/next/.github/workflows/test.yml#LL36C43-L36C43

dackers86 (approved):

yutak23:
relate to https://github.com/firebase/extensions/issues/1533

PR #1563, opened 2023-05-25
Specify location for taskQueues so that they dont default to us-central1

joehan:
It turns out that if you do not specify the location when getting a taskQueue, it defaults to us-central1 (https://github.com/firebase/firebase-admin-node/blob/master/src/functions/functions.ts#L50). This fixes an issue where these backfills would not work in other regions.


Rafer45 (approved):
LGTM

apascal07 (approved):

Issue #1565, opened 2023-05-31
Remove dev dependencies from dependencies list


cabljac:
Several extensions have things like `jest` and `@types/...` packages in their dependencies.

These should be moved to devDependencies.

In the past the reason for `@types/...` being in dependencies was because it needed to be for the build step of the extension. I'm not sure what the situation is now.

This should be checked across all 4 official repositories.

dackers86:
As the build processes uses docker, can we make sure this works as expected on install. Typically, in the past some dependencies will causes errors as the module cannot be found.

PR #1566, opened 2023-05-31
chore(delete-user-data): bump to node 16

cabljac:
- [ ] test installs fine

### DIFF
--- a/delete-user-data/extension.yaml
+++ b/delete-user-data/extension.yaml
@@ -62,7 +62,7 @@ resources:
       Cloud Storage.
     properties:
       location: ${param:LOCATION}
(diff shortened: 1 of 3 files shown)


cabljac:
> `handleSearch` and `handleDeletion` don't have a `runtime`, is the default nodejs16?

I had this thought as well. I think runtime *should* be at least 16 but I will double check before merging

pr-Mais:
guess it would be better if we add them explicitly

pr-Mais:
also let's move some dependencies to dev, per #1565

pr-Mais (approved):

dackers86 (approved):

PR #1614, opened 2023-07-01
Revert "Removing LOCATION param from all official extensions"

korturpazar:
Reverts firebase/extensions#1520


dackers86:
HI @korturpazar 

Can you explain the reason behind this PR? Thanks!

PR #1616, opened 2023-07-03
feat(firestore-bigquery-export): improved startup initialization and data synchronisation

dackers86:
- [x] Added cloud tasks for customisable concurrency when data syncing
- [x] Added lifecycle events for BQ initialisation

The lifecycle events ensure that the table updates now only occur when an extension is configured. This no longer happens on every update thanks to a new flag called `skipInit`. This only applies to the extension for this PR, and so does not affect the BQ scripts.


cabljac (approved):
lgtm!

Issue #1652, opened 2023-07-17
[firestore-bq-export] BREAKING CHANGE not mentioned. Change in Reference field type from string to object


dreinon:
- Extension name: firestore-bq-export
- Extension version: 0.1.33

#### Steps to reproduce:

1. Install mentioned extension at version 0.1.31 configuring a collection with Reference fields
2. See JSON ouput in bigquery (format: 'collectionName/id')
3. Update extension to version 0.1.33
4. See JSON output in bigquery (format has changed to an object instead of a string)

dackers86:
HI @dreinon 

I am currently unable to replicate. Are you referring to the `data` field in BQ?

Any screenshots would be helpful to narrow down which data has now become an object instead of a string.

Thanks!

dreinon:
Hi @dackers86
What changes is the Reference field representation inside the data JSON. Before, it was a string "collectionName/id", but now it's a full object with project information and an array of path segments

dreinon:
Before: "user/CYUqdFpKxDhyZAcxbZdzN4Pj7Bi2"
After: {"_firestore":{"projectId":"glissandoo-6eb90"},"_path":{"segments":["user","H3Z5IUghO3QDT2P4BSH6OVj9axZ2"]},"_converter":{}}

PR #1658, opened 2023-07-19
fix(firestore-bigquery-export): ensure data is correctly serialized before queuing

dackers86:
fixes: https://github.com/firebase/extensions/issues/1652


cabljac (approved):
lgtm

PR #1780, opened 2023-10-25
feat(gen-schema-view): added option to generate schema files to the local directory

### LINKED ISSUE #293: firestore-bigquery-export: typed arrays for schema views

This feature request is for extension: **`firestore-bigquery-export`**, and in particular for the [GENERATE_SCHEMA_VIEWS.md](https://github.com/firebase/extensions/blob/master/firestore-bigquery-export/guides/GENERATE_SCHEMA_VIEWS.md)

Support for arrays of different types. For example, some of the arrays in our project have maps inside them. A natural way to handle this would be to create a value field for each of the keys in the map. Even now, it's a bit strange that the array type doesn't specify if it's a string, number, boolean etc. 

### How would you use it?

To import and use bigquery on firestore objects that have arrays of nested maps.

dackers86:
fixes: https://github.com/firebase/extensions/issues/293

### DIFF
--- a/firestore-bigquery-export/scripts/gen-schema-view/src/createSchemaFiles.ts
+++ b/firestore-bigquery-export/scripts/gen-schema-view/src/createSchemaFiles.ts
@@ -1,0 +1,58 @@
+import * as fs from "fs";
+import {
+  FirestoreSchema,
+  changeLog,
+  latest,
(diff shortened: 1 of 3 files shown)


cabljac:
closing as stale

Issue #497, opened 2020-11-13
Extension: Trigger E-Mail, Feature: Flag rejectUnauthorized 


erostefano:
### Checkbox for flag rejectUnauthorized

This feature request is for extension:  (`Trigger E-Mail`)

### What feature would you like to see?

A checkbox to enable or disable the flag `rejectUnauthorized ` in the Trigger E-Mail environment.

### How would you use it?

The thirdparty smtps server uses a Shared Certificate which doesnt include my hostname in the altnames - and never will. Because this of this I get the following error message when sending mails:
`Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames`.

Setting the flag `{ rejectUnauthorized : false }` would solve this problem.

russellwheatley:
Hey @erostefano, thanks for the feature request. Here is the [option](https://nodemailer.com/smtp/#tls-options) in the nodemailer documentation that would be implemented. I'm sure the Firebase Extension's team will consider this feature and get back to you.

jhuleatt:
Hi @erostefano, thanks for the request! This sounds like it would be pretty easy to implement. However, we're also wary of increasing the number of parameters on an extension, since it makes the initial install experience more complex for developers new to the extension. We're going to hold off on implementing this until at least one other person requests similar functionality.

For anyone with a similar issue that's reading this: if you'd like this feature, please comment below!

aabc123:
Yes we're hitting this too.  Is there are work-around?

aabc123:
Ours may be due to our SMTP login having to be the full e-mail address of the account (including the @).  I don't think the extension likes the SMTP string having 2 @s in it.  I think it's using the e-mail address domain as the server, not the actual server we've entered.

Has anyone else got this issue?  Or could the parsing be improved to look for the last @ to find the server?

rhymer211:
Yes we're hitting this too.

L96Github:
Also needing that `rejectUnauthorized ` parameter to avoid getting
`Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames`

This is due to the SMTP server set-up but we are unable to change that set up at the moment...
Thanks

pumanitro:
I don't get it. It was working a week ago and now it stopped due to this error. So strange, but I've got it too and would like to solve it.

pumanitro:
[...]

PR #1788, opened 2023-11-09
feat(firestore-send-email): added tls configurable options

dackers86:
fixes: https://github.com/firebase/extensions/issues/497


cabljac (changes requested):
Can we add a default value for this param? otherwise people who leave it blank will have lots of warnings

cabljac (approved):

firestore-send-email/extension.yaml
```diff
@@ -155,6 +155,13 @@ params:
     validationErrorMessage: The value must be an integer value greater than zero.
     default: "1"
     required: true
+
+  - param: TLS_OPTIONS
+    label: Tls Options
```
Salakar, line 160:
```suggestion
    label: TLS Options
```

Issue #1522, opened 2023-03-31
[firestore-send-email] attachments from firebase storage not working with templates


elfasito:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

Extension name: firestore-send-email
Extension version: 0.1.24
Configuration values (redact info where appropriate):

### [REQUIRED] Step 3: Describe the problem
when I create a new document with attachment data on firestore pointing to a template. I get the error:
`Error: Invalid status code 403` when try to delivery.
this is with authentication rules for storage, if I allow access to all on storage, the email is sended but the attached image is broken.
anyways I using a public link url (w/ access token), so idk why have this problem.

PS: If I dont use a template the email is sended with attachments without problem.
PS2: if I use any image from web the email is sended correctly (with template)
PS3: if I put the URL from storage directly on the template (as a static value), the email is sended correctly with the attachment

#### Steps to reproduce:

What happened? How can we make the problem occur?
This could be a description, [log/console output](https://firebase.google.com/docs/extensions/manage-installed-extensions#view-logs), etc.

template content:
![Screenshot_4](https://user-images.githubusercontent.com/86434248/229251795-8096e4cd-2e33-43ad-9687-a09dc6852e96.png)

http://imgfz.com/i/5z2QI3i.png

firestore document content:
http://imgfz.com/i/BrxMShQ.png
![Screenshot_5](https://user-images.githubusercontent.com/86434248/229251755-6134b438-66b1-4e9b-ad07-c0b6e084a59a.png)

##### Expected result
email sended with attachments from the firebase storage

##### Actual result
email 403 error or corrupted attachments

RafaelZasas:
I'm experiencing bizzare issues too. I posted [This Question](https://stackoverflow.com/q/75944808/10673068) on stack overflow with more detail

fjtheknight:
[...]

PR #1820, opened 2023-11-22
fix(firestore-send-email): unescape path and filename

cabljac:
This fixes https://github.com/firebase/extensions/issues/1522 

However, I'm not sure this is the best solution. I *think* it's probably safe to disable escaping characters on the path and filename fields of the Attachments, but best to have a second opinion on this.

I experimented with a few different ways to modify the template from within handlebars.js and didn't have much success.

The alternative workaround is to make it clear in the docs that certain characters will be escaped by handlebars, and users can avoid this by using triple parentheses in their templates.


cabljac:
Closing as we decided this is an unnecessary change.

Issue #1877, opened 2023-12-04
🐛 [firestore-bigquery-export][Reopen] Partitioning is not enabled when I specify "timestamp" as time partitioning column name.


RyosukeNishi:
Reopened from https://github.com/firebase/extensions/issues/1819.
See [this](https://github.com/firebase/extensions/issues/1819#issuecomment-1837741106) for details.

cabljac:
Hey @RyosukeNishi thanks for your patience with this fix, what i can do for now is provide you with a branch with the fix that you can deploy from local. I've just triggered a release of `1.1.31` of the changetracker, I will link you a branch here, and then if you need any assistance deploying from local i'm happy to help. Thanks again!

cabljac:
Here's a branch with the latest dependency updated

https://github.com/firebase/extensions/tree/%40invertase/bump-changetracker-deps

cabljac:
I'm going to reopen this issue to track, so you know when the fix is live

cabljac:
Closing as the new release is out, which should fix this.

PR #1878, opened 2023-12-04
fix(firestore-bigquery-export): partition correctly if only timestamp is provided

RyosukeNishi:
fixes https://github.com/firebase/extensions/issues/1877


cabljac:
Hey @RyosukeNishi, thanks! I see what you're doing here, but i think currently the "omit" option is passed in by the extension, and is converted to `undefined` before it is given to the changetracker package, see here - 
https://github.com/firebase/extensions/blob/next/firestore-bigquery-export/functions/src/config.ts#L50
and - https://github.com/firebase/extensions/blob/next/firestore-bigquery-export/functions/src/index.ts#L42

The fix should be out in the next release of the extension, i'm just blocked by some branch protection changes right now.

Issue #911, opened 2022-03-11
[firestore-bigquery-export] Value for argument "limit" is not a valid integer


tgpsantos:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: firestore-bigquery-export@0.1.21 
- Extension version: 0.1.21 

### [REQUIRED] Step 3: Describe the problem

Hi guys,

Today I followed the import existing documents [guideline](https://github.com/firebase/extensions/blob/master/firestore-bigquery-export/guides/IMPORT_EXISTING_DOCUMENTS.md) but this problem occurred

![imagem](https://user-images.githubusercontent.com/48353485/157952558-0b1ac2d7-b62e-446d-9132-1e0f4bf11483.png)

Any idea on what's causing it?

PS: There is another issue, for some reason every time I run the script I have to introduce the values 2x (as it's also seen in the picture above)

dackers86:
Further investigation shows an issue when using the multi-thread option that will need to be fixed.

As a temporary solution, users should select no for this prompt.

nextlevelwebapps:
Hi, I'm getting the same result even when selecting No when prompted with the question of multiple threads.  I tried every configuration that I could think of, but it always resulted in this same error message.  I'm following the steps shown here https://github.com/firebase/extensions/blob/master/firestore-bigquery-export/guides/IMPORT_EXISTING_DOCUMENTS.md.  
![Screenshot (22)](https://user-images.githubusercontent.com/122231279/228896980-98adbc5d-441b-4e1e-8e40-fa51db8b80bf.jpg)

Roo2190:
still getting same error even set 'No' for the multiple threads :( any other solution?

diepminh707:
> Hi, I'm getting the same result even when selecting No when prompted with the question of multiple threads. I tried every configuration that I could think of, but it always resulted in this same error message. I'm following the steps shown here https://github.com/firebase/extensions/blob/master/firestore-bigquery-export/guides/IMPORT_EXISTING_DOCUMENTS.md. ![Screenshot (22)](https://user-images.githubusercontent.com/122231279/228896980-98adbc5d-441b-4e1e-8e40-fa51db8b80bf.jpg)



> still getting same error even set 'No' for the multiple threads :( any other solution?

You can use this
`npx @firebaseextensions/fs-bq-import-collection \
  --non-interactive \
  --project=[your_project] \
  --source-collection-path=[your_collection] \
  --query-collection-group=false \
[...]

PR #1930, opened 2024-01-22
fix(firestore-bigquery-export): parse batchSize correctly

### LINKED ISSUE #1470: [firestore-bigquery-export] Error importing Collection to BigQuery: Error: Value for argument "limit" is not a valid integer.
When I introduce the following values, I get an error message:

? What is your Firebase project ID? ****
? What is the path of the the Cloud Firestore Collection you would like to import from? (This may, or may not, be the same Collection for which you plan to mirror changes.) dim_stock
? Would you like to import documents via a Collection Group query? No
? What is the ID of the BigQuery dataset that you would like to use? (A dataset will be created if it doesn't already exist) firestore_dim_stock
? What is the identifying prefix of the BigQuery table that you would like to import to? (A table will be created if one doesn't already exist) dim_stock
? How many documents should the import stream into BigQuery at once? 300
? Where would you like the BigQuery dataset to be located? southamerica-east1
? Would you like to run the import across multiple threads? No
? Would you like to use the new optimized snapshot query script? No
? Would you like to use a local firestore emulator? No

Importing data from Cloud Firestore Collection: dim_stock, to BigQuery Dataset: firestore_dim_stock, Table: dim_stock_raw_changelog
Error importing Collection to BigQuery: Error: Value for argument "limit" is not a valid integer.

Any idea what is causing this?

jauntybrain:
Parse parameter `batchSize` correctly in the interactive `import` script flow.

Fixes #1470 and #911

### DIFF
--- a/firestore-bigquery-export/scripts/import/src/config.ts
+++ b/firestore-bigquery-export/scripts/import/src/config.ts
@@ -250,7 +250,7 @@ export async function parseConfig(): Promise<CliConfig | CliConfigError> {
     sourceCollectionPath: sourceCollectionPath,
     datasetId: dataset,
     tableId: table,
-    batchSize: batchSize,
+    batchSize: parseInt(batchSize),
     queryCollectionGroup: queryCollectionGroup,
     datasetLocation: datasetLocation,
     multiThreaded: multiThreaded,


cabljac (approved):
lgtm!

pr-Mais (approved):

PR #1935, opened 2024-01-29
feat(storage-resize-images): add parameter REGENERATE_TOKEN

### LINKED ISSUE #1925: 🐛 [storage-resize-images] configuration option for access token

Issues filed here should be about a feature request for a **specific extension in this repository**. To file a feature request that affects multiple extensions or the Firebase Extensions platform, please reach out to
[Firebase support](https://firebase.google.com/support/troubleshooter/report/features/) directly.

This feature request is for extension: **\_** (`storage-resize-images`, `firestore-send-email`, etc)

I propose adding a new configuration option to the storage-resize-images extension, specifically a boolean field named regenerateToken. This field will influence the behavior of the extension when generating resized images. When set to true, the extension will assign a new access token to the resized image, similar to the current default behavior. When set to false, the resized image will retain the same access token as the original image.

Currently, the only way to achieve this behavior is to manually modify the extension's code, specifically by commenting out lines in the generateResizedImage function. This process is cumbersome and not feasible for testing in local emulator suites or maintaining consistency across updates in production environments.

### How would you use it?

Implementing this feature would greatly enhance the flexibility and usability of the storage-resize-images extension, especially in scenarios where consistent access tokens between original and resized images are necessary. For example, in our application, we use the same access token for both original and resized images to simplify client-side image access and caching mechanisms.

Having the ability to configure this behavior through the extension's settings would streamline our development and deployment processes, allowing us to effectively test in local environments and ensure smooth updates in production without manually altering the extension's code. This addition would make the extension more adaptable to a variety of use cases and improve its overall utility for developers.

jauntybrain:
Allow users to decide whether the resized image will have the same access token as the old one, or a newly generated one.

Fixes #1925


storage-resize-images/extension.yaml
```diff
@@ -320,6 +320,20 @@ params:
       - label: No
         value: false
 
+  - param: REGENERATE_TOKEN
+    label: Assign new access token
+    description: >
+      Should resized images have a new access token assigned to them,  different
+      from the original image?
+    type: select
+    required: true
```
pr-Mais, line 329: i don't think this should be required, it might cause a breaking change for existing users, change it to none-required.

pr-Mais (changes requested):

pr-Mais (approved):
LGTM

almogtovim:
Can I use it already?

PR #1944, opened 2024-02-02
feat(storage-resize-images): preserve original image metadata by default

### LINKED ISSUE #1605: Resize Images: preserve EXIF tags and filename
This feature request is for extension: **firebase/storage-resize-images**

I would like to have slightly more control over the thumbnail file generation through the extension's settings.
- optionally preserve EXIF tags from source in the output image (especially copyright)
- optionally disable the resolution suffix in filename (original_**200x200**.jpg) if only one thumbnail size is requested
- 

### How would you use it?

When I have my extension configured to output images in the /thumbnails subfolder, I would like to tick two extra boxes to keep exif tags and filenames intact when generating the output into /thumbnails.

jauntybrain:
Adds a `keepMetadata()` to sharp calls to preserve original file metadata. 

Upgrades `sharp` to version 0.33.2

Fixes #1605.


pr-Mais:
@jauntybrain can you check the failing tests

PR #1993, opened 2024-03-20
feat(firestore-bigquery-export): add retry config

jauntybrain:


pr-Mais (changes requested):
can we remove the config from the functions as well as it's not doing anything

pr-Mais (approved):

PR #2068, opened 2024-04-19
ci(*): more efficient testing

cabljac:
Currently our CI builds ALL the extensions and runs ALL the tests, even on pull requests.

This proposed workflow filters on changes in the PR, and only runs tests on extensions which have changes.

There are still some small optimisations i'd like to make eventually, but:

1. We can still run all the tests on pushes to next. In PRs to next we can just test the PR's changes.
2. Currently test logs are super noisy, this slows us down as it's actually hard in the github action logs to work out which tests are failing.
3. This also runs the tests in parallel, so it'd (potentiall) be quicker even if you made changes in every extension.

### DIFF
--- a/.github/workflows/conditional-test.yml
+++ b/.github/workflows/conditional-test.yml
@@ -12,51 +12,78 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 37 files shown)


cabljac:
stale

Issue #2031, opened 2024-04-08
🐛 Critical: [delete-user-data] Recursive Deletion Fails to Remove Subcollections When Using Auto Discovery


cgilbu:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **delete-user-data**
- Extension version: **0.1.20**
- Configuration values:
  - **Cloud Firestore paths:** `accounts/{UID}`
  - **Cloud Firestore delete mode:** `recursive`
  - **Realtime Database instance:** (Not set)
  - **Realtime Database location:** `us-central1`
  - **Realtime Database paths:** (Not set)
  - **Cloud Storage bucket:** `myapp.appspot.com`
  - **Cloud Storage paths:** (Not set)
  - **Enable auto discovery:** `yes`
  - **Auto discovery search depth:** `1`
  - **Auto discovery search fields:** `accountId`
  - **Search function URL:** (Not set)

#### Steps to reproduce:

I have set the delete mode to "recursive" with the expectation that deleting a user would remove the user's document at the specified path, along with all its subcollections. However, while the document fields are deleted, its subcollections are not removed and remain intact in the database.

To reproduce this issue:
1. Configure the extension with recursive delete mode and a Firestore path.
2. Delete a user whose data is stored at the specified path.
3. Check Firestore and observe that the main document is cleared but subcollections still exist.

This issue occurs not only with documents located using auto discovery but also with those specified directly in the Firestore paths. In both cases, the subcollections remain undeleted, despite the document fields being cleared.

I have checked the API logs and found no errors related to this issue. The deletion process appears to complete without error logs, but the expected behavior of recursive deletion is not achieved.

##### Expected result

When the extension is triggered, the specified user's document and all its subcollections should be completely deleted from Firestore.

##### Actual result

The document fields are deleted, but subcollections under the document are not deleted, remaining visible and intact in Firestore.

[...]

PR #2073, opened 2024-04-23
fix(delete-user-data): recursive delete searched docs if enabled

cabljac:
fixes https://github.com/firebase/extensions/issues/2031


siroracoin (commented):
Hii

dackers86 (approved):

PR #2115, opened 2024-06-13
fix(storage-resize-images): maintain aspect ratio of resized images

### LINKED ISSUE #2106: 🐛 [storage-resize-images] Extension crops the image with default settings.

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

- Extension name: **storage-resize-images** (`storage-resize-images`, `firestore-send-email`, etc)
- Extension version: **0.2.5**
  - **\_**
  - **\_**

In the documentation it says, it will keep the aspect ratio when resizing images.
But currently it doesn't. It crops the image

#### Steps to reproduce:

1- Install the extension
2- Add 300x300, 756x756, 900x900 image resolutions
3- What the installation finished
4- Upload an image
5- Check the results

##### Expected result

Images should have been resized with the same aspect ratio and crop should have happen

##### Actual result

Images lost their aspect ratio and and it is cropped

gregives:
The extension documentation says

> This extension keeps the aspect ratio of uploaded images constant and shrinks the image until the resized image’s dimensions are at or under your specified max width and height.

However, it doesn't currently do that: this pull request fixes it.

Fixes #2106

### DIFF
--- a/storage-resize-images/functions/src/resize-image.ts
+++ b/storage-resize-images/functions/src/resize-image.ts
@@ -47,6 +47,7 @@ export function resize(file, size) {
   return sharp(file, ops)
     .rotate()
     .resize(parseInt(width, 10), parseInt(height, 10), {
+      fit: "inside",
       withoutEnlargement: true,
       ...sharpOptions,
     })


cabljac (approved):

cabljac:
Thanks for your contribution, and your patience in getting this fixed! I will put out a new release with the fix soon.

ralphferguson:
@cabljac & @gregives — do you think that this value could be configurable? I have essentially no experience with coding solutions for plugins like this (and the associated input field), but I'd be happy to look into it with some guidance, as I believe this would be a useful parameter to configure. What do you both think?

gregives:
Hi @ralphferguson, this option is actually configurable already, it is called "Sharp constructor options for resizing images" and it will override these default options

Issue #2090, opened 2024-05-10
🐛 [firestore-bigquery-export] multi-db selection isn't working for Firestore triggers


pr-Mais:
### [REQUIRED] Step 3: Describe the problem

The current version of the Stream Firestore to BigQuery extension allows specifying a secondary project for the Firestore triggers, however this doesn't work as the triggers are all 1st gen. Need to remove the feature.

See #2072

titan-graham:
Found this issue via (#2072):

We are somewhat confused/bemused that the current version of the extension allows a non `(default)` database ID, but a non default database ID is not supported by the underlying technology (gen1 functions) that the extension relies on. 

`Error creating trigger: The request was invalid: generic::invalid_argument: source filter must follow '//firestore.googleapis.com/projects/{projectId}/databases/(default)'.`

We've wasted pointless hours on this where in fact the extension as published simply cannot work in the way it's advertised. How is it possible this feature even made it into a release?

PR #2122, opened 2024-06-18
fix(firestore-bigquery-export): remove multiple database support

cabljac:
should fix https://github.com/firebase/extensions/issues/2090

### DIFF
--- a/firestore-bigquery-export/CHANGELOG.md
+++ b/firestore-bigquery-export/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 fixed - medium npm vulnerabilities
 
+fixed - rollback broken multiple database parameter option
+
 ## Version 0.1.50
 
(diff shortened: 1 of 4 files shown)


jggomezt:
When will you be supporting multi-db??

marcuscorreasmash:
i need too supporting multi-db :(

IM-CrazyDeveloper:
any chance to get fix or workaround to support multi DB?

rummanext:
1+ here

beyondprompting:
still waiting for this, is so basic to use the extension

calebripley:
Would love to use this extension as well but can't due to my data not being in default. Would also recommend updating the docs to be explicit that default db will be used.

cabljac:
I will escalate this to the team, we're blocked from a platform side at the moment, as v1 functions dont support non-default db, and extensions don't support v2 for firestore triggers.

PR #2141, opened 2024-08-08
chore: use pnpm in place of lerna

cabljac:
This PR replaces the outdated and deprecated lerna methods with pnpm.

It also fixes a bunch of npm vulnerabilities, and updates some versions.

TODO:
- Make sure use of pnpm doesn't break release process
- Regression testing before we release things (to check version bumps of packages haven't introduced errors


docs/firestore-bundle-builder/api-reference.md
```diff
@@ -90,7 +90,7 @@ type QueryCondition = {
       | "not-in"
```
cabljac, line 90: im not sure why we have docs for this extension here? We should remove irrelevant docs probably? @pr-Mais

cabljac:
CI is failing as it relied on lerna. need to switch to pnpm scripts.

cabljac:
CI for testing also needs to use npm probably or something? pnpm doesn't create node_modules where we need it for extensions

Salakar:
woops

Issue #2087, opened 2024-05-03
🐛 [firestore-bigquery-export] schema-views script doesn't work against multiple schema files


KevinKera:
### Step 2: Describe your configuration

- Extension name: **firestore-bigquery-export (schema-views script)** 
- Extension version: **0.1.49**
- Configuration values (redact info where appropriate):
  - **Cloud Functions location** : us-central1
  - **BigQuery Dataset location** : us
  - **BigQuery Project ID**: xxx
  - **Database ID:** (default)
  - **Collection path:** xxx
  - **Enable Wildcard Column field with Parent Firestore Document IDs(Optional):** false
  - **Dataset ID:** xxx
  - **Table ID**: xxx
  - **BigQuery SQL table Time Partitioning option type(Optional):** NONE
  - **BigQuery Time Partitioning column name(Optional):** Parameter not set
  - **Firestore Document field name for BigQuery SQL Time Partitioning field option(Optional):** Parameter not set
  - **BigQuery SQL Time Partitioning table schema field(column) type(Optional):** omit
  - **BigQuery SQL table clustering(Optional):** Parameter not set
  - **Maximum number of synced documents per second(Optional):** 100
  - **Backup Collection Name(Optional):** failed_bq_updates
  - **Transform function URL(Optional):** Parameter not set
  - **Use new query syntax for snapshots:** yes
  - **Exclude old data payloads(Optional):** yes
  - **Use Collection Group query(Optional):** no
  - **Cloud KMS key name(Optional):** Parameter not set
  - **Enabled events:** firebase.extensions.big-query-export.v1.sync.start
  - **fs-bq-schema-views command:**
```
npx @firebaseextensions/fs-bq-schema-views \
  --non-interactive \
  --project=xxx \
  --dataset=xxx \
  --table-name-prefix=xxx \
  --schema-files=./part_schema_all.json,./part_schema_consented.json,./part_schema_opposed.json
``` 

#### Steps to reproduce:

Setup the extension as mentioned in the documentation. Follow the instructions found in the [GENERATE_SCHEMA_VIEWS](https://github.com/firebase/extensions/blob/master/firestore-bigquery-export/guides/GENERATE_SCHEMA_VIEWS.md#step-3-run-the-script) guide until step 3.   

There, the note says 

[...]

PR #2223, opened 2024-11-29
[Bug] Fix issue #2087: Handle multiple schema files in gen-schema-vie…

Gustolandia:
Description:

Issue: The gen-schema-views script failed to process multiple schema files passed via the --schema-files flag. When multiple schema files were provided as a comma-separated string, the script treated them as a single file path, resulting in the error: No schema files found!

Cause: The --schema-files flag input was not being split into individual file paths in non-interactive mode, and the expandGlobs function did not handle this case.

Fix:

Updated the expandGlobs function in schema-loader-utils.ts to:
Split comma-separated schema file paths into individual paths.
Trim and process each path individually.
Logically expand glob patterns for each path.
Added functionality in parseConfig to ensure that schema file paths are split correctly before being passed to readSchemas.
Tested the fix with:
A single schema file.
Multiple schema files passed as a comma-separated list.
Cleaned up debugging console.log statements after confirming the fix.
Testing:

Successfully tested the fix locally using:
npx ts-node ./src/index.ts \
  --non-interactive \
  --project=extensions-testing \
  --big-query-project=extensions-testing \
  --dataset=firestore_export \
  --table-name-prefix=posts \
  --schema-files=./schemas/test.json,./schemas/user_full.json
Verified that multiple schema files were processed correctly, and BigQuery views were created for each schema.
Notes:

This PR addresses issue #2087.
The fix ensures backward compatibility for single schema files and supports both interactive and non-interactive modes.

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/package-lock.json
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@firebaseextensions/firestore-bigquery-change-tracker",
-  "version": "1.1.37",
+  "version": "1.1.38",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@firebaseextensions/firestore-bigquery-change-tracker",
-      "version": "1.1.37",
+      "version": "1.1.38",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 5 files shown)


cabljac (commented):
Can we add tests to cover this case please? Otherwise lgtm!

Gustolandia:
I added a test.

sumikn233-al (approved):
Qscd[]()[]()**** @ ****

Issue #2132, opened 2024-07-16
🐛 [firestore-bigquery-export] Import Script Throws 'Request Entity Too Large'


larstbone:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: firestore-bigquery-export
- Extension version: 0.1.51
- Configuration values (redact info where appropriate):

> BigQuery Dataset location->us
> BigQuery Project ID->xxxxxxxxxxxxxxxxx
> Collection path->xxxxxxxxxxxxxxxxxxx
> Enable Wildcard Column field with Parent Firestore Document IDs (Optional)->false
> Dataset ID->firestore_events
> Table ID->events
> BigQuery SQL table Time Partitioning option type (Optional)->NONE
> BigQuery Time Partitioning column name (Optional)->Parameter not set
> Firestore Document field name for BigQuery SQL Time Partitioning field option (Optional)->Parameter not set
> BigQuery SQL Time Partitioning table schema field(column) type (Optional)->omit
> BigQuery SQL table clustering (Optional)->Parameter not set
> Maximum number of synced documents per second (Optional)->100
> Backup Collection Name (Optional)->Parameter not set
> Transform function URL (Optional)->Parameter not set
> Use new query syntax for snapshots->yes
> Exclude old data payloads (Optional)->yes
> Use Collection Group query (Optional)->no
> Cloud KMS key name (Optional)->Parameter not set
> 

### [REQUIRED] Step 3: Describe the problem
We cannot import existing data that may be up to 900KB.

#### Steps to reproduce:

1. Using the extension configured on some collection
2. Reconfigure the extension by setting EXCLUDE_OLD_DATA set to `true`
3. When reconfiguring, confirm DO_BACKFILL is no longer available
4. Create a document that is 900KB
5. Confirm the document does not sync and fails with error `task size too large`
6. Run the script `fs-bq-import-collection` to try to import existing data
7. Confirm error `Request Entity Too Large`

During installation of the extension cannot backfill existing data because DO_BACKFILL is currently disabled by #2005.  

##### Expected result

1. DO_BACKFILL is available during both installation and reconfiguration of the extension
2. script `fs-bq-import-collection` does not throw `Request Entity Too Large` error if we are ignoring old data by setting EXCLUDE_OLD_DATA flag to true

##### Actual result

1. DO_BACKFILL is not available when the extension is installed nore when it is reconfigured
2. script `fs-bq-import-collection` fails with `Request ENtity too Large`

[...]

Issue #2055, opened 2024-04-18
🐛 [fs-bq-import-collection] Documents Aren't Passing Between Projects


joeframpton:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **fs-bq-import-collection** 
- Extension version: **0.1.21** 
- Configuration values (redact info where appropriate): The CLI command was (either of the following): 
  -  _npx @firebaseextensions/fs-bq-import-collection --non-interactive --project=projectA* --source-collection-path=event-data --query-collection-group=true --use-new-snapshot-query-syntax=true --big-query-project=projectB* --dataset=event_data --table-name-prefix=ed --dataset-location=eu --multithreaded_
  - _npx @firebaseextensions/fs-bq-import-collection --non-interactive -P projectA* -B projectB* -q true -s event-data -d event_data -t ed -l eu -m true -u true_
*redacted specific project names

#### Steps to reproduce: 
 - Set up projectA and populate firestore collection group event-data
 - Set up projectB but do not populate firestore
 - Run the command above

#### What happened? How can we make the problem occur?
The program failed with an error - I didn't screenshot the error but it was related to line 26 in worker.js, specifically that `docs[0].id` was undefined. 
I added some additional logging to worker.js and found that while the CLI config was being parsed correctly, the instance of firebase that was initialised was that of projectB **not** projectA. The error that occurred was that the collection-group-query was searching projectB's firestore, and was not finding any documents so docs[0].id was indeed null/undefined. 

I tried running the command while in a directory that didn't have an instance of firebase.json and found the same issue. 

You can reproduce these steps by logging the config that is passed into the function in line 7 of worker.js and logging the query object before line 25 to see that the query is happening to the wrong project. 

##### Expected result
The expected result would be that a table would be created in bigquery in projectB, and is filled with the data from the collection group 'event-data' from firestore in projectA

[...]

Issue #1852, opened 2023-11-29
🐛 [firestore-bigquery-export] The import script fails when run on multiple threads with millions of records.


RyosukeNishi:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: **firestore-bigquery-export**
- Extension version:
  - The extension itself
    - **0.1.41**
  - The import script
    - **0.1.19**
- Configuration values (redact info where appropriate):
<img width="646" alt="スクリーンショット 2023-11-29 12 02 15" src="https://github.com/firebase/extensions/assets/46443906/12967046-4450-4834-b412-e6cc412db413">
<img width="646" alt="スクリーンショット 2023-11-29 12 02 23" src="https://github.com/firebase/extensions/assets/46443906/7e412d91-9e01-4c43-a666-b28764a56f37">

#### Steps to reproduce:
```
fs-bq-import-collection \
  --non-interactive \
  --project ${MY_PROJECT_NAME} \
  --source-collection-path chats/{chat_token}/messages/{message_unique_id}/reaction_groups \
  --query-collection-group true \
  --dataset firestore_export \
  --table-name-prefix reaction_groups \
  --batch-size 300 \
  --dataset-location us \
  --multi-threaded true \
  --use-new-snapshot-query-syntax true \
  --use-emulator false
```
What happened? How can we make the problem occur?
This could be a description, [log/console output](https://firebase.google.com/docs/extensions/manage-installed-extensions#view-logs), etc.

##### Expected result
The script succeeds.

##### Actual result
```
[...]

PR #2264, opened 2025-01-30
feat(firestore-bigquery-export): import script updates

cabljac:
provides fixes for:

fixes https://github.com/firebase/extensions/issues/2132

fixes https://github.com/firebase/extensions/issues/2055

fixes https://github.com/firebase/extensions/issues/1852

This PR:

- fixes path_params in multi-threaded, they no longer import as `null` in BQ

- sets the bigquery project ID in the changetracker instance (instead of an environment variable, which was pointing the firebase client at the wrong project)

- ensures that the multi-thread feature does not exceed the batch size option when writing to bigquery (mostly done in worker.ts)

- on a failed batch, instead of the whole script failing, the script will now just log the paths of the failed docs, and there is an option to output a `.txt` file with newline separated document paths of failed imports

### DIFF
--- a/firestore-bigquery-export/scripts/import/src/index.ts
+++ b/firestore-bigquery-export/scripts/import/src/index.ts
@@ -55,9 +55,9 @@ const run = async (): Promise<number> => {
     process.env.FIRESTORE_EMULATOR_HOST = "127.0.0.1:8080";
   }
   // Set project ID, so it can be used in BigQuery initialization
-  process.env.PROJECT_ID = bigQueryProjectId;
-  process.env.GOOGLE_CLOUD_PROJECT = bigQueryProjectId;
-  process.env.GCLOUD_PROJECT = bigQueryProjectId;
+  process.env.PROJECT_ID = projectId;
+  process.env.GOOGLE_CLOUD_PROJECT = projectId;
+  process.env.GCLOUD_PROJECT = projectId;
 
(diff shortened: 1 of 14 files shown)


firestore-bigquery-export/scripts/import/src/index.ts
```diff
@@ -80,6 +80,7 @@ const run = async (): Promise<number> => {
     datasetLocation,
     wildcardIds: queryCollectionGroup,
     useNewSnapshotQuerySyntax,
+    bqProjectId: bigQueryProjectId,
```
cabljac, line 83: this is to fix https://github.com/firebase/extensions/issues/2055

firestore-bigquery-export/scripts/import/src/worker.ts
```diff
@@ -19,57 +20,102 @@ async function processDocuments(
...
+
+    console.log(
+      `Processing batch of ${docs.length} docs: ${docs[0].ref.path} -> ${
+        docs[docs.length - 1].ref.path
+      }`
+    );
+
+    const rows: FirestoreDocumentChangeEvent[] = getRowsFromDocs(docs, config);
+
+    try {
```
cabljac, line 83: we catch import errors now, instead of the whole script failing

dackers86 (approved):

PR #2330, opened 2025-03-12
feat(firestore-biqquery-export): add log level config param

CorieW:
draft as still needs tested


firestore-bigquery-export/firestore-bigquery-change-tracker/src/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.runIfLogLevel(levels.error, funcsLogger.error, ...args);
+  }
+
+  log(...args: any[]): void {
+    this.info(...args);
+  }
+
+  private runIfLogLevel(level: number, func: Function, ...args: any[]): void {
+    if (this.logLevel <= level) {
+      func(...args);
```
cabljac, line 73:
```suggestion
     func.bind(funcsLogger)(...args);
```

might be safer, in case `func` uses its own `this` ?

firestore-bigquery-export/extension.yaml
```diff
@@ -461,6 +461,27 @@ params:
...
+        value: debug
+      - label: Info
+        value: info
+      - label: Warn
+        value: warn
+      - label: Error
+        value: error
+      - label: Silent
+        value: silent
+    default: info
```
cabljac, line 482: not sure what the default should be, as it seems too verbose at the moment. info seems the right level to be default, maybe we need to downgrade some of the logs to debug (im thinking the logs in the package where they log "inserting" and "inserted"

firestore-bigquery-export/firestore-bigquery-change-tracker/src/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  error(...args: any[]): void {
+    this.runIfLogLevel(levels.error, funcsLogger.error, ...args);
+  }
+
+  log(...args: any[]): void {
+    this.info(...args);
+  }
+
+  private runIfLogLevel(level: number, func: Function, ...args: any[]): void {
```
cabljac, line 71:
```suggestion
  private runIfLogLevel<T extends any[]>(level: number, func: (...args: T) =>
```

Maybe something like this would be better typed?

cabljac: not a fan of all the any[]

cabljac: oh this is how the firebase logger is typed maybe, right

firestore-bigquery-export/firestore-bigquery-change-tracker/src/logger.ts
cabljac: ignore this, just looked at firebase logger code

firestore-bigquery-export/firestore-bigquery-change-tracker/src/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+export class Logger {
+  private logLevel: number;
+
+  constructor(logLevel: LogLevel | string | number = LogLevel.INFO) {
+    this.setLogLevel(logLevel);
+  }
+
+  setLogLevel(logLevel: LogLevel | string | number): void {
+    if (typeof logLevel === "string") {
+      this.logLevel = levels[logLevel];
```
cabljac, line 43:
```suggestion
      this.logLevel = levels[logLevel as keyof typeof levels] ?? levels.info;
```

or `logLevel as string` maybe

the main thing is the fallback `?? levels.info`  or something equivalent, just in case a bad string comes through and we'd have `undefined`

firestore-bigquery-export/functions/__tests__/config.test.ts
```diff
@@ -6,6 +6,7 @@ import * as yaml from "js-yaml";
 import mockedEnv from "mocked-env";
 
 import { clustering } from "../src/config";
+import { log } from "console";
```
cabljac, line 9: is this for some mocking purpose or something? why do we now import it

firestore-bigquery-export/functions/src/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+export class Logger {
+  private logLevel: number;
+
+  constructor(logLevel: LogLevel | string | number = LogLevel.INFO) {
+    this.setLogLevel(logLevel);
+  }
+
+  setLogLevel(logLevel: LogLevel | string | number): void {
+    if (typeof logLevel === "string") {
+      this.logLevel = levels[logLevel];
```
cabljac, line 43: similar here right, need `??` in case a bad string comes through

firestore-bigquery-export/functions/src/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+    } else if (typeof logLevel === "number") {
+      this.logLevel = logLevel;
```
cabljac, line 45: nitpicky but technically should we check it's in the right range 0 <= logLevel <= 4

and do something else otherwise

firestore-bigquery-export/functions/src/logger.ts
```diff
@@ -0,0 +1,78 @@
+/*
```
cabljac, line 1: is this code identical to the logger in the npm package?

We could maybe import it instead of repeating it, idk if that's best though

firestore-bigquery-export/functions/__tests__/config.test.ts
CorieW: Whoops accidental import. Good spot

firestore-bigquery-export/functions/src/logger.ts
CorieW: Hmm, I don't like that. Think instead, I'm going to remove `number` as a possible type for the logLevel param.

firestore-bigquery-export/functions/src/logger.ts
CorieW: Wouldn't it be pretty much the same amount of code?

cabljac:
also run `npm run format` from root (you may have to do `npm i` from the root first, to ensure the prettier rules used etc match the repo)

firestore-bigquery-export/functions/src/logger.ts
CorieW: The extension functions doesn't install from local source, so I don't think it'd work.

CorieW: It's annoying having the duplicate stuff. I wonder if it's possible to setup a common directory or something 🤔

PR #2364, opened 2025-04-07
feat(firestore-bigquery-changetracker): support non-default db

cabljac:
This PR moves to v2 on the change tracker package, so that the extension can also move to v2.

During this PR I had to add a bunch of jest.spyOn as the current tests are intractable when one fails.

I also had to regen package-locks etc as locally the change tracker had been `npm link`'d to other packages and the extension.
 
It seems Lerna doesn't respect package.json and tries to link the change tracker anyway.


CorieW (approved):
lgtm

Issue #2372, opened 2025-04-17
🐛 [firestore-send-email] undefined (reading 'attachments') error using SendGrid dynamic templates


CorieW:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **firestore-send-email**
- Extension version: **0.1.37**
- Configuration values (redact info where appropriate):
  - Just use SendGrid and default config

### [REQUIRED] Step 3: Describe the problem
`"TypeError: Cannot read properties of undefined (reading 'attachments')"` when defining:
```js
  to: "test-assertion@email.com",
  sendGrid: {
    templateId: "d-61eb13...",
    dynamicTemplateData: {}
  },
```

Can be fixed by defining a message, as such:
```js
  to: "test-assertion@email.com",
  message: {},
  sendGrid: {
    templateId: "d-61eb13...",
    dynamicTemplateData: {}
  },
```

#### Steps to reproduce:

1. Set the extension up to use SendGrid
2. Create the above doc in the Firestore collection

##### Expected result
Sends email

##### Actual result
`"TypeError: Cannot read properties of undefined (reading 'attachments')"` when defining:

PR #2385, opened 2025-04-22
fix(firestore-send-email): fix issue with empty message property causing error when using SendGrid

CorieW:
Resolves https://github.com/firebase/extensions/issues/2372

Tested


cabljac (approved):
lgtm

Issue #2407, opened 2025-05-06
🐛 [fs-bq-import-collection] Cursor file causing issues


CorieW:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **fs-bq-import-collection**
- Extension version: **0.1.23**

### [REQUIRED] Step 3: Describe the problem

When importing, there is a cursor file that keeps track of where you got up to when importing. I believe this should be removed on completion of importing, as it just prevents you from re-importing if necessary. It's useful keeping in the context of an interrupted import, but once import has completed, remove it.

#### Steps to reproduce:

1. Import new docs successfully
2. Import again

After some additional testing, it seems to occur when the raw_changelog was recently recreated.

##### Expected result
Re-imports.

##### Actual result
Will import 0 new documents.

Issue #2406, opened 2025-05-06
🐛 [fs-bq-import-collection] Inconsistent, importing different numbers of rows to BigQuery


CorieW:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **fs-bq-import-collection**
- Extension version: **0.1.23**

### [REQUIRED] Step 3: Describe the problem

Inconsistent with the number of Firestore rows being imported. With a collection of a 1000 documents, I'll sometimes receive 1000 (as expected), but other times I'll receive 700, 500, or just 100.

#### Steps to reproduce:

It'll be easier (as pointer file can be easily deleted) to just build from source run with the following:
```bash
node ./lib/index.js --non-interactive -P vertex-testing-1efc3 -B dev-extensions-testing -s corie_testing_23947 -d corie_testing_23947 -m false -t corie_testing_23947 -l us-east1 -q false
```
Note: `./lib/index.js` points to the built version of **fs-bq-import-collection**. Also, the pointer file (in the lib directory) has been removed each time, so that shouldn't affect things.

After some additional testing, it seems to occur when the raw_changelog was recently recreated.

##### Expected result
Consistently imports the number of rows the collection contains.

##### Actual result
Imports differing numbers of rows every so often.

PR #2413, opened 2025-05-08
fix(import-script): await runSingleThread

cabljac:
This PR resolves https://github.com/firebase/extensions/issues/2407 and also adds additional handling for when the collection cannot be found.

Also fixes https://github.com/firebase/extensions/issues/2406


CorieW (approved):
lgtm

Issue #2419, opened 2025-05-15
🐛 [firestore-send-email] Validation in version 0.2.1 breaks existing behaviour


philitell:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs for a **specific extension in this repository**.
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

- For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
  with the firebase tag.
- For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
  google group.
- To file a bug against the Firebase Extensions platform, or for an issue affecting multiple extensions, please reach out to
  [Firebase support](https://firebase.google.com/support/troubleshooter/contact/) directly.

### [REQUIRED] Step 2: Describe your configuration

- Extension name: `firestore-send-email`
- Extension version: 0.2.1

### [REQUIRED] Step 3: Describe the problem

New validation in Version 0.2.1 breaks existing behaviour. The new validation forces html- or text- property to be not null (error: "ValidationError: Invalid email configuration: Field 'message.html' must be a string."). When using templates those fields are not in use (at least in my case). So updating to version 0.2.1 breaks my production app.
The field "template" is also not nullable anymore (error: "Field 'template' must be object")
Is there a way to deactivte the validation? Is there a way for my to go back to a previous version?

cabljac:
HI, can you provide your extension configuration? Are you using sendgrid, or some other smtp provider?

cabljac:
ah reproduced this, looking into a fix now. Thanks for raising!

cabljac:
In the meantime if you want to downgrade you can alter the URL here, specify the version at the end:

https://console.firebase.google.com/project/_/extensions/install?ref=firebase%2Ffirestore-send-email@0.2.0

cabljac:
Reopening until i release this. After testing I need to make some more changes before we release.

philitell:
> HI, can you provide your extension configuration? Are you using sendgrid, or some other smtp provider?

I using a standard smtp provider (username + passwort) without any OAuth2 
Thanks for working on this issue!!

cabljac:
No worries! Plan to get a release out tomorrow fixing this.

philitell:
For the sake of completeness: this is an example for an email that works fine in version 0.2.0 and that fails due to validation errors in version 0.2.1:
```
{
  "delivery": {
    "attempts": 1,
[...]

PR #2423, opened 2025-05-15
fix(firestore-send-email): preparePayload fix

cabljac:
This PR is to actually fix https://github.com/firebase/extensions/issues/2419

We extract preparePayload into a dependency injection pattern, and cover it with unit tests, and ensure template/message merging is being done correctly


CorieW (approved):
Lgtm

Issue #2436, opened 2025-05-23
🐛 [firestore-bigquery-export] Wrong reference path passed to firestore doc


YounesAmalou:
### Describe your configuration

- Extension name: **firestore-bigquery-export**
- Extension version: **0.1.24**
- Configuration values (redact info where appropriate):
  - **Firebase project ID**: mtp-dev-001
  - **BigQuery project ID**: mtp-dev-001
  - **Firestore collection path**: Users/{userid}/Entries
  - **Use Collection Group query**: Yes
  - **BigQuery dataset ID**: firestore_export
  - **BigQuery table prefix**: userentries
  - **Documents per import batch**: 300
  - **BigQuery dataset location**: us
  - **Use multithreaded import**: Yes
  - **Use optimized snapshot query**: Yes
  - **Transform function URL**: *(None)*
  - **Use local Firestore emulator**: No
  - **Failed import output location**: *(None)*

#### Steps to reproduce:

I had pre-existing collections in Firestore, and while trying to import them using GCP Shell, I faced the error down below, I tried with different settings, and the same error keeps occurring.
The error happens for each document, and no output is resolved.

##### Expected result
The documents should be imported into the destinated BigQuery table.
##### Actual result
```
{"severity":"INFO","message":"BigQuery dataset already exists: firestore_export"}
{"severity":"WARNING","message":"Did not add partitioning to schema: Partitioning not enabled"}
{"severity":"INFO","message":"Clustering removed on userentries_raw_changelog"}
{"severity":"INFO","message":"Created BigQuery table: userentries_raw_changelog"}
{"severity":"WARNING","message":"Error caught creating table Provided Schema does not match Table mtp-dev-001:firestore_export.userentries_raw_latest. Field path_params is missing in new schema"}
Wait a few seconds for the dataset to initialize...
Importing data from Cloud Firestore Collection (via a Collection Group query): Users/{userid}/Entries, to BigQuery Dataset: firestore_export, Table: userentries_raw_changelog
(node:4445) AutopaginateTrueWarning: Autopaginate will always be set to false in stream paging methods. See more info at https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md#auto-pagination for more information on how to configure paging calls
(Use `node --trace-warnings ...` to show where the warning was created)
[...]

PR #2437, opened 2025-05-23
fix(firestore-bigquery-export): fixed reference path

YounesAmalou:
Fixed the Firestore doc reference path by passing the part after the "/documents/" only, which is what is needed for the doc function to work properly.
#2436

### DIFF
--- a/firestore-bigquery-export/scripts/import/src/worker.ts
+++ b/firestore-bigquery-export/scripts/import/src/worker.ts
@@ -49,12 +49,16 @@ async function processDocuments(
   // Apply partition boundaries from the serialized query
   // These define the range of documents this worker should process
   if (serializableQuery.startAt?.values?.[0]?.referenceValue) {
-    const startPath = serializableQuery.startAt.values[0].referenceValue;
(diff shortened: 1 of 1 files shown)


YounesAmalou:
#2437

cabljac:
Hi, thanks for opening this! Really appreciate contributions. I'll review ASAP

Hoppi164:
BUMP


This fixes a [critical bug](https://github.com/firebase/extensions/issues/2436) that prevents the [import script](https://github.com/firebase/extensions/blob/master/firestore-bigquery-export/guides/IMPORT_EXISTING_DOCUMENTS.md) from working at all.

It's been broken for over 2 months.

Can you please implement this fix, or provide another solution?

jketcham:
Just to confirm the importance of this fix, I've used this extension multiple times in the past and never had an issue with the import script. I went to add the extension for another collection of firestore records and when I went to run the import script as I have in the past, I got the error referencing the `documentPath` for each document having an odd number of components due to the `/documents/` being present in the path.

I applied the fix proposed in this PR locally and confirmed it fixes the issue with no change to the command I had been trying to run.

PR #2467, opened 2025-07-10
feat(firestore-translate-text): add models and provider

CorieW:
- [ ] Tested

### DIFF
--- a/firestore-translate-text/extension.yaml
+++ b/firestore-translate-text/extension.yaml
@@ -132,26 +132,47 @@ params:
     required: false
 
   - param: TRANSLATION_MODEL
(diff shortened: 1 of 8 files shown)


firestore-translate-text/extension.yaml
```diff
@@ -132,26 +132,64 @@ params:
     required: false
 
   - param: TRANSLATION_MODEL
-    label: Translation Model
+    label: Translation Engine
```
HassanBahati, line 135: Wouldn't `Translation Model` work better given the param is `TRANSLATION_MODEL`. Also considering the file's param-label naming convention?

Issue #2506, opened 2025-08-18
🐛 [storage-resize-images] executions & exponential costs in backfill


TomDUVAL-MAHE:
We are experiencing a **critical incident** related to the official Firebase `storage-resize-images` extension (Google).
We have a function that has been running in backfill mode since July 16 to process our images. We have not made any changes to our extension configuration since July 16. However, starting on July 25, the function began to see its number of calls per minute increase in several steps, with a critical increase on August 8 between 01:00 and 03:00, screens 1–2 show the number of executions per second of the Firebase extension functions (red = image generation, blue = backfill), with a peak at **300 + 160 requests/s** sustained for one hour over approximately four days.

Two functions are implicated and operate together:
- `ext-storage-resize-images-backfillResizedImages`
- `ext-storage-resize-images-generateResizedImage`

🔧  Environment & configuration

- **Region**: `europe-west1`
- **Memory**: `FUNCTION_MEMORY=2048`
- **Backfill**: `DO_BACKFILL=true`
- **Types & sizes**:
  `IMAGE_TYPE=avif`
  `IMG_SIZES=640x480,1200x800,1920x1280`
  `IS_ANIMATED=true`
  `OUTPUT_OPTIONS={"avif":{"quality":100}}`
  `SHARP_OPTIONS={"fit":"cover"}`
- **Paths**:
  `RESIZED_IMAGES_PATH=thumbnails`
  `EXCLUDE_PATH_LIST=/maps/*/thumbnails,/pois/*/thumbnails`
- **Other**:
  `DELETE_ORIGINAL_FILE=false`
  `MAKE_PUBLIC=true`
  `CONTENT_FILTER_LEVEL=OFF`
  `REGENERATE_TOKEN=false`
- **Impacted instances**:
  `ext-storage-resize-images-backfillResizedImages`
  `ext-storage-resize-images-generateResizedImage`
- **Dataset**: ~600k images (JPG/PNG) → multi-size AVIF generation (~200 GB)

> **Important**: We did not modify the extension’s code, only its parameters via the .env file. We have not yet migrated to the recently published version, exposing additional concurrency/batch parameters.

⏱️ Timeline & metrics
**IC** = Instance Count
**I/s** = Instances per second
**Jul 20 → 24:** IC 0,015k → 0,017k ; I/s 1.5 → 2
**Jul 25 → 29:** IC 0,038k → 0,073k ; I/s 3.45 → 5
[...]

PR #2534, opened 2025-09-30
fix!(storage-resize-images): disable backfill

cabljac:
This PR disables the backfill feature of storage-resize-images, to prevent issues like https://github.com/firebase/extensions/issues/2506

After investigation, we believe the created thread of Cloud Tasks could be branching, leading to non-linear growth of the queue, and the extension functions attempting to process images that have already been resized.

We have not been able to reproduce this directly, but this draft PR is a preventative step.


nathantangent:
Hey, do we know when the backfill will be fixed?

PR #2736, opened 2026-04-06
Update firestore-counter Dart lockfile

hizzlwizzl:
This PR contains only the lockfile update referenced by the barfuss_yoga workspace.

Changes:
- updates firestore-counter/clients/dart/pubspec.lock

Reason:
- preserve the exact extensions commit referenced by the main workspace while keeping the PR diff minimal and reviewable


hizzlwizzl:
Closing this PR because the lockfile update is not compatible with the current pubspec.yaml SDK constraints and is kept only as a reference branch.

Issue #2573, opened 2025-12-26
🐛 [firestore-translate-text] Broken formatting when using Gemini


JanKubicek:
### Configuration

- Extension name: **firestore-translate-text**
- Extension version: **0.1.25**
- Configuration values:
  - Translation Provider: **Gemini**

### Describe the problem

When using the extension to translate text, the original formatting is not preserved. Specifically, line breaks (newlines) are removed or collapsed into a single paragraph after translation.

common.ts:120 contains this code:

const sanitizedText = text
        .replace(/\\/g, "\\\\")
        .replace(/"/g, '\\"')
        .replace(/\n/g, " ");

which probably causes this issue.

#### Steps to reproduce:

1. Select or input text containing multiple lines / paragraphs
2. Use the extension to translate the text
3. Observe the translated output

##### Expected result

The translated text should preserve the original formatting.

##### Actual result

All line breaks are removed, and the translated text is returned as a single block of text.

JanKubicek:
https://github.com/firebase/extensions/pull/2574

dackers86:
Hi @JanKubicek 

Thank-you for providing the issue and accompanying PR. We will review this and get back to you with any questions and/or feedback.

cabljac:
Reproduced - and the cause is right where your PR points: the input sanitisation in `functions/src/translate/common.ts` replaces `\n` with spaces *before* the text reaches Gemini, so no prompt instruction can preserve formatting that's already been stripped. Thanks for #2574, and apologies it's sat in the queue (the release pipeline was blocked for a while). It's tracked and in the review backlog - no timeline commitment, but the issue is now a confirmed bug with a candidate fix attached. Marking reproduced.

PR #2574, opened 2025-12-27
fix(firestore-translate-text): removed unnecessary sanitization

JanKubicek:
Removes unnecessary text sanitization that caused formatting issues during translation.

PR #2745, opened 2026-04-08
chore(firestore-translate-text/deps): switch from `googleai` and `vertexai` Genkit plugins to `google-genai`

CorieW:
**Description:**

Fixes the issue https://github.com/firebase/extensions/pull/2573 when merged alongside https://github.com/firebase/extensions/pull/2574. From testing, Genkit's `googleai` plugin seems to do some sanitization. I believe this is likely not an issue in the plugin's code, as I couldn't find any sanitization there. This problem likely occurs in the version of the GoogleAI SDK used in the `googleai` plugin. This PR introduces `google-genai` plugin to fix this sanitization problem. Switching to using `google-genai` plugin is also recommended, as it's more up-to-date.

**Testing:**

image

### DIFF
--- a/firestore-translate-text/CHANGELOG.md
+++ b/firestore-translate-text/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Version 0.1.28
 
-chore - switch from `googleai` and `vertexai` Genkit plugins to `google-genai`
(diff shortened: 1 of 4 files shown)


cabljac (approved):
lgtm!

Issue #2762, opened 2026-04-13
🐛 [storage-resize-images] Blocked images by Vertex AI are not moved to the failed path due to bypassed catch block


tomassator:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: **storage-resize-images**
- Extension version: **0.3.2**
- Configuration values (redact info where appropriate):
  - **Cloud Storage path for failed images**: `failed`
  - **Content filter level**: `High strictness` 
  - **Path to placeholder image**: `placeholder/blocked-content.png`
  - **Deletion of original file**: `Delete on any resize attempt` (Bug happens with 'Keep' as well)
  - **Convert image to preferred types**: `webp`

#### Steps to reproduce:

1. Install the `storage-resize-images` extension with Content Filtering enabled (Vertex AI API enabled).
2. Configure a valid path for `Cloud Storage path for failed images` (e.g., `failed`).
3. Configure a custom placeholder image.
4. Upload an image that triggers the content filter (e.g., inappropriate content).
5. Review the Storage bucket and the Cloud Function logs.

##### Expected result
According to the official documentation:
> "If you’ve configured a failed images path, the original image is moved to that location with metadata indicating it was blocked by content filtering."

The expected behavior is that the extension generates the resized placeholder versions AND moves the original malicious image to the `failed` path, adding the rejection metadata.

##### Actual result
The extension successfully detects the inappropriate content and successfully generates the resized placeholder images (e.g., `_200x200.webp`, `_600x600.webp`). However, **the original blocked image is NEVER moved to the failed images path**, nor is any metadata applied to it. 

**Technical context (Why this is happening):**
Looking at the Cloud Function logs, when Vertex AI blocks the content, the extension successfully swaps the image buffer with the placeholder buffer and completes the resizing process. Because this placeholder replacement is successful, the function finishes with `status: 'ok'`. 

It appears the logic to move the original file to the `failed` path and apply rejection metadata is currently placed inside an error handling block (`catch`). Since the placeholder generation prevents the function from throwing an error, the `catch` block is completely bypassed, leaving the original blocked image in its source directory.

Log sequence observed:
1. `Replacing filtered image with configured placeholder...`
2. `Image content blocked by Vertex AI content filters.`
3. `Successfully replaced filtered image with placeholder...`
4. `Resizing image at path...`
5. `Function execution took X ms, finished with status: 'ok'`
*(Missing: Any log indicating "Moving original file to failed path")*

cabljac:
[...]

PR #2780, opened 2026-04-20
fix: images that fail content filter are added to failed directory

### LINKED ISSUE #2769: 🐛 [storage-resize-images] Content filter does not work
Yes — bug specific to `storage-resize-images`.

- Extension name: `storage-resize-images`
- Extension version: latest (master at time of filing)
- Cloud Function region: `europe-west2`
- Configuration values:
  - Content filter level: `BLOCK_LOW_AND_ABOVE` (also reproduced with `BLOCK_MEDIUM_AND_ABOVE`)
  - Custom content filter prompt: tested with multiple variants (see below)
  - Path to placeholder image: default
  - All other settings: defaults

The content filtering feature silently fails open on borderline NSFW images. Inappropriate images that should be blocked are processed normally with no placeholder substitution, despite the filter being configured at the strictest level. The root cause is a mismatch between the error shape the extension's catch handler expects and the actual error shape produced by Gemini 2.5 Flash when its input-side safety refuses to respond.

There are two related bugs in `functions/src/content-filter.ts`:

**Bug 1 — Schema validation failure on safety refusal is not treated as a block.**

When Gemini 2.5 Flash's input-side safety declines to respond to an image (which happens often on borderline content — exactly the content the filter should be catching), it returns an empty response with no `finishReason`. Genkit's structured output validator then throws `INVALID_ARGUMENT: Schema validation failed` because `null` doesn't match the required `{response: string}` schema.

The catch handler in `performContentCheck` only recognises errors with `error.detail?.response?.finishReason === "blocked"`:

```ts
} catch (error) {
  if (error.detail?.response?.finishReason === "blocked") {
    log.contentFilterBlocked();
    return false;
  }
  throw error;
}
```

The schema validation error doesn't have that shape, so it gets rethrown. After 3 retries (which all fail identically because the failure is deterministic, not transient), the error propagates up to `processContentFilter`, which catches it, sets `failed = true`, and returns. Because the placeholder substitution logic only runs when `filterResult === false` (not when `failed === true`), the original image is processed without any filtering applied.

**Bug 2 — The no-prompt path cannot block based on model verdict.**

When no custom prompt is configured, `effectivePrompt` is `"Is this image appropriate?"` with `maxOutputTokens: 1`, but the verdict-checking branch is gated on `prompt !== null`:

```ts
if (result.output?.response === "yes" && prompt !== null) {
  log.customFilterBlocked();
  return false;
}
return true;
```

This means without a custom prompt, the model's response is never inspected — the function always returns `true` unless the catch handler fires. Combined with Bug 1, this means users who select a content filter level but don't set a custom prompt get effectively no filtering at all, even though the UI implies otherwise.

#### Steps to reproduce:

1. Install `storage-resize-images` with `CONTENT_FILTER_LEVEL=BLOCK_LOW_AND_ABOVE`.
2. Set `CUSTOM_FILTER_PROMPT` to anything moderation-shaped, e.g.:
```
   You are classifying images for a family-friendly platform. Look at the image and decide whether it depicts any of: people in bed together, people who appear undressed or covered only by sheets, romantic kissing, or intimate physical contact. Answer 'yes' if any of these are present, otherwise 'no'.
```
3. Upload a borderline NSFW image (e.g. a stock photo of people in bed together, no explicit nudity).
4. Observe Cloud Functions logs for `ext-storage-resize-images-generateResizedImage`.

##### Expected result
The image is recognised as inappropriate and replaced with the placeholder image before resizing. A log entry indicates the content filter blocked the image.

##### Actual result
Three retry attempts fail with identical errors:

```
Unexpected Error whilst evaluating content of image with Gemini (Attempt 2/3).
ValidationError [GenkitError]: INVALID_ARGUMENT: Schema validation failed. Parse Errors:

- (root): must be object

Provided data:

null

Required JSON schema:

{
  "type": "object",
  "properties": {
    "response": { "type": "string" }
  },
  "required": ["response"],
  "additionalProperties": true,
  "$schema": "http://json-schema.org/draft-07/schema#"
}
    at parseSchema (/workspace/node_modules/@genkit-ai/core/lib/schema.js:104:21)
    at GenerateResponse.assertValidSchema (/workspace/node_modules/@genkit-ai/ai/lib/generate/response.js:78:37)
    ...
  status: 'INVALID_ARGUMENT',
  code: 400,
```

The retries don't help because the failure is deterministic (same image + same prompt → same null response every time). After the final retry, the error propagates to `processContentFilter`, the image is resized normally, and **no placeholder is substituted**. From the user's perspective, the filter has silently allowed inappropriate content through.

Removing the custom prompt and relying only on `CONTENT_FILTER_LEVEL` does not help, because of Bug 2 above — the no-prompt path's verdict check is gated on `prompt !== null`, so the model's response is never inspected.

#### Suggested fix

The catch handler in `performContentCheck` should treat schema validation failures with null content as implicit blocks, since in practice an empty response from Gemini on an image input almost always indicates the input-side safety system declined to engage:

```ts
} catch (error) {
  if (error.detail?.response?.finishReason === "blocked") {
    log.contentFilterBlocked();
    return false;
  }
  // Gemini 2.5 Flash returns null content (rather than a clean
  // SAFETY finishReason) when input-side safety refuses on image
  // inputs. Genkit surfaces this as a schema validation error.
  // Treat it as a block.
  if (
    error.status === "INVALID_ARGUMENT" &&
    error.originalMessage?.includes("Provided data:\n\nnull")
  ) {
    log.contentFilterBlocked();
    return false;
  }
  throw error;
}
```

Additionally, the `prompt !== null` gate on the verdict check should be removed so the no-prompt path can actually use the model's response (or the no-prompt path should be removed entirely if it isn't intended to do verdict-based filtering).

Without one of these fixes, the content filtering feature does not function as documented for the most common use case (borderline NSFW imagery on a user-upload platform), and fails silently in a way that's only discoverable by reading Cloud Functions logs and the extension source code.

IzaakGough:
Fixes https://github.com/firebase/extensions/issues/2769 and https://github.com/firebase/extensions/issues/2762.

Refactors content-filter and adds a fix for the silent fail-open on Gemini 2.5 Flash safety refusals reported in #2769.

## What changed

- **Bug 1 fix (#2769)** — any genkit `ValidationError` thrown by the moderation `ai.generate` call is now treated as an implicit block. We control the prompt and schema in that path, so the only thing that can fail validation is the model's response. Observed shapes from Gemini 2.5 Flash safety refusals: `data === null` (empty content), `data === {}` (refusal text that extractJson coerced to an empty object), and we also block on plain type-mismatch for the same reason (a model that won't produce a usable verdict on borderline imagery is almost always an input-side refusal, and failing open there is the original bug). Detection is structural: `status === "INVALID_ARGUMENT"` plus a populated `detail.errors` array. Returned directly without the retry queue (deterministic, retries just burn API calls).
- **Earlier on the branch** — refactor of `content-filter.ts`, blocked-image routing to the failed-image path before placeholder substitution, placeholder swap on a copy of the original (so blocked originals are preserved), data-URL contentType fix.
- Drops a leftover `as any` and clarifies the `maxOutputTokens` comment per Izaak's review on #2762.

## Tests

- 4 new unit tests in `__tests__/content-filter.test.ts`:
  - null-content `ValidationError` → blocks **without retrying** (asserts `mockGenerate.toHaveBeenCalledTimes(1)`)
  - empty-object `ValidationError` → blocks without retrying
  - type-mismatch `ValidationError` → blocks without retrying
  - non-`ValidationError` (e.g. ECONNRESET) → still goes through retries and propagates
- New env-gated live integration suite at `__tests__/integration/content-filter.live.test.ts`. Runs against real Vertex AI Gemini 2.5 Flash:
  - safe text image + `BLOCK_ONLY_HIGH` → passes
  - gun image + weapon-detection custom prompt → blocked
  - safe image + weapon-detection custom prompt → not blocked (model doesn't over-trigger)
  - 2 borderline-regression tests gated on `LIVE_BORDERLINE_IMAGE_PATH` (image not committed; supply your own)
- Run via `npm run test:live-content-filter --prefix storage-resize-images/functions` (env vars: `GCLOUD_PROJECT`, optionally `LIVE_BORDERLINE_IMAGE_PATH`).
- All 3 always-on live tests verified passing against `izaak-testing` (~9s).

## Behaviour change / out of scope

- The bug report also flags a "Bug 2" about the no-custom-prompt path ignoring the model's yes/no verdict. **Not addressed in this PR** — separate behaviour question with non-trivial blast radius for existing installs. Tracked for a future follow-up.
- `extension.yaml` bumped to `0.3.4`; CHANGELOG updated.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` clean
- [x] `npx jest __tests__/content-filter.test.ts` — 12 pass (8 existing + 4 new)
- [x] `npx jest __tests__/unit/generateResizedImageHandler.test.ts` — 4 pass unchanged
- [x] `npm run test:live-content-filter` against `izaak-testing` — 3/3 pass
- [ ] Bug 1 regression test against a real borderline image (set `LIVE_BORDERLINE_IMAGE_PATH`)
- [ ] Manual repro of the original report: deploy + upload borderline image, confirm `contentFilterBlocked` log + placeholder substitution

### DIFF
--- a/storage-resize-images/functions/src/content-filter.ts
+++ b/storage-resize-images/functions/src/content-filter.ts
@@ -3,13 +3,6 @@ import { genkit, z } from "genkit";
 import type { SafetyThreshold } from "./config";
 import * as fs from "fs";
 import * as path from "path";
-import { Bucket } from "@google-cloud/storage";
-import { ObjectMetadata } from "firebase-functions/v1/storage";
-import {
-  replaceWithConfiguredPlaceholder,
-  replaceWithDefaultPlaceholder,
-} from "./util";
-// Import the logging functions from your log.ts module
 import * as log from "./logs";
 import { globalRetryQueue } from "./global";
 
@@ -218,48 +211,4 @@ export async function checkImageContent(
 
   // Start the first attempt (not via queue)
   return await attemptWithRetry(1);
-}
-
-/**
- * Runs content filtering and, if the image is blocked, swaps the local file
- * with a placeholder image.
- *
- * @returns `true` if the image passed the filter, `false` if it was blocked
- * (the local file has been replaced with a placeholder).
- * @throws If the filter call errors or the placeholder swap errors. Callers
- * are responsible for treating these as failures.
- */
-export async function processContentFilter(
-  localFile: string,
-  object: ObjectMetadata,
-  bucket: Bucket,
-  config: any
-): Promise<boolean> {
-  const passed = await checkImageContent(
-    localFile,
-    config.contentFilterLevel,
-    config.customFilterPrompt,
-    object.contentType
-  );
-
-  if (passed) {
-    return true;
-  }
-
-  log.contentFilterRejected(object.name);
-
-  if (config.placeholderImagePath) {
-    log.replacingWithConfiguredPlaceholder(config.placeholderImagePath);
-    await replaceWithConfiguredPlaceholder(
-      localFile,
-      bucket,
-      config.placeholderImagePath
-    );
-  } else {
-    log.replacingWithDefaultPlaceholder();
-    await replaceWithDefaultPlaceholder(localFile);
-  }
-  log.placeholderReplaceComplete(localFile);
-
-  return false;
 }
--- a/storage-resize-images/functions/src/content-filter.ts
+++ b/storage-resize-images/functions/src/content-filter.ts
@@ -64,97 +64,9 @@ function createSafetySettings(filterLevel: SafetyThreshold) {
 }
 
 /**
- * Performs the actual content check with the AI model
- * @param localOriginalFile Path to the local image file
- * @param filterLevel The content filter level to apply
- * @param prompt Optional custom prompt to use for content checking
- * @param contentType The content type of the image
- * @returns Promise<boolean> - true if the image passes the filter, false otherwise
- */
-async function performContentCheck(
-  localOriginalFile: string,
-  filterLevel: SafetyThreshold | null,
-  prompt: string | null,
-  contentType: string
-): Promise<boolean> {
-  const dataUrl = createImageDataUrl(localOriginalFile);
-
-  // Initialize Vertex AI client
-  const ai = genkit({
-    plugins: [
-      vertexAI({
-        location: process.env.LOCATION ?? "us-central1",
-        models: ["gemini-2.5-flash"],
-      }),
-    ],
-  });
-
-  // Determine the effective safety settings and prompt to use
-  const effectiveFilterLevel: SafetyThreshold =
-    filterLevel === null ? "BLOCK_NONE" : filterLevel;
-  const effectivePrompt =
-    prompt !== null
-      ? prompt +
-        '\n\n Please respond in json with either { "response": "yes" } or  { "response": "no" }'
-      : "Is this image appropriate?";
-
-  const effectiveOutput =
-    prompt !== null
-      ? {
-          format: "json",
-          schema: z.object({
-            response: z.string(),
-          }),
-        }
-      : undefined;
-
-  // Determine max tokens based on whether we're using custom prompt
-  const maxOutputTokens = prompt !== null ? 100 : 1;
-
-  try {
-    const result = await ai.generate({
-      model: gemini("gemini-2.5-flash"),
-      messages: [
-        {
-          role: "user",
-          content: [
-            {
-              text: effectivePrompt,
-            },
-            {
-              media: {
-                url: dataUrl,
-                contentType,
-              },
-            },
-          ],
-        },
-      ],
-      output: effectiveOutput,
-      config: {
-        temperature: 0.1,
-        maxOutputTokens,
-        safetySettings: createSafetySettings(effectiveFilterLevel),
-      },
-    });
-
-    if (result.output?.response === "yes" && prompt !== null) {
-      log.customFilterBlocked();
-      return false;
-    }
-
-    return true;
-  } catch (error) {
-    if (error.detail?.response?.finishReason === "blocked") {
-      log.contentFilterBlocked();
-      return false;
-    }
-    throw error;
-  }
-}
-
-/**
- * Checks if an image content is appropriate based on the provided filter level and optional custom prompt
+ * Entry point for image moderation: short-circuits when disabled, otherwise runs a single
+ * Vertex/Gemini call per attempt with retries and queue-backed backoff on transient errors.
+ *
  * @param localOriginalFile Path to the local image file
  * @param filterLevel The content filter level to apply ('LOW', 'MEDIUM', 'HIGH', or null to disable)
  * @param prompt Optional custom prompt to use for content checking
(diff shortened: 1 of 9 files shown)


storage-resize-images/functions/src/content-filter.ts
```diff
@@ -67,26 +41,34 @@ function createSafetySettings(filterLevel: SafetyThreshold) {
   return HARM_CATEGORIES.map((category) => ({
     category,
     threshold: filterLevel,
-  }));
+  })) as any;
```
cabljac: not a fan of `as any` escape hatch here, can we fix

storage-resize-images/functions/src/content-filter.ts
```diff
@@ -43,20 +25,12 @@ function sleep(ms: number): Promise<void> {
...
-  | "HARM_CATEGORY_HATE_SPEECH"
-  | "HARM_CATEGORY_DANGEROUS_CONTENT"
-  | "HARM_CATEGORY_HARASSMENT"
-  | "HARM_CATEGORY_SEXUALLY_EXPLICIT";
-
-const HARM_CATEGORIES: ReadonlyArray<SafetyCategory> = [
+const HARM_CATEGORIES = [
   "HARM_CATEGORY_HATE_SPEECH",
   "HARM_CATEGORY_DANGEROUS_CONTENT",
-  "HARM_CATEGORY_UNSPECIFIED",
```
cabljac, line 56: I think this was an actual category from the API, we should double check we're OK to remove this

storage-resize-images/functions/src/content-filter.ts
```diff
@@ -115,80 +99,53 @@ async function performContentCheck(
         }
       : undefined;
 
-  // Determine max tokens based on whether we're using custom prompt
```
cabljac, line 118: let's not remove this comment

storage-resize-images/functions/src/content-filter.ts
```diff
@@ -71,22 +45,64 @@ function createSafetySettings(filterLevel: SafetyThreshold) {
...
+    originalMessage?: string;
+    detail?: { errors?: unknown };
+  };
+  if (e.status !== "INVALID_ARGUMENT") return false;
+  const errors = e.detail?.errors;
+  if (!Array.isArray(errors) || errors.length === 0) return false;
+  return (
+    typeof e.originalMessage === "string" &&
+    /Provided data:\s*\n+\s*null/.test(e.originalMessage)
+  );
```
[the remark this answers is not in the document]
cabljac: New isModerationSchemaRefusal is now broad: any schema-validation failure -> "blocked", not just safety refusals. Future prompt/schema bug → silent false-positive block instead of error. JSDoc :55-60 already justifies (fail-closed > fail-open on NSFW). Accepted policy, just call it out so reviewers don't think it's accidental. We should make sure to detail the change in behaviour in the versioning and changelog

storage-resize-images/functions/src/index.ts
```diff
@@ -79,54 +81,91 @@ const generateResizedImageHandler = async (
...
+          bucket,
+          localOriginalFile,
+          object,
+          parsedPath,
+          blockedByFilter
+        );
+      }
     } else {
       if (config.deleteOriginalFile === deleteImage.onSuccess) {
         await deleteRemoteFile(remoteOriginalFile, filePath);
```
cabljac, line 171: correction to my comment: this onSuccess-deletes-the-blocked-original behaviour isn't introduced by this PR. it already exists on `next` via processContentFilter (placeholder swap, recordSuccessEvent for blocked, and the success-branch onSuccess delete are all pre-existing).

so not a merge blocker. but it's a latent data-loss path: with `deleteOriginalFile = onSuccess` and no `failedImagesPath`, a blocked image gets no copy kept anywhere and the original is deleted. this PR actually improves it (the blocked original is now preserved to `failedImagesPath` when that's set), it just doesn't cover the no-`failedImagesPath` case.

fine to leave as-is here and track the no-`failedImagesPath` gap separately. flagging so it's not lost.

CorieW (approved):
lgtm

PR #2817, opened 2026-05-08
fix(firestore-bigquery-change-tracker): fix the table update check and clustering sync

### LINKED ISSUE #2194: 🐛 [firestore-bigquery-export] tableRequiresUpdate always returning true
### Describe your configuration

- Extension name: firestore-bigquery-export
- Extension version: latest
  - clustering: null

### Describe the problem

The `tableRequiresUpdate` function appears to be bugged in a couple different ways depending on your config. This results in the update check always returning true and the table config being repeatedly updated unnecessarily.

#### Issue 1
If `clustering` config is `null` then the table will always require updates due to this faulty check:
```
const configCluster = JSON.stringify(config.clustering); // Evaluates to 'null'
const tableCluster = JSON.stringify(metadata.clustering?.fields || []); // Evaluates to '[]'
if (configCluster !== tableCluster) return true; // Always true
```

#### Issue 2
Within `initializeRawChangeLogTable` the `pathParamsColExists` is allowed to be `undefined`
```
const pathParamsColExists = fields.find(
        (column) => column.name === "path_params"
      ); // Evaluates to undefined if not present, rather than boolean
```
This is then passed into `tableRequiresUpdate`, and will always fail the strict equality check:
```
if (!!config.wildcardIds !== pathParamsColExists) return true; // Evaluates to boolean !== undefined -> always true
```
#### Steps to reproduce:

What happened? How can we make the problem occur?
Supply either `clustering:null`or don't have `path_params` column.

##### Expected result
The table shouldn't need updating.

##### Actual result
The table always updates on every run.

### LINKED ISSUE #2801: 🐛 [firestore-bigquery-export] handleFailedTransactions crashes on warm containers, masking real BigQuery errors
### Describe your configuration

- Extension name: `firestore-bigquery-export`
- Extension version: `0.3.1`

### Describe the problem

`handleFailedTransactions` in `firestore-bigquery-change-tracker` calls `db.settings({ ignoreUndefinedProperties: true })` inside the exported async function body, then performs batch operations on the same Firestore instance. The Firestore Node SDK's [`settings()` contract](https://googleapis.dev/nodejs/firestore/latest/Firestore.html#settings) requires it be called at most once **and** before any other method is invoked on the same instance.

Because `getFirestore(config.firestoreInstanceId!)` returns the per-process singleton, the second invocation on any warm Cloud Functions container satisfies both blocking conditions (already-called AND other-methods-ran). The result is a `Firestore has already been initialized` error that **masks the original BigQuery insert failure** that triggered the dead-letter path in the first place.

### Steps to reproduce

1. Install `firestore-bigquery-export@0.3.1` with `BACKUP_COLLECTION` set.
2. Trigger an intentional BigQuery insert failure (e.g. schema mismatch, missing dataset, quota exhaustion) so `insertData` exhausts its retries and falls through to `handleFailedTransactions`.
3. First invocation: the dead-letter path succeeds and writes rows to the configured backup collection.
4. Second invocation on the same warm container: throws with the stack trace below.

### Expected result

The dead-letter path writes to the backup collection on every failure, without conflicting with the Firestore instance's initialization.

### Actual result

```
Error: Firestore has already been initialized. You can only call settings() once, and only before calling any other methods on a Firestore object.
    at Firestore.settings (/workspace/node_modules/@google-cloud/firestore/build/src/index.js:554:19)
    at exports.default (/workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/lib/bigquery/handleFailedTransactions.js:11:8)
    at FirestoreBigQueryEventHistoryTracker.insertData (/workspace/node_modules/@firebaseextensions/firestore-bigquery-change-tracker/lib/bigquery/index.js:192:62)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

### Root cause

The 2022 version of [`handleFailedTransactions.ts`](https://github.com/firebase/extensions/blob/ef272b27/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/handleFailedTransactions.ts) called `settings()` once inside an `if (!firebase.apps.length)` guard at module scope — correct usage.

[PR #2364](https://github.com/firebase/extensions/pull/2364) ("support non-default db", April 2025) refactored the module to use `getFirestore(config.firestoreInstanceId!)` but moved the `settings()` call out of the init guard and into the exported function body. Every invocation now retries `settings()` on the singleton Firestore instance, which throws after the first call.

[Current code at commit `b899c58e`](https://github.com/firebase/extensions/blob/b899c58e338b5da380935d00905628bd627f5f11/firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/handleFailedTransactions.ts#L10-L22):

```ts
export default async (
  rows: any[],
  config: ChangeTrackerConfig,
  e: Error,
): Promise<void> => {
  const db = getFirestore(config.firestoreInstanceId!);
  db.settings({
    ignoreUndefinedProperties: true,
  });
  const batchArray = [db.batch()];
  // ... batch.set(), batch.commit() on the same db
};
```

### Suggested fix

Either:

1. **Init-once at module scope.** Call `settings()` in a module-level guard (keyed by `firestoreInstanceId` if non-default DBs need distinct settings), restoring the pattern from the 2022 version.
2. **Use a distinct Firestore instance.** Scope the `settings()` call to a new `Firestore` object rather than the process singleton.

### Impact

- The dead-letter path silently crashes on every failure after the first on a warm container, instead of recording failed rows to the backup collection.
- The original BigQuery error — which is typically what operators need to see — is masked by the secondary `settings()` crash.
- Operators troubleshooting insert failures waste cycles chasing the Firestore-initialization error before realizing it's a wrapper around something else entirely.

### Additional context

Filed after hitting this while upgrading from `0.2.6` to `0.3.1` in a production Cloud Functions environment. No matching upstream reports found on a search of `"Firestore has already been initialized"` / `"can only call settings"` / `handleFailedTransactions` — this appears to be the first public report.

Happy to open a PR implementing option (1) if that's the preferred direction.

cabljac:
Fixes #2194. (#2801, the `db.settings()` guard, was originally part of this branch but landed on next with #2937; this PR no longer touches that file.)

#2194: `tableRequiresUpdate` always returned true, via two accidental triggers: `JSON.stringify(config.clustering)` gives `"null"` vs the table's `"[]"`, and `fields.find(...)` fed `undefined` into a boolean comparison. Fixed with `config.clustering || []` and `fields.some(...)`. The `find` -> `some` change is behavioral, not cosmetic: clustering sync on existing tables depended on the always-true trigger, because `updateClustering` mutated the shared metadata before `tableRequiresUpdate` compared it (desired vs desired). A third commit moves `updateClustering` after the check, inside `shouldUpdate`. The `some` fix has no behavioral regression guard (offline tests pass booleans directly, and the live suites assert final table state, which the spurious update also produced), so `fields` is now typed `TableField[]` and a revert to `find` fails compilation.

Rebased onto next after #2937 merged. The behavior change disclosed earlier is now narrower: `tableRequiresUpdate` gained a trigger that fires when the configured custom partition column is missing from the table's schema, so the column is added by a real metadata update on the next initialize instead of by the old spurious path, and the trigger stops firing once the column lands. What remains: an insert racing the column's schema propagation fails terminally with its rows backed up intact, since the column is deliberately not stripped for a lag retry (a null there misfiles the row into the wrong partition permanently).

Verified against live BigQuery: clustering, insertRetry, backupSettings, and checkUpdates suites all pass; clustering, checkUpdates, and partitioning (with `RUN_BIGQUERY_INTEGRATION_TESTS=true`) re-run after the review follow-up.


firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
```diff
@@ -281,11 +281,12 @@ export class FirestoreBigQueryEventHistoryTracker
...
     }
 
-    // The custom partition column is deliberately not allowlisted. Initialize
-    // can still add it to an existing table (the clustering comparison in
-    // `tableRequiresUpdate` returns true for default installs), so it can lag;
-    // when it does, the insert fails terminally and the row is backed up
-    // intact rather than stripped and retried. Safe in the lossy direction.
+    // The custom partition column is deliberately not allowlisted.
+    // `addPartitioningToSchema` runs whenever any `tableRequiresUpdate`
+    // trigger fires, but only the unpartitioned-table case fires reliably;
```
IzaakGough: On an already-partitioned existing table the custom partition column now never reaches the schema at all, and inserts carrying it fail permanently.

`isValidPartitionForExistingTable()` returns true only when the table is unpartitioned, and it is the only partitioning trigger in `tableRequiresUpdate`. So with the clustering and wildcard triggers fixed, `shouldUpdate` is false and `addPartitioningToSchema` does not run.

Failure path: installed with ingestion-time or `timestamp` partitioning (table created partitioned), then reconfigured with `TIME_PARTITIONING_FIELD=my_ts` plus a Firestore field, or the field just renamed. `getPartitionValue` puts `my_ts` in every row whose document has that field. The insert runs with `ignoreUnknownValues: false`, `my_ts` is not in `columnsAddedToExistingTables()` so `unknownFieldColumn` returns null, the error is not transient, and the batch goes terminal: every row to the backup collection plus a rethrow. It cannot recover, because the terminal path sets `_initialized = false` and the next `initialize` again decides no update is needed.

One option that keeps the rest of this intact: add a trigger to `tableRequiresUpdate` for the desired partition field being absent from `metadata.schema.fields`. `getNewPartitionField()` already computes the field.

firestore-bigquery-export/firestore-bigquery-change-tracker/CHANGELOG.md
```diff
@@ -8,5 +8,8 @@ Insert-failure semantics changed. Since 2020, the retry guard in `insertData` wa
 - Every other failure is terminal: the original, unmodified rows are written to the configured backup collection (`backupTableId`, keyed by `insertId`) and the insert error is rethrown to the caller.
 - Fixed the backup path calling `db.settings()` more than once per Firestore instance, which made every backup after the first throw and replaced the insert error. A failed backup write is now logged and the original insert error is still the one thrown.
 - Backup documents now populate `error_details` (the field previously existed but was always empty).
+- Stopped the spurious table update on every cold start: a broken comparison made `tableRequiresUpdate` return true for default installs (null clustering compared against `[]`, and a truthiness bug in the wildcard-column check), so `setMetadata` ran on every initialize. Both comparisons are fixed; this reduces, not eliminates, repeat updates (a few pre-existing perpetual-update configurations remain, such as wildcards toggled off with `path_params` still present).
```
IzaakGough: Two more configurations still re-fire the update on every cold start, and naming them here would help operators who see repeated `setMetadata` calls and wonder if it is new:

- Partitioning configured against a pre-existing unpartitioned table. Nothing in the `tableExists` branch calls `partitioning.updateTableMetadata`, so the table never actually becomes partitioned and `isValidPartitionForExistingTable()` stays true.
- An invalid clustering config, including `CLUSTERING=a, b`, where `split(",")` leaves an untrimmed ` b`. `hasInvalidFields` declines to apply it, so the mismatch persists.

firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/checkUpdatesUnit.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    datasetId: "d",
+    tableId: "t",
+    datasetLocation: "us",
+    wildcardIds: false,
+    clustering: null,
+    ...overrides,
+  } as ChangeTrackerConfig;
+}
+
+const allColumnsPresent = {
```
IzaakGough: These tests pin the clustering comparison nicely. They cannot pin the `find` to `some` change though: that bug is in `initializeRawChangeLogTable` passing a `TableField` object into a `boolean` parameter, and these cases pass booleans in directly, so they behave identically either way.

The case that would at least exercise the no-op direction of the wildcard comparison, `wildcardIds: true` with `pathParamsColExists: true`, is not reachable here since `allColumnsPresent` fixes `pathParamsColExists` to false. Worth adding, and worth flagging that the `some` change has no regression guard as things stand.

firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/checkUpdatesUnit.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  test("reordered clustering fires an update", async () => {
+    const result = await tableRequiresUpdate({
+      table: stubTable(["data", "timestamp"]),
+      config: baseConfig({ clustering: ["timestamp", "data"] }),
+      ...allColumnsPresent,
+    });
+    expect(result).toBe(true);
+  });
+
+  test("wildcards off with no path_params column is a no-op", async () => {
```
IzaakGough: Minor: this has the same inputs as "null clustering config against an unclustered table is a no-op" above (`stubTable(undefined)`, `clustering: null`, `wildcardIds: false`), so it does not add an assertion.

Also `allColumnsPresent` sets `pathParamsColExists: false`, which reads a little oddly against the name.

firestore-bigquery-export/firestore-bigquery-change-tracker/src/bigquery/index.ts
cabljac: Implemented as you suggested: `tableRequiresUpdate` now fires when the configured partition column is absent from `metadata.schema.fields`. The check lives on `Partitioning` as `customPartitionFieldMissingFromSchema(fields)`, which is pure (config plus the fields passed in, no network call), and `addPartitioningToSchema` is now gated on the same method. Convergence follows from that sharing: the update pushes exactly the column the trigger checks for, so once `setMetadata` lands it in the schema the trigger stops firing. Configs without a custom partition column are a no-op (`getNewPartitionField()` returns null), and FIRESTORE_TIMESTAMP partitioning stays quiet because `timestamp` is always in the changelog schema.

Offline unit tests pin the missing-column trigger (red before the fix), the present-column no-op, and the timestamp no-op. Live clustering, checkUpdates, and partitioning (`RUN_BIGQUERY_INTEGRATION_TESTS=true`) suites all pass after the change.

firestore-bigquery-export/firestore-bigquery-change-tracker/CHANGELOG.md
cabljac: Done. The 2.1.0 entry now names both: partitioning configured against a pre-existing unpartitioned table (the `tableExists` branch never applies the partitioning to the table itself, so the check keeps firing), and invalid clustering configs including the untrimmed `CLUSTERING=a, b`.

firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/checkUpdatesUnit.test.ts
cabljac: Added the missing cell: wildcards on with `path_params` present is a no-op. Also its sibling, wildcards off with `path_params` still present fires an update, which pins the changelog's perpetual-update example.

On the `find` -> `some` change: you're right that these tests can't pin it, and it turned out worse than that. I reverted `some` back to `find` and ran the live clustering and checkUpdates suites; all 40 tests still pass, because they assert final table state, which the spurious always-update also produced. So nothing pinned the change at all. Rather than contrive an offline pin, `fields` in `_initializeRawChangeLogTable` is now typed `TableField[]`, so a revert to `find` fails compilation (TS2322: `ITableFieldSchema` is not assignable to `boolean`).

firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/checkUpdatesUnit.test.ts
cabljac: Done. The duplicate test is folded away, and `allColumnsPresent` is renamed to `columnsForNonWildcardTable`.

PR #2913, opened 2026-07-17
feat: support deletion of nested field values

### LINKED ISSUE #2910: 🐛 [delete-user-data] Not all data is deleted

- Extension name: **delete-user-data** (`storage-resize-images`, `firestore-send-email`, etc)
- Extension version: **0.1.29**
  - Cloud Firestore paths: **empty**
  - Cloud Firestore delete mode: **recursive**
  - Enable auto discovery: **yes**
  - Auto discovery search depth: **3**
  - Auto discovery search fields (Optional): **uid,userId,creatorId,scope.owner.id**

Hi! 

It looks like the extension has the implicit assumption that every document key contains a user ID but that may not always be the case. Or the auto discovery is not being applied. For example, consider a table "session" that has `userId` as a property but uses keys like `session/<session ID>`. Looking at the current documentation and implementation, it seems like it is not possible to delete document for `session`

Case 1: Cloud Firestore paths is empty -> Enable auto discovery is not being applied
Case 2: Cloud Firestore paths is set to `sessions` -> This will remove all sessions according to the current code I believe:
```js
  const paths = extractUserPaths(databasePaths, uid);
  const promises = paths.map(async (path) => {
    try {
      logs.rtdbPathDeleting(path);
      await admin.database().ref(path).remove(); // <-- Deletes whole table, we don't want that
      logs.rtdbPathDeleted(path);
    } catch (err) {
      logs.rtdbPathError(path, err);
    }
  });
```
Case 3: Cloud Firestore paths is set to `sessions/{UID}` -> This will not work because the UID is not part of the key. 

Are my assumptions correct or is it somehow possible to delete `session` with the current version? If not, I'd say it's a bug that the auto discovery isn't applied. Or is the issue that properties like `A.B.C` are not being resolved? What happens if the key is inside an object within the document?

Thanks a lot!

#### Steps to reproduce:

See above

##### Expected result

I'd expect that the auto discovery finds the documents in Case 1 and deletes it.

##### Actual result

Documents are not being deleted.

IzaakGough:
Fixes #2910 

This PR adds support for deleting documents where the uid is a nested value.

## Testing
- Verified the fix working in emulator.
- Automated tests pass


filipre:
Can we run "GitHub Actions Scan"?

Issue #495, opened 2020-11-13
(Question) Apply tranlate-text to the existed fields.


ssohymind:
Hi, I'm trying to use the translate-text for my project. I found that this extension only works for the document which is created after the installation of the extension. Could you let me know the way to apply this extension to my previous documents which have the same field name for the extension?

russellwheatley:
Hey @ssohymind, I'm afraid it isn't possible to translate existing records as the cloud function `onWrite` is triggered on create, update or deleting of records.

You could write your own script. Here's a sample how to implement translation of a [string](https://github.com/googleapis/nodejs-translate/blob/master/samples/quickstart.js).

In your script, you would query your Firestore records, loop through them and update the field of your choice with the translation.

If you have a lot of records, you will need to query your database in batches following Firestore [best practices](https://firebase.google.com/docs/firestore/best-practices#realtime_updates). 

The [import](https://github.com/firebase/extensions/blob/next/firestore-bigquery-export/scripts/import/src/index.ts#L215-#L272) script from the big query extension uses a cursor to write documents in batches.

This [example](https://stackoverflow.com/a/60963679/5681607) on Stack Overflow is also pretty good but it is written in Java.

PR #2932, opened 2026-08-04
feat(bigquery-firestore-export): port extension to kits

cabljac:
Full port of the Export BigQuery to Firestore extension (GoogleCloudPlatform/firebase-extensions@68ef3fa, v0.2.2) to kits/, per the migration plan. 13th and final kit scaffolded from the original 13.

## Shape

Follows the firestore-bigquery-export reference pattern: direct v2 wiring with deploy-time param expressions, lazy HandlerContext, requiresRole declarative IAM, side-effect-free ./lib.

- processMessages: v1 pubsub.onPublish -> onMessagePublished; topic from PUBSUB_TOPIC param whose deploy-time default derives from INSTANCE_ID via CEL thenElse, mirroring the runtime fallback (legacy name ext-<instanceId>-processMessages preserved)
- upsertTransferConfig: task queue + all three lifecycle events -> one onTaskDispatched wired to afterFirstDeploy AND afterRedeploy (reconcile is idempotent)
- New: provisioning task creates the notification topic idempotently (Extensions runtime used to auto-create it)
- New: INSTANCE_ID param replaces EXT_INSTANCE_ID; keeps the extInstanceId Firestore tag as the migration key so existing installs are adopted, not duplicated
- DTS serviceAccountName (creation-only) resolved at runtime from the metadata server; omitted when unavailable
- Extensions-runtime glue removed (setProcessingState -> logs; partitioning-field removal terminal, no retry loop)
- Dead transferConfigName link-existing branch not ported (hardcoded undefined upstream at v0.2.2)
- Dropped deps: @invertase/firebase-extension-utilities, lodash.mapvalues

## Tests

64 vitest tests across 6 suites: ported upstream dts/helper/convert/parse coverage plus new config-resolution and provisioning suites (topic idempotency, create vs update path, terminal partitioning error, SA lookup fallback). Firestore document contract (extInstanceId, runs/{runId}, runs/latest, output subcollection) unchanged and pinned by tests. Upstream e2e suite descoped (SPEC risks).

## Deploy verification (pending)

- [ ] DTS run -> topic -> rows at {collection}/{configId}/runs/{runId}/output
- [ ] Unchanged-config upsert re-run is a no-op (empty update mask)
- [ ] Migration path: INSTANCE_ID set to an existing extension instance adopts its transfer config and topic
- [ ] First-deploy topic race check (SPEC risks; pre-create topic if the trigger deploy fails)

### DIFF
--- a/kits/bigquery-firestore-export/.gitignore
+++ b/kits/bigquery-firestore-export/.gitignore
@@ -1,0 +1,2 @@
+lib
+*.tsbuildinfo
--- a/kits/bigquery-firestore-export/README.md
+++ b/kits/bigquery-firestore-export/README.md
@@ -1,0 +1,139 @@
+# @firebase/bigquery-firestore-export
+
+Schedule BigQuery queries and export the results to Firestore. This is the
+Export BigQuery to Firestore Firebase Extension as an npm package you add to
+your own Firebase Functions codebase and deploy.
+
+A BigQuery Data Transfer Service scheduled query runs your SQL on your
+schedule; when a run completes, DTS publishes a notification to a Pub/Sub
+topic, and the `processMessages` function copies the run's result table into
+Firestore. The functions run in your own Firebase project; there is no hosted
+version, so you deploy them yourself.
+
+## Install
+
+```sh
+npm install @firebase/bigquery-firestore-export
+```
+
+## Required IAM and APIs
+
+The package declares the roles below with `requiresRole(...)`. Firebase CLI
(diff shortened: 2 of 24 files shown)


cabljac:
Closing in favour of promoting Corie's earlier port (FirebasePrivate #495, feat/bigquery-firestore-export-extension-apis) - it predates this by two weeks and already removed the factory. Branch kept; the INSTANCE_ID migration key, topic provisioning, metadata-server SA lookup, and the test suites here are worth grafting during promotion.

PR #2940, opened 2026-08-12
Add release workflow

inlined:
I think this has to be in main for it to actually be accessible to run. So much for testing before reviewing =/


.github/workflows/release-kit.yaml
```diff
@@ -0,0 +1,250 @@
...
+  # =========================================================
+  # 1. TEST JOB (Low Permissions: Read-Only)
+  # =========================================================
+  test:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
```
ajperel: Isn't zizmor going to come through and say this must be a SHA and be right?

.github/workflows/release-kit.yaml
```diff
@@ -0,0 +1,250 @@
...
+      - name: Bump Version (Clear CHANGELOG on Stable Release Only)
+        id: versioning
+        working-directory: ${{ inputs.target_kit }}
+        run: |
+          BUMP_TYPE="${{ steps.config.outputs.bump_type }}"
+          IS_PRERELEASE="${{ inputs.is_prerelease }}"
+          DRY_RUN="${{ inputs.dry_run }}"
+
+          if [ "$DRY_RUN" = "true" ]; then
+            npm version $BUMP_TYPE --preid=rc --no-git-tag-version
```
ajperel, line 160: Do I understand correctly that this still updates the version locally just doesn't commit it via git. Should we do even this much or revert it later?

.github/workflows/release-kit.yaml
```diff
@@ -0,0 +1,250 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      target_kit:
+        description: "Kit / Package to publish"
+        required: true
+        type: choice
+        options:
+          - "kits/bigquery-firestore-export"
```
ajperel, line 11: We should coordinate with @cabljac and Invertase. If we move definitive location for some of these kits back to the google cloud extensions repo to be next to the Extension implementations we'll need a similar workflow there.

.github/workflows/release-kit.yaml
inlined: Yeah, this is still a draft because I'm working through stuff. Trying to figure out what I need to do to get this to show up in the actions dashboard for e.g.

.github/workflows/release-kit.yaml
inlined: We do this because npm publish has a dry run as well which verifies auth and other requirements and will be driven by this.

.github/workflows/release-kit.yaml
inlined: Yeah, there's some stuff we can play with like making the branch configurable or making this not prefill. Easy changes that don't block getting started testing this though IMO.

inlined (commented):
Sorry this stayed in draft a while. I had one branch for multiple changes because Gemini lied to me and said I could try to run the action from a branch. Now I'm not entirely sure that's true and I'm trying to split changes across different branches with different base targets which is reasonably annoying.

.github/workflows/release-kit.yaml
ajperel: Ok. I guess it gets cleaned up at some point? Could be worth a comment since it's an interesting semi real thing happening during dry run.

ajperel (approved):

PR #2941, opened 2026-08-13
ci: satisfy zizmor policy in remaining workflows

IzaakGough:
The GitHub Actions Scan check runs zizmor over any workflow file a PR touches, and fails on its mandatory rules (`unpinned-uses`, `template-injection`, and five others) as well as on any medium finding. `validate.yml` was brought in line in #2928; this does the same for the other four, so touching them no longer blocks a PR.

- Every action is pinned to the commit its moving tag resolves to today, with the resolved version in a comment. Nothing changes about what actually runs.
- `test.yml` and `readmes-updated.yml` get `contents: read`; `release.yml` gets `contents: write` because `release.sh` creates releases.
- `readmes-updated.yml` passes the npm prefix step output through `env` rather than interpolating it into the `run` body, which is the `template-injection` finding.

Verified locally with zizmor 1.25.2 and the org config the check downloads: no mandatory and no medium findings remain across all five workflows. Six low `artipacked` findings (checkout without `persist-credentials: false`) are left as they are, since the gate allows low.

`validate.yml` stays on the checkout and setup-node pins it already has, v3.6.0 and v3.8.1 against the v3.7.0 and v3.9.1 pinned here, rather than widening the diff.

### DIFF
--- a/.github/workflows/npm_publish_bq_scripts.yml
+++ b/.github/workflows/npm_publish_bq_scripts.yml
@@ -14,9 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     name: publish_if_newer_version
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
       - name: Setup node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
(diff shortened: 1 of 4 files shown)


inlined:
Sorry I didn't see this. I have another PR to add a release action for kits and it forced me to fix zizmor already so this may not be necessary.

IzaakGough:
no worries, I can close this if not needed

IzaakGough:
Closed as already fixed

PR #2945, opened 2026-08-13
feat(kits): add test script to all packages under kits

inlined:
Adds a test script to every package under kits/. Runs vitest for packages with a tests folder, and prints a warning for packages without tests.


ajperel (approved):
I'm trusting that vitest works.

PR #2947, opened 2026-08-17
refactor(storage-resize-images): use Node built-ins for uuid and mkdir

IzaakGough:
Stacked on #2946. Retarget to `kits` once that merges.

**What was broken:** the kit depends on `mkdirp` and `uuid` for things Node covers natively. The extension removed both, along with `uuidv4` and `rimraf`, in `9bb0c1fa` and `c6fbf3b7`, which landed before the kit was created.

**What changed:** six call sites move to `crypto.randomUUID()` and `fs.promises.mkdir(dir, { recursive: true })`, matching the extension. Both dependencies come out of `package.json`. `mkdirp` leaves the tree entirely; `uuid` remains transitively via genkit and the Google client libraries, so this is dependency hygiene rather than a fix for the `uuid` advisory.

**Verification:** `npm ci` and `npm run build` are clean. Exercised the changed paths directly: `downloadOriginalFile` writes to a uuid-named file under the temp dir with contents intact, two consecutive calls produce distinct paths and neither fails on the already-existing directory, `replaceWithDefaultPlaceholder` swaps in the bundled placeholder, and `constructMetadata` regenerates `firebaseStorageDownloadTokens` as a uuid when `REGENERATE_TOKEN` is on. The full Storage upload path was not exercised.


CorieW (approved):

IzaakGough:
Noting for anyone following this: this PR merged into its base branch, `fix/storage-resize-images-sharp-0.35`, rather than into `kits`. That base had already landed on `kits` via #2946 moments earlier, so these changes went onto a branch that was no longer in use and never reached `kits`.

Relanded in #3184 against `kits`, with the same change.

PR #2948, opened 2026-08-17
docs(kits): document differences from the legacy extensions

IzaakGough:
Every kit's changelog said only "Initial release", and nothing told a user moving off an installed extension instance what would behave differently.

Each kit README gains a `## Differences from the <extension> extension` section, and the changelog entry points at it. Twelve kits are covered. `firestore-incremental-capture` is untouched: this kit is still in the process of being implemented and tested.

Each section covers only what a migrating user acts on: settings renamed or newly defaulted, install-time validation that now fails at runtime, secrets that must exist by name, region and generation changes, data shape changes, and tooling that did not come across. Shared template differences (`extension.yaml` to npm package, params, `HandlerContext`, gen1 to gen2 boilerplate) are deliberately left out.

Verified by reading the extension and kit source for each pair, plus the `extension.yaml` params. Nothing was exercised against a live deploy, and each section says so where a claim depends on one.

Two decisions for a reviewer: whether the section belongs before `## API surface`, where it is now, or higher up in each README; and whether to leave the `firestore-incremental-capture` changelog as it is.

### DIFF
--- a/kits/bigquery-firestore-export/CHANGELOG.md
+++ b/kits/bigquery-firestore-export/CHANGELOG.md
@@ -1,1 +1,1 @@
-- Initial release
+- Initial release of kit, see README for differences between the legacy extension and this kit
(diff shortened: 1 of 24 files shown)


cabljac (approved):
let's get this in so we can iterate from it

PR #2949, opened 2026-08-18
test(delete-user-data): add kit test suite

IzaakGough:
The `delete-user-data` kit had no tests, and `test` was a placeholder that echoes a warning.

Adds two suites. `pnpm test` runs 54 unit tests over the handlers, helpers, publisher and custom search function, using test doubles for the injected clients. `pnpm test:emulator` runs 16 tests against the Firestore, Auth, Pub/Sub and Database emulators, porting the extension's emulator suite: recursive deletes, `hasValidUserPath`, the discovery and deletion loop end to end, the search depth limit, foreign paths being refused, and deleting an auth user clearing its Firestore data.

The emulator suite loads the built kit through a small app under `tests/emulator/app`, which carries its own `package.json` because the Functions emulator rejects the kit's `engines: ">=22"` and wants an exact major. Its `.env` sets every param so the emulator never prompts and caches a stale `.env.local`, and `kits/.gitignore` gains one negation so that file can be committed.

Both suites pass. I mutated the source to check they bite: the ownership check and the search depth limit each fail a test when removed, though depth needs both of its redundant guards removed since either alone still stops the recursion.

One thing to decide: the emulator `.env` has to set `SELECTED_DATABASE_INSTANCE`, because without it `getContext` throws `Can't determine Firebase Database URL` and every invocation dies. That reproduces on a live deploy too, and the extension avoids it by calling `admin.database()` lazily inside the RTDB path rather than eagerly per invocation. Worth its own PR.


IzaakGough:
Closing as already done in https://github.com/firebase/extensions/commit/a3dea2d13377359a5981009478b3923435ad1e10

PR #2952, opened 2026-08-18
test(storage-resize-images): add kit test suite

IzaakGough:
The `storage-resize-images` kit had no tests, and `test` was a placeholder that echoes a warning.

Adds two suites. `pnpm test` runs 101 unit tests covering path matching, `shouldResize`, image conversion, resize, the retry queue, the content filter including its schema-refusal cases, and the resize handler's filter and placeholder routing. `pnpm test:emulator` runs the extension's 4 e2e tests against the Storage and Functions emulators, with the real trigger writing the files the tests assert on.

Two deliberate differences from the extension's unit tests. Its config test snapshots env vars, which does not apply now that config comes from Firebase params, so it is replaced with tests over `configFromEnv`, `validatePathListsFromEnv` and `resolveResizeImagesConfig`. Image assertions use `sharp().metadata()` rather than adding `image-type` and `image-size`.

The emulator suite loads the built kit through a small app under `tests/emulator/app`, which carries its own `package.json` because the Functions emulator rejects the kit's `engines: ">=22"` and wants an exact major. `kits/.gitignore` gains one negation so the app's `.env` can be committed.

Both suites pass. I mutated the source to check they bite: seven mutations each fail at least one unit test, and forcing `shouldResize` to false fails all four e2e tests.

Not ported: the extension's `vulnerability.test.ts` and `content-filter.live.test.ts`, which need a real project and billing.


IzaakGough:
Closing as already done in https://github.com/firebase/extensions/commit/9f67d2fa511c4b0a0c6ca8ecaa5b3b1d6315afc0

PR #2955, opened 2026-08-18
fix(delete-user-data): resolve the RTDB client on first use

IzaakGough:
## What was broken

`getContext` built `ctx.database` with an eager `admin.database()`. With no RTDB instance configured there is no `databaseURL` to initialize the app with, so the call threw `Can't determine Firebase Database URL` before any deletion work, on every invocation of `clearData`, `handleSearch` and `handleDeletion`. A Firestore-only configuration deleted nothing.

Explicit `initializeApp` options also suppress the `FIREBASE_CONFIG` fallback, so the project's default database URL is never picked up.

## What changed

`ctx.database` becomes a getter resolved on first use, matching where the extension calls it. `tests/context.test.ts` asserts the context builds without touching the RTDB client, and that the getter still resolves it when the deletion path reads it.

## Verification

Unit suite 89 passing, `tsc -b` clean, and reverting the getter fails both new tests.

The crash was reproduced on a live 2nd gen deploy of `0.0.2-rc.2`: all three functions threw at `lib/index.js:97` and deleted nothing. The fix was verified against the same project by driving the real `getContext`, not by redeploying.

## For the reviewer

This stops the crash. It does not settle what RTDB should do with no instance configured: skip it as in #2988, or fall back to the project default as the README promises. The two need to agree.

### DIFF
--- a/kits/delete-user-data/src/index.ts
+++ b/kits/delete-user-data/src/index.ts
@@ -74,7 +74,11 @@ function getContext(): HandlerContext {
   ctx = {
     firestore: getFirestore(resolved.firestoreDatabaseId),
     storage: admin.storage(),
-    database: admin.database(),
+    // Resolved on first use. Without a configured RTDB instance there is no
+    // databaseURL to initialize the app with, and admin.database() throws.
+    get database() {
+      return admin.database();
+    },
     pubsub: new PubSub({ projectId: resolved.projectId }),
     config: resolved,
   };
--- a/kits/delete-user-data/tests/context.test.ts
+++ b/kits/delete-user-data/tests/context.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { describe, expect, test, vi } from "vitest";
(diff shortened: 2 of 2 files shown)


cabljac:
The getter fix looks right, and it composes cleanly with the gate from #2988. One thing to fix in the tests:

"resolves the RTDB client when the deletion path reads it" depends on the first test having run: it reads `vi.mocked(handleClear).mock.calls[0]`, which is only populated by the first test's `clearData` call. Run it alone and it fails on the destructure:

    vitest run tests/context.test.ts -t "resolves the RTDB client"

Please make it self-contained: call `clearData(deletionEvent("uid-2"))` inside the test itself, then read the last call. `getContext` caches `ctx` at module level, so `handleClear` receives the same context object and the getter assertion still holds. The transferable rule: a test that reads mock state should be the test that produced it, otherwise `.only`, `-t`, and shuffled ordering all break it.

cabljac (approved):

PR #2976, opened 2026-08-25
Add a README for the kits directory

IzaakGough:
The `kits` directory had no README, so there was nothing explaining what a kit is or how the pieces fit together, and the root README did not mention kits at all.

Adds `kits/README.md`: what a kit is compared with an installed extension, a table of the 13 kits linking to their own READMEs, the `firebase.json` `kit` stanza with instance ids and the `kits` experiment, and a short note on what to check when moving off an installed extension. The root README gets one paragraph linking to it.

Content is drawn from the kit READMEs and `package.json` files in the tree, not from reading each kit's source. Two things a reviewer may want to weigh in on:

- The per-kit READMEs all say `npm install @firebase/<name>`, but every `package.json` declares `@firebase-function-kits/<name>`. Since I could not tell which is intended, the new README has no install command. Worth resolving separately.
- The kit descriptions in the table are the `package.json` descriptions, lightly trimmed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Each directory in this repo contains the source code for the extension and a REA
 
 When you find an extension that solves a need for your app or project, all you do is install and configure the extension. With extensions, you don't spend time researching, writing, and debugging the code that implements functionality or automates a task for your app or project.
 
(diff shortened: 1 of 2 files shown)


cabljac (approved):
lgtm

PR #2982, opened 2026-08-26
fix(bigquery-firestore-export): restore install-time param validation

IzaakGough:
## What was broken

Two params lost their validation in the port to the kit. `COLLECTION_PATH` became a plain string, so an even-segment path such as `a/b` is accepted at deploy and then fails inside Firestore at runtime. `BIGQUERY_DATASET_LOCATION` became free-form, so a typo means the query job location does not match the dataset and the scheduled query fails.

## What changed

`COLLECTION_PATH` carries the upstream regex `^[^/]+(/[^/]+/[^/]+)*$` and its error message. `BIGQUERY_DATASET_LOCATION` is a `select` over the same 35 options as upstream (33 regions plus the US and EU multi-regions), following the existing `LOG_LEVEL` pattern. Resolved config is unchanged for valid input. `params` is now exported so tests can assert the declared inputs.

## Verification

Tests and build run in a worktree holding only this commit on top of `kits`: 36 tests pass, `tsc -b` exits 0. The regex is exercised against `a`, `a/b/c`, `a/b/c/d/e`, `a/b`, `a/b/c/d`, `/a`, `a/` and the empty string.

## Notes

Validation only fires on the interactive prompt. Driving firebase-tools 15.28.1's own `resolveParams`, a value supplied from a `.env` file is taken verbatim, with no regex check and no select-membership check, while the prompt path retries with the error message. So this catches typos during an interactive deploy, not in CI.

The extension behaves the same way, so this is not a gap the kit opened. `validateCommandLineParams` still exists in firebase-tools but has had no caller in the product path since `e16fe0a80` (March 2023), and the extensions deploy planner reads `.env` params without validating them. That was checked in the CLI only; whether the backend rejects a bad value at instance-create time is untested.

One deviation from upstream: the label `EU Mutli-Region (EU)` is spelled `EU Multi-Region (EU)` here.

### DIFF
--- a/kits/bigquery-firestore-export/src/config.ts
+++ b/kits/bigquery-firestore-export/src/config.ts
@@ -27,12 +27,59 @@ import type {
 } from "./export-config";
 
 const LOG_LEVEL_OPTIONS = ["debug", "info", "warn", "error", "silent"] as const;
+
+/**
+ * Mirrors the BigQuery dataset locations offered by the upstream extension. The
+ * query job must run in the same location as the dataset it reads, so a value
+ * outside this list only fails once the scheduled query runs.
+ */
+const BIGQUERY_DATASET_LOCATION_OPTIONS: Record<string, string> = {
+  "Columbus, Ohio (us-east5)": "us-east5",
+  "Iowa (us-central1)": "us-central1",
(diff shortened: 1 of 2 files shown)


IzaakGough:
Closing as done in #2984

PR #2983, opened 2026-08-26
fix(bigquery-firestore-export): stop retrying deploy-time misconfigurations

IzaakGough:
### What was broken

`upsertTransferConfig` retried five times over 2.5 minutes before dead-lettering, even for failures no retry can fix: a missing `TRANSFER_CONFIG_NAME`, a stored config document with no `name`, an unparseable transfer config, a scheduled query deleted out from under a stored document, and any param that fails validation. The logs never said what to change.

### What changed

Those paths throw `PermanentConfigurationError`. `handleUpsertTransferConfig` catches only that type, logs it at error level and returns, so Cloud Tasks sees a 2xx and stops. Everything else still throws and still retries.

### How it was verified

32 unit tests and a clean build. Deployed to a real project, attempts counted in Cloud Logging:

| Case | Baseline `b8cc1113` | Head |
| --- | --- | --- |
| Blank `DATASET_ID` | 5 attempts, HTTP 500 | 1 attempt, HTTP 204 |
| Vanished transfer config | 5 attempts, HTTP 500 | not deployed |
| Happy path | not deployed | run output reached Firestore |

### Notes

Treating a vanished config as permanent matches the client's retry policy: `GetTransferConfig` retries on `DEADLINE_EXCEEDED` and `UNAVAILABLE` only, so a transient blip surfaces as one of those and still retries.

The extension reported these through the Extensions runtime, so the message showed in the console. Kits have lifecycle hooks but no status channel, since `setProcessingState` is scoped to an extension instance via `EXT_INSTANCE_ID`. The message goes to Cloud Logging at error level with the remediation steps in it, so a deploy that stops without acting is easier to miss.

`PARTITIONING_FIELD` is also on the permanent list, but that guard is itself wrong (#2985). Not fixed here.

Tracked in #2974 ([parity analysis](https://github.com/firebase/extensions/issues/2974#issuecomment-5395098640)), rows 7a to 7c.

### DIFF
--- a/kits/bigquery-firestore-export/src/dts.ts
+++ b/kits/bigquery-firestore-export/src/dts.ts
@@ -15,6 +15,7 @@
  */
 
 import * as bigqueryDataTransfer from "@google-cloud/bigquery-data-transfer";
+import { PermanentConfigurationError } from "./errors";
 import type { ResolvedBigqueryFirestoreExportConfig } from "./export-config";
 import * as logs from "./logs";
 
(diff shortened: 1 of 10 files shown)


cabljac:
Superseded by #3086 (merged), which stops retrying the two failures the extension did not fail on. The wider permanent-vs-transient classification is tracked as a follow-up in #3008.

PR #2986, opened 2026-08-26
fix(firestore-genai-chatbot): restore the extension's 540s timeout

CorieW:
## Summary

- Restores the extension's `timeout: 540s` on `generateMessage`, tracked as the kit's highest-priority gap in #2974: the kit set no `timeoutSeconds`, so the function took the 60s default and long generations timed out mid-stream.
- The timeout is declared in `envDeployOptions()` / `DeployTimeOptions` rather than inline, so library consumers get it too and the deploy-options suite asserts it.

## Testing

- New deploy-options assertion: `timeoutSeconds === 540`.
- `tsc --noEmit` clean; all 22 tests pass.
- End-to-end against a consumer project: kit rebuilt from this branch, `npm pack`ed, re-vendored into its function-kits source; discovery endpoint spec carries `timeoutSeconds: 540` (with the `API_KEY` secret binding intact).
- Live deploy of all three genai codebases (`a91f6c2e`, `overrides`, `vertex`) — successful updates, 0 failures.
- Cloud-side verification via `firebase functions:list`: all three deployed `generateMessage` functions report `timeoutSeconds: 540`.

### DIFF
--- a/kits/firestore-genai-chatbot/src/config.ts
+++ b/kits/firestore-genai-chatbot/src/config.ts
@@ -242,5 +242,8 @@ export function envDeployOptions(): DeployTimeOptions {
     // `.env`, the `{messageId}` wildcard stays literal (matches extension.yaml's
     // `${COLLECTION_NAME}/{messageId}` trigger resource).
     document: expr`${params.collectionName}/{messageId}`,
+    // extension.yaml sets `timeout: 540s`; without this the function takes
+    // the 60s default and long generations time out.
+    timeoutSeconds: 540,
   };
 }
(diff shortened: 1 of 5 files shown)


IzaakGough (approved):
Lgtm with one nit:

On the "library consumers get it too" line in the description: the 540 lives in `envDeployOptions()`, which `index.ts` imports but does not re-export, and `lib.ts` does not export it either. So someone following the README's "import from `./lib`" path for their own trigger does not pick it up.

Adding `envDeployOptions` to the `lib.ts` re-exports would cover it.

PR #2988, opened 2026-08-26
fix(delete-user-data): gate RTDB deletion on a configured instance

CorieW:
## Summary

- Restores the extension's RTDB deletion gate, tracked in #2974 as the kit's deletion risk: the extension deletes only when `rtdbPaths && databaseURL`; the kit gated on `rtdbPaths` alone and deleted against whatever default RTDB instance the app resolved.
- `handleClear` now requires `getDatabaseUrl(rtdbInstance, rtdbLocation)` to resolve before scheduling RTDB deletion, logging `rtdbNotConfigured` otherwise — matching the extension's behaviour with `RTDB_PATHS` set and `SELECTED_DATABASE_INSTANCE` empty.
- Existing rtdb deletion tests gain explicit `rtdbInstance` / `rtdbLocation` config; a new test asserts paths are skipped (not deleted) when no instance is configured.

## Testing

- `tsc --noEmit` clean; all 85 tests pass (84 existing + 1 new skip-assertion).
- End-to-end against a consumer project: kit rebuilt from this branch, `npm pack`ed, re-vendored into its function-kits source.
- Runtime gate test against the shipped artifact (the packed `lib`, via the consumer install): `rtdbPaths` set with no instance → zero removals and the "Realtime Database paths are not configured, skipping" log; instance + location configured → deletes the expected path.
- Live deploy of both delete-user-data codebases (`a91f6c2e`, `recursive`) — all 6 functions updated, 0 failures.

### DIFF
--- a/kits/delete-user-data/src/handlers.ts
+++ b/kits/delete-user-data/src/handlers.ts
@@ -19,6 +19,7 @@ import type { DocumentReference } from "firebase-admin/firestore";
 import { FieldPath } from "firebase-admin/firestore";
 import chunk from "lodash.chunk";
 import * as events from "./events";
+import { getDatabaseUrl } from "./export-config";
 import { extractUserPaths, hasValidUserPath } from "./helpers";
 import * as logs from "./logs";
 import { recursiveDelete } from "./recursiveDelete";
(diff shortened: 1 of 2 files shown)


IzaakGough (approved):
lgtm, two nits AI found (non-blocking):

medium — kits/delete-user-data/README.md:170: the section "Realtime Database deletion no longer needs a database instance" documents exactly the behaviour this PR removes ("falling back to your project's default database when no instance is named"). Someone following it (set RTDB_PATHS, leave SELECTED_DATABASE_INSTANCE empty) now silently gets zero RTDB deletions. Delete or rewrite that section; the config table at line 95 could note the instance is required.

low — kits/delete-user-data/src/handlers.ts:160: the else branch reuses logs.rtdbNotConfigured() ("Realtime Database paths are not configured, skipping"), but in the new case the paths are configured and the instance isn't, so the message points at the parameter the operator already set. Worth its own log naming SELECTED_DATABASE_INSTANCE / SELECTED_DATABASE_LOCATION.

Issue #2978, opened 2026-08-25
release-kit workflow: version bump never commits, so re-runs republish the same version (403)


cabljac:
Publishing `rtdb-limit-child-nodes` failed:

> npm error 403 Forbidden - You cannot publish over the previously published versions: 0.0.2-rc.0.

Two bugs in `.github/workflows/release-kit.yaml`:

1. **`npm version` silently skips the git commit/tag.** It only does git operations when `.git` sits next to `package.json`; kits are in `kits/<name>/` with `.git` at repo root. Logs confirm: bump prints only `v0.0.2-rc.0`, then `git push origin kits --follow-tags` says `Everything up-to-date`. No release commit exists anywhere; branch package.json is still `0.0.1`.
2. **Next version comes from package.json only.** Every run therefore computes `0.0.1` -> `0.0.2-rc.0`. First run per kit succeeded (Aug 20-21); every re-run collides, and npm reserves published versions forever (even unpublished).

**Fix**

- Bump with `--no-git-tag-version`, then explicit `git add`/`commit`/`tag` from repo root; fail if no commit created.
- Pick the next free version from the registry (`npm view $PKG time --json` includes unpublished versions), not package.json.
- Optionally push the commit only after successful publish.

Next rtdb prerelease must be `0.0.2-rc.1`, but without the commit fix the collision recurs one run later.

Issue #2977, opened 2026-08-25
Update pinned actions in release-kit workflow (Node 20 runtime deprecation)


cabljac:
The `Release Kit` workflow (`.github/workflows/release-kit.yaml`) emits this warning on runs:

> Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

This is not about the `node-version: "24"` we pass to setup-node. The pinned action versions themselves declare a Node 20 runtime:

- `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683` (v4)
- `actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020` (v4.4.0)

**Fix**

Bump both pins (in the `test` and `release` jobs) to the latest v5 releases, which run on Node 24. Keep full SHA pinning as we do now.

**Notes**

- Other workflows in the repo likely pin the same v4 SHAs and will warn too. Worth sweeping them in the same PR.
- Warning-only for now, but GitHub will remove the Node 20 runtime, at which point these runs fail.

PR #2990, opened 2026-08-27
fix(release-kit): registry-driven versioning, explicit release commits, v7 action pins

cabljac:
Fixes #2978.

- **npm version silently skipped git commit/tag** (.git not next to kit's package.json): bump now uses --no-git-tag-version, with explicit commit + annotated per-kit tag (`<pkg>@<version>`) from repo root. Commit fails loudly if nothing staged.
- **Re-runs recomputed the same version from package.json and 403'd**: next version now derived from the registry (`npm view <pkg> time`, includes unpublished versions) unioned with package.json.
- **Bump level silently ignored when an rc existed**: level now applied to the highest stable version; an rc line continues only if the requested level doesn't open a higher one.
- **npm-shrinkwrap.json bumped but not committed**: now staged in the release commit.
- **Concurrent runs raced the branch push**: workflow-level concurrency group; push is atomic and happens only after a successful publish.

Action pin bumps for #2977 split out into #2991; the two merge independently (non-overlapping hunks).

Verified: dry-run dispatch green, computes 0.0.2-rc.1 for rtdb (the version #2978 requires); version logic passed 10 synthetic registry scenarios locally. The non-dry-run commit/tag/push path has not been exercised live.

### DIFF
--- a/.github/workflows/release-kit.yaml
+++ b/.github/workflows/release-kit.yaml
@@ -41,6 +41,12 @@ on:
         default: true
         type: boolean
 
+# All releases push to the same branch; a concurrent run would fail its push
+# after already publishing to npm, leaving a version with no commit or tag.
+concurrency:
+  group: release-kit
+  cancel-in-progress: false
+
 jobs:
   # =========================================================
   # 1. TEST JOB (Low Permissions: Read-Only)
@@ -125,6 +131,8 @@ jobs:
           # branch can lag behind what was actually released.
           NPM_CONFIG_REGISTRY: https://registry.npmjs.org
         run: |
+          set -o pipefail
+
           PKG_NAME=$(node -p "require('./package.json').name")
           LOCAL_VER=$(node -p "require('./package.json').version")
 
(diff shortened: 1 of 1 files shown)


IzaakGough (commented):
Lgtm - Two things AI flagged (mostly just things to be aware of than fixes required afaik)

**The `time.unpublished.versions` branch cannot be reached**

`npm view <pkg> <field>` resolves the spec before reading the field, and a fully unpublished package has no versions to match, so it errors first and the parser exits at the E404 branch. Verified against `left-pad-test`, a real tombstone:

```
registry packument:  time.unpublished.versions = ["0.0.1"]
npm view … time:     {"error":{"code":"E404","summary":"Unpublished on 2017-01-13…"}}
current parser:      (prints nothing, exits 0)
```

E404 makes "never published" and "fully unpublished" indistinguishable, so the reserved version the code was added to capture is dropped. The scenario needs a 72 hour window or npm support, so the risk is negligible either way; the reason to touch it is that three lines currently read as an active safeguard. Deleting them resolves it as well as fixing them does.

If fixing, reading the packument directly makes the branch reachable and drops the dependency on CLI output shape:

```sh
curl -sf "https://registry.npmjs.org/$PKG_NAME"
```

**Push after publish, for the record**

Moving the push after publish is the better trade. The old ordering could leave a pushed commit and tag for a version that never reached npm, and since versions now come from the registry, npm running ahead of git is the recoverable direction.

The residual: if `kits` moves between the release job's checkout and its push, the push is rejected after publish succeeded, and `--atomic` takes the tag with the branch.

```
! [rejected]  HEAD -> main (fetch first)
! [rejected]  pkg@1.0.0 -> pkg@1.0.0 (atomic push failed)
```

The GitHub Release step is skipped too, so the version sits on npm with no commit, tag or release. The run goes red, and cleanup is a version bump, commit, tag and `gh release create`. The window is checkout to push, under a minute in the run logs, against roughly two pushes a day on `kits`. Optional insurance, if the manual cleanup is worth avoiding:

```sh
git fetch origin "$REF_NAME"
git rebase "origin/$REF_NAME"
```

That does mean the tag points at a tree including `kits/` changes that landed after the build, so it no longer matches the published tarball. Pushing the tag first on its own avoids that, since its name cannot race.

**Optional: pin npm**

npm 12 returns `npm view … time --json` as an array rather than an object, which the parser would read as one key `"0"`, falling back to package.json alone. Not live today, since `node-version: "24"` gives npm 11.19.0, though Node does move npm majors mid-line (v18.14.0, v18.19.0, v20.7.0). `npm i -g npm@11` settles it, and the `curl` above makes it moot.

cabljac:
Nice catches, both reproduced exactly as you described: `npm view left-pad-test time --json` E404s while the raw packument still carries `time.unpublished.versions`, so the unpublished branch really was unreachable, and npm 12's `--json` array shape would have silently emptied the taken-set.

Fixed both in 28813dcb by fetching the packument straight from the registry:

```sh
HTTP_STATUS=$(curl -sS --retry 3 -o "$PACKUMENT" -w '%{http_code}' "$NPM_CONFIG_REGISTRY/$PKG_NAME")
```

404 now means never-published only (tombstones return 200 with `time.unpublished`, verified against left-pad-test), and the parse depends on the registry document shape rather than npm CLI output, so the npm pin becomes unnecessary too.

On the push race: agreed it's the better trade and we're leaving it as is. The fetch+rebase would make the tag point at a tree that no longer matches the published tarball, which is quieter but worse than a loud red run with a short manual cleanup.

IzaakGough (approved):

PR #3002, opened 2026-09-01
fix(storage-resize-images): degrade gracefully against a partial env

CorieW:
## Summary

- Fixes the two remaining cold-start crashes from #2974's runtime-contract table: the extension read `process.env` directly with fallbacks and degraded gracefully against a missing variable; the kit's params layer crashed. Only reachable outside a normal CLI deploy (library consumers of the params entry point, emulator runs, hand-rolled `.env`).
- `IMAGE_TYPE`: `ListParam.value()` JSON-parses the raw var, so an unset value threw `SyntaxError` at cold start. `configFromEnv` now returns `undefined` when absent (resolver falls back to the default image type) and passes non-JSON input through as a raw string, which the resolver's comma-splitting `toArray` handles — an extension-style `jpeg,webp` value now works instead of crashing.
- `FUNCTION_MEMORY`: `IntParam` yields `0` when unset or non-numeric, which `normalizeMemory` passed through as an invalid `"0"` memory option. The `0` sentinel now maps to `undefined`, so the resolver applies its `1GiB` default.
- Not changed (verified already in parity): `SHARP_OPTIONS` — `""` already degrades via `|| "{}"` in the resolver, so that #2974 row is stale; `IS_ANIMATED` / `REGENERATE_TOKEN` — unset reads `false` on both the extension and the kit.
- Tests: two crash-characterization tests superseded by three graceful-degradation tests (unset `IMAGE_TYPE` → `["false"]`, ext-style comma list → `["jpeg","webp"]`, unset `FUNCTION_MEMORY` → `1GiB`); the partial-env characterization test now documents the memory exception.

## Testing

- `tsc --noEmit` clean; 175 tests pass (6 skipped).

### DIFF
--- a/kits/storage-resize-images/src/config.ts
+++ b/kits/storage-resize-images/src/config.ts
@@ -285,6 +285,24 @@ function optional(value: string): string | undefined {
   return value.length > 0 ? value : undefined;
 }
 
+/**
+ * `ListParam.value()` JSON-parses the raw env var, so an unset IMAGE_TYPE (or
+ * an extension-style comma-separated list) throws at cold start. The extension
+ * read `process.env.IMAGE_TYPE` directly and degraded gracefully; do the same,
+ * handing non-JSON input to the resolver's comma-splitting `toArray`.
+ */
+function imageTypesFromEnv(): ReadonlyArray<string> | string | undefined {
+  const raw = process.env.IMAGE_TYPE;
+  if (raw === undefined) {
+    return undefined;
+  }
+  try {
+    return params.imageTypes.value();
+  } catch {
+    return raw;
+  }
+}
+
 export function configFromEnv(): ResizeImagesConfig {
   return {
     bucket: params.bucket.value(),
@@ -296,11 +314,14 @@ export function configFromEnv(): ResizeImagesConfig {
     excludePathList: optional(params.excludePathList.value()),
     failedImagesPath: optional(params.failedImagesPath.value()),
     cacheControlHeader: optional(params.cacheControlHeader.value()),
-    imageTypes: params.imageTypes.value(),
+    imageTypes: imageTypesFromEnv(),
     outputOptions: optional(params.outputOptions.value()),
     sharpOptions: params.sharpOptions.value(),
     isAnimated: params.isAnimated.value(),
-    memory: params.memory.value(),
+    // IntParam yields 0 when FUNCTION_MEMORY is unset or non-numeric; the
+    // extension always supplied a value, so treat that as unset and let the
+    // resolver fall back to the default memory.
+    memory: params.memory.value() || undefined,
     regenerateToken: params.regenerateToken.value(),
     contentFilterLevel:
       params.contentFilterLevel.value() as ResizeImagesConfig["contentFilterLevel"],
--- a/kits/storage-resize-images/src/config.ts
+++ b/kits/storage-resize-images/src/config.ts
@@ -286,10 +286,11 @@ function optional(value: string): string | undefined {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


kits/storage-resize-images/src/config.ts
```diff
@@ -285,6 +285,24 @@ function optional(value: string): string | undefined {
...
+ * read `process.env.IMAGE_TYPE` directly and degraded gracefully; do the same,
+ * handing non-JSON input to the resolver's comma-splitting `toArray`.
+ */
+function imageTypesFromEnv(): ReadonlyArray<string> | string | undefined {
+  const raw = process.env.IMAGE_TYPE;
+  if (raw === undefined) {
+    return undefined;
+  }
+  try {
+    return params.imageTypes.value();
```
IzaakGough: `IMAGE_TYPE=false` slips past this fallback, and that is the legacy extension's default value.

`extension.yaml` declares `IMAGE_TYPE` with `default: false` and maps the `original` option to value `false`, so a migrated `.env` most often carries exactly `IMAGE_TYPE=false`. That parses as JSON fine (to boolean `false`), so nothing throws and the `catch` never runs. `ListParam.runtimeValue()` then returns `[]` for a parsed value that isn't a `string[]`, and in the resolver `toArray([]) ?? [DEFAULT_IMAGE_TYPE]` keeps `[]` because an empty array isn't nullish. Config ends up with `imageTypes: []`.

Downstream, `resizeImages` queues zero tasks, and since #3037 `handleObjectFinalized` treats zero resize outputs as a failed run: every upload logs `logs.failed()` and is copied to `failedImagesPath`, while a success event still goes out with empty outputs.

Verified by driving the real `firebase-functions/params` `defineList("IMAGE_TYPE", { default: ["false"] })` through this helper plus `toArray`:

```
"false"          -> []          (no throw, so the catch is bypassed)
"[\"false\"]"    -> ["false"]    correct
"original"       -> SyntaxError  caught, falls back to raw
"jpeg,webp"      -> SyntaxError  caught, falls back to raw
```

`null`, `0`, `[]` and `["jpeg",5]` land on `[]` the same way.

To be clear this isn't introduced here: the previous `params.imageTypes.value()` returned `[]` for `false` too. But the fallback is shaped around throwing input, and the most common extension-style value doesn't throw, so it stays broken. Something like falling back to `raw` whenever the parsed list is empty but `raw` isn't would cover it:

```ts
const parsed = params.imageTypes.value();
return parsed.length > 0 ? parsed : raw;
```

Happy for this to be scoped out and tracked separately if you'd rather keep the PR tight, just flagging that `false` is the case most installs will hit.

IzaakGough:
Related to the `IMAGE_TYPE` handling in this PR, and outside the diff so noting it here: the `original` option is declared as `"False"` with a capital F, while the param default just above is `["false"]`, `DEFAULT_IMAGE_TYPE` in `export-config.ts` is `"false"`, and `resize-image.ts` decides whether to convert with a case-sensitive `format !== "false"`.

So a `"False"` value takes the convert branch instead of the keep-original branch. `SUPPORTED_IMAGE_CONTENT_TYPE_MAP["False"]` is undefined, and the resized file gets a `.False` extension.

This is the same shape as #3064, which maps the `"False"` content-filter option to `"OFF"`. Lining the two up in one pass might be easier than fixing them separately.

Two things I did not verify: what the installer actually writes to `.env` when `original` is selected, and what sharp does with `toFormat("False")`. The mismatch between the option value, the declared default, and the comparison in `resize-image.ts` is from reading the code.

IzaakGough (approved):
lgtm!

PR #3004, opened 2026-09-01
fix(delete-user-data): default the auto discovery search depth

IzaakGough:
## What was broken

`defineInt` resolves a missing or blank env var to `0` (`parseInt(env || "0", 10) || 0`), so the declared `default: 3` never reached runtime, and `resolveDeleteUserDataConfig` could not recover it because `0` is not nullish.

With `ENABLE_AUTO_DISCOVERY=true` and the depth left blank, `handleClear` still published a search message for every root collection, but `handleSearch` returned immediately on `depth > searchDepth` (`1 > 0`) for every one. Discovery deleted nothing and still logged `Successfully removed data for user`. The extension read a blank value as 3.

## What changed

`configFromEnv` reports an absent, blank or whitespace-only value as `undefined` so the resolver applies the documented default. An explicit `0` is a real setting the extension honoured, so it is preserved rather than folded into "unset".

New cases live in `config-runtime.test.ts` because `config.test.ts` fakes `firebase-functions/params` and cannot see how the real `IntParam` resolves a missing variable.

## Verification

Suite 92 passing, `tsc -b` clean. Resolved values: `"  5  "` and `"5"` give 5, `"0"` gives 0, `"   "` and `""` give 3. Reverting the fix fails the unset and blank cases, treating `0` as unset fails the explicit zero, dropping the trim fails whitespace-only.

Checked end to end against live Firestore with the depth blank and discovery on. Before: resolved `0`, nothing deleted. After: resolved `3`, documents at depth 1 and 2 deleted.

### DIFF
--- a/kits/delete-user-data/src/config.ts
+++ b/kits/delete-user-data/src/config.ts
@@ -19,6 +19,7 @@ import {
   defineInt,
   defineString,
   expr,
+  type IntParam,
   projectID,
   select,
   storageBucket,
@@ -26,8 +27,6 @@ import {
 import type { DeleteUserDataConfig } from "./export-config";
 
 const instanceId = defineString("INSTANCE_ID");
-
-const SEARCH_DEPTH_PARAM = "AUTO_DISCOVERY_SEARCH_DEPTH";
 
 const params = {
   instanceId,
@@ -122,7 +121,7 @@ const params = {
 
     default: false,
   }),
-  searchDepth: defineInt(SEARCH_DEPTH_PARAM, {
+  searchDepth: defineInt("AUTO_DISCOVERY_SEARCH_DEPTH", {
     label: "Auto discovery search depth",
     description:
       "If auto discovery is enabled, how deep should auto discovery find collections and documents. For example, setting to `1` would only discover root collections and documents, whereas setting to `9` would search sub-collections 9 levels deep. Defaults to `3`.",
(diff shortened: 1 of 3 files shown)


kits/delete-user-data/src/config.ts
```diff
@@ -27,6 +27,8 @@ import type { DeleteUserDataConfig } from "./export-config";
 
 const instanceId = defineString("INSTANCE_ID");
```
cabljac: pre-existing but instanceId should be FIREBASE_KIT_INSTANCE_ID i think

kits/delete-user-data/src/config.ts
```diff
@@ -27,6 +27,8 @@ import type { DeleteUserDataConfig } from "./export-config";
...
 
+const SEARCH_DEPTH_PARAM = "AUTO_DISCOVERY_SEARCH_DEPTH";
```
cabljac: how come this one gets a const but the others don't? is there a reason?

IzaakGough: No good reason. I will change optionalInt to just take the param and we can get rid of this const.

kits/delete-user-data/src/config.ts
IzaakGough: @cabljac is it worth changing replacing all INSTANCE_ID across the kits in a separate PR, or should I just fix this one here?

CorieW (approved):

PR #3005, opened 2026-09-01
test(kits/firestore-incremental-capture): pin the array element wire format

cabljac:
The pipeline-side fix for GoogleCloudPlatform/firebase-extensions#1147 (GoogleCloudPlatform/firebase-extensions#1150) tells an array element that is a tagged value from one that is a bare field map by requiring the tag to be a string. These tests pin the element encoding that rule reads: tagged scalars, a null element tagged `null` rather than the legacy `object`, binary, reference and timestamp elements, nested arrays tagged `array`, and a user map element whose own keys are `type` and `value`, which stays unambiguous because its `type` serializes to an object.

Tests only, no behavior change. The kit still pins pipeline release v0.1.0, which lacks the fix, so the known issue documented in the README stands until that pin moves to a release containing it.

### DIFF
--- a/kits/firestore-incremental-capture/tests/wire-format.test.ts
+++ b/kits/firestore-incremental-capture/tests/wire-format.test.ts
@@ -195,6 +195,76 @@ describe("changelog wire format", () => {
     });
   });
 
+  test("tags a null array element as 'null'", () => {
+    // Not `{ type: "object" }`: the original extension tagged null elements
+    // with `typeof null`, which the pipeline cannot tell from a map.
+    expect(serializeDocument({ arrayValue: [null] })).toEqual({
+      arrayValue: {
+        type: "array",
+        value: [{ type: "null", value: null }],
+      },
+    });
+  });
+
+  test("tags a Buffer, DocumentReference or Timestamp array element", () => {
+    const timestampValue = Timestamp.fromDate(
+      new Date("2026-01-02T03:04:05.000Z")
+    );
+
+    expect(
+      serializeDocument({
+        arrayValue: [Buffer.from("hi"), ref, timestampValue],
+      })
+    ).toEqual({
+      arrayValue: {
+        type: "array",
+        value: [
+          { type: "binary", value: "aGk=" },
+          { type: "reference", value: "products/abc" },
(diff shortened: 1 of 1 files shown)


IzaakGough (approved):

kits/firestore-incremental-capture/tests/wire-format.test.ts
```diff
@@ -195,6 +195,76 @@ describe("changelog wire format", () => {
...
+    // with `typeof null`, which the pipeline cannot tell from a map.
+    expect(serializeDocument({ arrayValue: [null] })).toEqual({
+      arrayValue: {
+        type: "array",
+        value: [{ type: "null", value: null }],
+      },
+    });
+  });
+
+  test("tags a Buffer, DocumentReference or Timestamp array element", () => {
```
IzaakGough, line 209: Non-blocking: this covers Buffer, DocumentReference and Timestamp, which leaves GeoPoint as the only type in `serializeArrayElement`'s exclusion list with no element-level test. It would fit in this assertion alongside the other three.

It does look load-bearing. Dropping `!(element instanceof GeoPoint) &&` from the guard keeps all 15 wire-format tests and the whole kit suite green, while a GeoPoint sitting directly in an array changes from `{ type: "geopoint", value: { latitude: ..., longitude: ... } }` to a bare field map. Under the `isTaggedValue` rule that bare shape reads as a field map, so it would restore as a two-key map rather than a GeoPoint. The pipeline side pins the case already as `geopointArrayElement`.

IzaakGough (commented):
Ran this branch locally: the 15 wire-format tests and the full 71-test kit suite pass, and each new expectation matches what `src/serializer.ts` emits. The `type`/`value` map element case is the valuable one here, and it lines up with the `isTaggedValue` rule in the pipeline fix.

Two non-blocking notes. One inline, about a missing GeoPoint element case. The other is documentation: the file header says the format is pinned against pipeline release v0.1.0, but these new array element assertions pin what the fix in GoogleCloudPlatform/firebase-extensions#1150 reads, and v0.1.0 predates it. A line saying so would stop a later reader taking v0.1.0 as the reference for the array element tags.

Issue #3007, opened 2026-09-01
fix(storage-resize-images): failed resize can delete the original image


cabljac:
An empty `imageTypes` produces zero outputs, which counts as success, so the original is deleted under `deleteOriginal: on_success`. The extension never deleted unless a resize actually succeeded.

Parity ledger: #2974, storage-resize-images §6.

PR #3037, opened 2026-09-01
fix(storage-resize-images): failed resize can delete the original image

cabljac:
In the kit, an empty `imageTypes` (or empty `sizes`) yields zero resize tasks, `resizeResults.some(...)` is false, and the run counts as success, so under `deleteOriginal: on_success` the original is deleted with nothing produced. The legacy extension avoided this through param validation (`IMAGE_TYPE` is `required: true` in extension.yaml, unset `IMG_SIZES` crashed at module load), not in code: an unset `IMAGE_TYPE` reaching its handler would hit the same zero-task deletion, and an empty-string type actually resizes successfully. Kits have no such install-time gate, so the handler must guard.

The fix treats zero resize outputs as a failed run in `generateResizedImageHandler`: `logs.failed()` fires, `handleFailedImage` runs (a no-op without `failedImagesPath`), and deletion only happens when at least one resize actually succeeded. No config-time validation was added, keeping clear of #3002's config changes.

Two new vitest cases pin the behavior; the kit's full suite passes (179 passed, 3 skipped). No live Storage E2E was run.

Fixes #3007

### DIFF
--- a/kits/storage-resize-images/src/handlers.ts
+++ b/kits/storage-resize-images/src/handlers.ts
@@ -153,10 +153,12 @@ export async function generateResizedImageHandler(
         },
       });
 
-      resizeFailed = resizeResults.some(
-        (result) =>
-          result.status === "rejected" || result.value.success === false
-      );
+      resizeFailed =
+        resizeResults.length === 0 ||
+        resizeResults.some(
+          (result) =>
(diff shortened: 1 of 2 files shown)


CorieW (approved):
lgtm

Issue #3010, opened 2026-09-01
fix(firestore-vector-search): queryOnWrite retriggers itself


cabljac:
The handler writes back to the document it watches with no status/idempotency guard, where the extension had one. Unbounded invocation loop, billing risk.

Parity ledger: #2974, firestore-vector-search §7.

PR #3038, opened 2026-09-01
fix(firestore-vector-search): stop queryOnWrite retriggering itself

cabljac:
handleQueryOnWrite ran a vector query and wrote the result back onto the same document it watches, re-firing itself in a self-sustaining loop, billed one embedding call plus one vector query per pass (parity ledger #2974, firestore-vector-search section 7).

An earlier revision guarded on the event's before/after snapshots, but that races: if a slow run overwrites a newer query's result, the echo sees unchanged inputs plus a result and skips, stranding query B's inputs with query A's result permanently. The guard now uses stored state instead: the result write also stores the request (query, limit, prefilters) that produced it in the status field, and a trigger skips only when the document's current inputs deep-match that stored record. A stale overwrite mismatches, re-runs on the current inputs, and the corrective echo matches and stops. This is in the spirit of the legacy extension's FirestoreOnWriteProcessor (which the kit cannot consume, being firebase-functions v1-tied) but deliberately stronger: the legacy processor never re-ran a completed query document at all, even when the query changed, and did not track prefilters. The kit re-runs on any input edit, matching its embed-path rule; the README documents the new query-doc status shape and this divergence.

Vitest pins the guard, the re-run paths, a legacy doc without a record, and the full race sequence; guard mutations fail 5-11 tests depending on the mutation, including an asymmetric write-path mutation that would loop forever. Kit suite: 65 passed. Unit tests only, not verified against live Firestore.

Fixes #3010


kits/firestore-vector-search/src/handlers.ts
```diff
@@ -125,6 +126,26 @@ export async function handleQueryOnWrite(
...
+        query: before.query,
+        limit: before.limit,
+        prefilters: before.prefilters,
+      },
+      { query, limit: data.limit, prefilters: data.prefilters }
+    ) &&
+    data.result
+  ) {
+    return;
+  }
```
CorieW: Is this to improve parity, or to fix an existing issue?

AI returned the following:

There is a race if the document’s query changes while an earlier query is still running. The result is calculated from the event’s original data in `handlers.ts:L150-L157`, but `handlers.ts:L159` writes it to the current document without checking whether the query has changed.

For example, query A starts, the document changes to query B, and B finishes first. If A finishes afterward, the write at `L159` can overwrite B’s result with A’s older result.

The new guard at `handlers.ts:L135-L148` changes what happens next. Previously, A’s stale write would trigger another query using the document’s current B inputs, although that also contributed to the loop being fixed here. With the new guard, the follow-up invocation sees unchanged B inputs and an existing result, so it returns. This can leave the document permanently containing query B with query A’s result.

IzaakGough (approved):

CorieW:
### Test report

Verified the kit unit suite and a live deploy of the branch.

**Unit tests** — `npx vitest run` in `kits/firestore-vector-search`: **59/59 pass**.

Regression-proved the guard by reverting `src/handlers.ts` to the base and re-running `tests/handlers.test.ts`: exactly the 4 new guard cases go red, the other 16 stay green.

```
× skips the query when the query is unchanged and a result exists
× skips the first echo, where only the result was added
× skips when query, limit, and prefilters are all unchanged
× skips when the unchanged limit is NaN
Tests  4 failed | 16 passed (20)
```

`isDeepStrictEqual` is the right choice for the NaN case — it uses SameValue, so `NaN === NaN` holds, and Firestore value objects (Timestamp, DocumentReference) inside `prefilters` compare correctly because they share a prototype and only carry own data properties.

**Live deploy** — repacked the kit into a local functions-kit project and deployed instance `firestore-vector-search-a91f6c2e` to a real project. All 9 functions created, `afterFirstDeploy` hook queued, `Deploy complete!`.

### One problem: merge conflict with #3065

Both this branch and #3065 append a new `describe` block (and the same `handleQueryOnWrite` / `VectorWriteEvent` imports) to `kits/firestore-vector-search/tests/handlers.test.ts`. `git merge-tree` against the merge base reports **4 conflict hunks, all in that test file**:

```
kits/firestore-vector-search/tests/handlers.test.ts
+<<<<<<< .our   (x4)
```

`src/handlers.ts` is also touched by both but auto-merges cleanly and the merged result is semantically correct (this PR's echo guard sits above `performTextQuery`, #3065's `parsePrefilters` replaces an argument inside it). So only the test file needs a manual resolution — whichever of the two lands second.

### Not verified

Could not smoke test the deployed `queryOnWrite` behaviour: the Application Default Credentials on this machine need reauth (`invalid_grant` / `invalid_rapt`), so nothing could write a query document to trigger the function. The echo-loop behaviour is therefore only confirmed at unit level, not against a live Firestore write.

kits/firestore-vector-search/src/handlers.ts
cabljac: Good catch, this race was real: the result is computed from the event's data but written to the current doc, so a slow run A could overwrite B's result, and the before/after guard would then see unchanged inputs plus a result and skip forever.

Reworked the guard to the legacy extension's FirestoreOnWriteProcessor mechanism (reimplemented here since the helper is tied to firebase-functions v1): the result write also stores the request (query, limit, prefilters) that produced it under the status field, and a trigger skips only when the doc's current inputs deep-match that stored record. A's stale write stores A's request while the doc holds B's inputs, so the echo mismatches, re-runs on B's inputs, and the corrective write's echo matches and stops. The full sequence is pinned in the test "a stale overwrite from a slow concurrent run self-heals, then stops" in tests/handlers.test.ts.

cabljac: One correction to my earlier reply: calling this the extension's FirestoreOnWriteProcessor mechanism reimplemented overstated the equivalence. The legacy processor skipped unconditionally once status reached a final state - a completed query document never re-ran, even when the query changed - and it tracked only query and limit. The kit's guard stores the request that produced the result and re-runs on any input mismatch, so it is deliberately stronger: it heals the race you found AND re-runs completed docs on input edits, matching the kit's embed-path rule. README now documents the query-doc status shape and this divergence.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -135,7 +155,13 @@ export async function handleQueryOnWrite(
     config: ctx.config,
   });
 
-  await event.data.after.ref.set(result, { merge: true });
+  await event.data.after.ref.set(
+    {
+      ...result,
+      [ctx.config.statusFieldName]: { state: "COMPLETED", request },
```
CorieW, line 161: Small problem, which would really be the user causing their own problems:

> The status field can replace important data
> 
> The name of the status field is configurable. It normally uses `"status"`, but users can choose another name.
> 
> If someone configures it as `"result"`, the function effectively writes:
> 
> ```
> {
>   result: { ids: ["product-1"] },
>   result: { state: "COMPLETED" }
> }
> ```
> 
> The second value replaces the first, so the actual search result disappears.
> 
> The same problem occurs if it is configured as `"query"`, `"limit"`, or `"prefilters"`. The code should prevent those names from being used as the status field.

kits/firestore-vector-search/README.md
```diff
@@ -251,9 +251,22 @@ longer written, only `COMPLETED` and `ERROR`. Anything reading
...
-`status.textQuery`, so if you were waiting on that to know a query had finished,
-wait for `result` instead.
+Query documents carry the flat shape too, with a `request` record alongside the
+state:
+
+```
+status: { state: "COMPLETED", request: { query, limit, prefilters } }
+```
+
+They previously carried `status.textQuery`, so anything reading that path needs
```
CorieW, line 261: AI-flagged minor README issue:
> An existing result may belong to the previous query
> 
> Suppose the document already contains:
> 
> ```ts
> {
>   query: "red shoes",
>   result: { ids: ["red-shoe-1"] }
> }
> ```
> 
> The user changes the query to `"blue shoes"`. While the new search is running, the document temporarily contains:
> 
> ```ts
> {
>   query: "blue shoes",
>   result: { ids: ["red-shoe-1"] } // still the old result
> }
> ```
> 
> The README says consumers can wait for `result`, but `result` already exists. A consumer could therefore treat the old red-shoe result as the completed blue-shoe result.
> 
> The documentation should explain that the query is complete only when `status.request` matches the current query inputs.

kits/firestore-vector-search/tests/handlers.test.ts
```diff
@@ -70,6 +75,28 @@ function request(data: unknown, auth: unknown = { uid: "test-user" }) {
...
+  return {
+    exists: data !== undefined,
+    data: () => data,
+    get: (field: string) => data?.[field],
+    ref: { set, path: "test-collection/doc-1" },
+  };
+}
+
+/** A write event over the same document, with a spy on the after ref's set. */
+function writeEvent(
```
CorieW, line 91: AI-flagged minor issue:
> The test cannot tell which snapshot was written to
> 
> Firestore events contain two snapshots:
> 
> * `before`: the document before the change
> * `after`: the document after the change
> 
> The handler should write through `after.ref`. However, the test gives both snapshots the same mock write function.
> 
> That means this incorrect code:
> 
> ```ts
> event.data.before.ref.set(...)
> ```
> 
> would look identical to the correct code:
> 
> ```ts
> event.data.after.ref.set(...)
> ```
> 
> The production code is currently correct; the problem is that the test would not catch a future regression.

kits/firestore-vector-search/src/handlers.ts
cabljac: Valid, and worth guarding even though it takes a misconfiguration. With statusFieldName "result" the status record wins over the spread result inside the same set payload, so the search result silently disappears; with "query", "limit", or "prefilters" the merge write clobbers the user's input field. resolveVectorSearchConfig now throws for those four names, so both the kit params path and direct consumers fail at load time instead of losing data (1746d6ab, cases in tests/export-config.test.ts). A collision with inputFieldName or outputFieldName on embedding documents is the same class of misconfiguration but exists in the legacy extension too, so I left that out of this PR.

kits/firestore-vector-search/README.md
cabljac: Agreed, the "waiting on result also works" line was wrong for exactly this window. Fixed in 6ae1db5f: the README now warns that result alone can still hold the previous query's result while a re-run is in flight, and defines completion as status.state COMPLETED with status.request matching the document's current query, limit, and prefilters. While in there I also documented that the legacy extension wrote query-document status to the literal "status" field regardless of STATUS_FIELD_NAME (its query processor never received the param), whereas the kit honors the param on query documents too, and listed the reserved names from the other thread.

kits/firestore-vector-search/tests/handlers.test.ts
cabljac: Fixed in 034c1d02: writeEvent now gives each snapshot its own set spy (set for after, beforeSet for before) and the create test asserts beforeSet is never called. Mutation-checked by rerouting the handler's write through before.ref: with the shared spy that mutant passed the whole suite, with the split spies it fails 8 tests. Also added a test in the same file pinning that the request record is stored and matched under a non-default statusFieldName.

cabljac:
Thanks for the test report and the live deploy check, Corie. Addressed your three inline comments in 1746d6ab, 034c1d02, and 6ae1db5f: reserved statusFieldName values now throw at config resolution, the README defines completion as status.state COMPLETED with a matching status.request instead of "waiting on result also works", and the test harness now spies before.ref and after.ref separately. Suite is now 72/72 (was 59 when you ran it), tsc clean. On the #3065 overlap: still confined to tests/handlers.test.ts, now 5 conflict hunks instead of 4 (the writeEvent helper change adds one); src/handlers.ts still auto-merges. Whichever PR lands second resolves that file manually.

Issue #3036, opened 2026-09-01
chore(kits): minor parity cleanups


cabljac:
Batch of one-liners from #2974, one commit each or per kit:

- firestore-bigquery-export: `logs.start()` called with no argument (§4) - #3039 (merged)
- firestore-vector-search: custom-endpoint content-type check dropped (§4c) - #3042 (open)
- firestore-send-email: `formatZodError` `invalid_string` branch dropped - #3040 (open)
- storage-resize-images: `BLOCK_NONE` accepted where the extension threw (§4); `cacheControlHeader: ""` set literally (§7) - #3044 (merged)
- firestore-genai-chatbot: local `SafetySetting` type (no enum check); `PROJECT_ID` error text - #3046 (open)
- bigquery-firestore-export: dropped log helpers - #3045 (open)
- firestore-bigquery-export: `MAX_STALENESS` doc example 8h to 4h (the original #2974 ledger line mislabeled the kit; the param lives in firestore-bigquery-export)
- delete-user-data: `node-fetch@2` where the extension used global `fetch` - #3041 (merged)
- firestore-translate-text: dead backfill remnants (`filterLanguagesFn`, unused log helpers) - #3043 (merged)

cabljac:
All nine items merged: #3039, #3040, #3041, #3042, #3043, #3044, #3045 (log helpers + MAX_STALENESS doc example), #3046.

PR #3041, opened 2026-09-01
chore(delete-user-data): replace node-fetch with the Node global fetch

cabljac:
Drops the node-fetch@2 dependency from the delete-user-data kit in favor of the Node built-in global fetch (the kit's engine is Node >=22), matching the legacy extension. Removes the import, the node-fetch and @types/node-fetch entries from package.json, and regenerates npm-shrinkwrap.json via npm (removals only); the tests that mocked the node-fetch module now stub the global with vi.stubGlobal. The call site only uses response.ok/text()/json(), so no node-fetch-specific behavior (streams, agents, FetchError) needed handling. Two known semantic deltas, both benign here: a network failure now rejects with a TypeError instead of node-fetch's FetchError (nothing inspects the error class), and undici rejects a searchFunction URL with embedded credentials up front with a TypeError where node-fetch would have sent the request (spec-correct, and not a supported configuration). Rebased onto kits after #3060; full suite is green (98 tests), tsc and prettier clean. Part of #3036.

### DIFF
--- a/kits/delete-user-data/npm-shrinkwrap.json
+++ b/kits/delete-user-data/npm-shrinkwrap.json
@@ -12,12 +12,10 @@
         "@google-cloud/pubsub": "^4.3.3",
         "firebase-admin": "^13.2.0",
         "firebase-functions": "^7.3.2",
-        "lodash.chunk": "^4.2.0",
-        "node-fetch": "^2.6.2"
+        "lodash.chunk": "^4.2.0"
       },
       "devDependencies": {
         "@types/lodash.chunk": "^4.2.7",
-        "@types/node-fetch": "^2.6.2",
         "typescript": "^5.9.3",
         "vitest": "^4.1.10"
       },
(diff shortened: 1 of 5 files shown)


IzaakGough (approved):
lgtm

cabljac:
Re the review feedback on the module-level vi.stubGlobal calls: leaving as is.

1. Hoisting: imports do evaluate before the stub runs, but no module in this kit references fetch at module init. The only reference is inside runCustomSearchFunction's body, resolved at call time, so the stub is always in place before any fetch happens.
2. Cross-file pollution: the kit runs vitest with default config, so each test file gets an isolated worker context; a module-level stub cannot leak into other test files.

Given both concerns don't apply here, and the module-level stub sits in the same slot the vi.mock("node-fetch") call did, keeping the diff to one line per file seemed preferable to adding lifecycle hooks (also keeps churn away from the test areas #2955 touches).

PR #3042, opened 2026-09-01
fix(firestore-vector-search): reject non-JSON responses from the custom embeddings endpoint

cabljac:
The legacy extension (firestore-vector-search v0.1.3) checked the custom embeddings endpoint's content-type header and rejected non-JSON responses with a clear error; the kit dropped that check, so a non-JSON response (for example an HTML error page) surfaced as a raw `Unexpected token` parse error. This restores the check in `CustomEndpointClient.getEmbeddings`, throwing "Error getting embeddings from custom endpoint: response is not JSON" before parsing, matching the legacy wording. Following review, it also restores two more legacy behaviors in the same method: the schema parse is wrapped so valid JSON of the wrong shape throws "Error getting embeddings from custom endpoint: response does not match expected schema" instead of a raw ZodError, and the length-mismatch message reverts to legacy's exact "response does not contain embeddings" wording. Adds four vitest cases (JSON success with charset parameter, non-JSON content-type, wrong-shape JSON, missing embeddings) - each failing case was written and verified red before its fix. Full kit suite passes (53 tests) and `tsc -b` is clean; note `npx prettier --check src` already fails on origin/kits because prettier 2.7.1 cannot parse `satisfies` in `src/config.ts` (pre-existing, untouched). Part of #3036.

### DIFF
--- a/kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
+++ b/kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
@@ -57,12 +57,18 @@ export class CustomEndpointClient extends BaseEmbedClient {
     }
 
     const data = await response.json();
-    const parsed = z
-      .object({ embeddings: z.array(z.array(z.number())) })
-      .parse(data);
+    const schema = z.object({ embeddings: z.array(z.array(z.number())) });
+    let parsed: z.infer<typeof schema>;
+    try {
+      parsed = schema.parse(data);
+    } catch {
+      throw new Error(
+        "Error getting embeddings from custom endpoint: response does not match expected schema"
+      );
+    }
     if (parsed.embeddings.length !== inputs.length) {
       throw new Error(
-        "Error getting embeddings from custom endpoint: response does not contain embeddings for all inputs"
+        "Error getting embeddings from custom endpoint: response does not contain embeddings"
       );
     }
     return parsed.embeddings;
--- a/kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
+++ b/kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
@@ -50,6 +50,12 @@ export class CustomEndpointClient extends BaseEmbedClient {
       );
     }
 
+    if (!response.headers.get("content-type")?.includes("application/json")) {
+      throw new Error(
+        "Error getting embeddings from custom endpoint: response is not JSON"
+      );
+    }
+
     const data = await response.json();
     const parsed = z
       .object({ embeddings: z.array(z.array(z.number())) })
--- a/kits/firestore-vector-search/tests/embeddings.test.ts
+++ b/kits/firestore-vector-search/tests/embeddings.test.ts
@@ -206,6 +206,40 @@ describe("CustomEndpointClient", () => {
     await expect(client.getEmbeddings(["input"])).resolves.toEqual([[1, 2, 3]]);
   });
 
+  test("throws a clear error when the JSON does not match the expected schema", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(
+        async () =>
+          new Response(JSON.stringify({ foo: "bar" }), {
+            headers: { "content-type": "application/json" },
+          })
+      )
+    );
+
+    const client = new CustomEndpointClient(customConfig());
+    await expect(client.getEmbeddings(["input"])).rejects.toThrow(
+      "Error getting embeddings from custom endpoint: response does not match expected schema"
+    );
+  });
+
+  test("throws a clear error when embeddings are missing for some inputs", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(
+        async () =>
+          new Response(JSON.stringify({ embeddings: [[1, 2, 3]] }), {
+            headers: { "content-type": "application/json" },
+          })
+      )
+    );
+
+    const client = new CustomEndpointClient(customConfig());
+    await expect(client.getEmbeddings(["input1", "input2"])).rejects.toThrow(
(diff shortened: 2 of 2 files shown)


IzaakGough (commented):
The guard is placed where the legacy one was, with the same message and the same optional-chain semantics, so this is a faithful port. Two optional notes on the condition and on the parse just below it.

kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
```diff
@@ -50,6 +50,12 @@ export class CustomEndpointClient extends BaseEmbedClient {
       );
     }
 
+    if (!response.headers.get("content-type")?.includes("application/json")) {
```
IzaakGough, line 53: Two small things about this condition, both optional given it matches legacy exactly.

MIME types are case-insensitive per RFC 9110, so `Content-Type: Application/JSON` is a legal response that this rejects. A `toLowerCase()` before the `includes` covers it.

The other is that an absent `content-type` lands in the same branch as `text/html`. A handler doing a bare `res.end(JSON.stringify(...))` sends no `content-type` and returns perfectly good JSON, and it would now fail with "response is not JSON", which points at the wrong thing. Requiring the header to be present and wrong keeps that case working:

```ts
const contentType = response.headers.get("content-type")?.toLowerCase();
if (contentType && !contentType.includes("application/json")) {
```

That does give up the guard for a no-header endpoint returning HTML, which then falls through to `response.json()` and a raw `SyntaxError`, so it is a trade rather than a strict win. Fine to leave as is if legacy parity is the goal.

kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
```diff
@@ -50,6 +50,12 @@ export class CustomEndpointClient extends BaseEmbedClient {
...
+      throw new Error(
+        "Error getting embeddings from custom endpoint: response is not JSON"
+      );
+    }
+
     const data = await response.json();
     const parsed = z
```
IzaakGough: One line below, this is the same class of raw-error leak the PR is fixing: a JSON response with the wrong shape surfaces a bare `ZodError`. Legacy wrapped the parse in a try/catch and threw "Error getting embeddings from custom endpoint: response does not match expected schema". Might be worth folding in while you are here.

IzaakGough (approved):
Lgtm - AI flagged two points inline, most likely not worth addressing since this PR is exactly in parity with the legacy extension

CorieW:

**Unit tests** — `npx vitest run` in `kits/firestore-vector-search`: **53/53 pass**, including all four new `CustomEndpointClient` cases:

```
✓ CustomEndpointClient > returns embeddings from a JSON response
✓ CustomEndpointClient > throws a clear error when the JSON does not match the expected schema
✓ CustomEndpointClient > throws a clear error when embeddings are missing for some inputs
✓ CustomEndpointClient > throws a clear error when the response is not JSON
```

Regression-proved all three fixes by reverting `src/embeddings/client/text/custom_function.ts` to the base:

```
× throws a clear error when the JSON does not match the expected schema
× throws a clear error when embeddings are missing for some inputs
× throws a clear error when the response is not JSON
Tests  3 failed | 11 passed (14)

Expected: "Error getting embeddings from custom endpoint: response is not JSON"
Received: "Unexpected token '<', \"<html>ok</html>\" is not valid JSON"
```

That `Received` line is exactly the raw-parse-error symptom described in the PR body, so the check lands where intended. Using `.includes("application/json")` rather than an equality check is the right call — it keeps `application/json; charset=utf-8` working, and the suite covers that case.

**Live deploy** — repacked the kit into a local functions-kit project and deployed all three vector-search instances (`-a91f6c2e`, `-euclid`, `-openai`) to a real project:

```
24 x "Successful create operation."
+  Deploy complete!
```

`-openai` and `-euclid` both exercise a non-Gemini embedding path, so the changed module is on the live code path for those instances.

**Not verified:** no live call against a real custom endpoint. Application Default Credentials on this machine need reauth (`invalid_grant` / `invalid_rapt`), so I could not write documents to drive the deployed functions — the content-type behaviour is confirmed at unit level only.

No problems found.

kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
```diff
@@ -50,6 +50,12 @@ export class CustomEndpointClient extends BaseEmbedClient {
       );
     }
 
+    if (!response.headers.get("content-type")?.includes("application/json")) {
+      throw new Error(
+        "Error getting embeddings from custom endpoint: response is not JSON"
+      );
+    }
```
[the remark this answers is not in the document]
cabljac: Leaving as-is for legacy parity: the kit's check matches the legacy v0.1.3 condition exactly, and this PR's scope is restoring that behavior, not extending it. A lowercase content-type is what every mainstream server sends; broadening the match is a behavior change better made in both places deliberately.

kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
cabljac: Agreed on both points, and leaving the condition as-is per your note - the PR's goal is parity with the legacy check, and the absent-header case is a trade rather than a strict win. If we want to broaden the match later it should be its own change with tests for the no-header case.

kits/firestore-vector-search/src/embeddings/client/text/custom_function.ts
cabljac: Applied in cc7e04cd: the schema parse is wrapped and wrong-shape JSON now throws "Error getting embeddings from custom endpoint: response does not match expected schema", with a vitest case pinning it.

Issue #3048, opened 2026-09-01
test(delete-user-data): config tests expect INSTANCE_ID but the code declares FIREBASE_KIT_INSTANCE_ID


cabljac:
Two tests in `kits/delete-user-data/tests/config.test.ts` fail on a clean `kits` checkout ("declares the params the extension exposes", "defaults the topic names to kit-{instanceId}-* expressions"): they expect a param named `INSTANCE_ID` while the code declares `FIREBASE_KIT_INSTANCE_ID`. The param was renamed without updating the tests, so the kit's suite is red on the main branch. Reproduced independently on base `eb244557` during the #3041 review.

Fix: update the test expectations (or the param name, if the rename was unintended - check which side is right against the other kits' conventions).

cabljac:
Fixed by #3060 (merged 2026-09-01). delete-user-data suite is green on kits (101/101).

PR #3051, opened 2026-09-01
fix(delete-user-data): restore the INSTANCE_ID param name

cabljac:
The two failing tests in kits/delete-user-data/tests/config.test.ts pin the correct behavior: every other kit declares this param as INSTANCE_ID, the kit's README documents INSTANCE_ID, and firebase-tools reserves the FIREBASE_ env prefix (RESERVED_PREFIXES in lib/functions/env.js), so a param named FIREBASE_KIT_INSTANCE_ID cannot be set through .env files at all. The rename landed as a lone drive-by commit (03dd40b4) on the unrelated #3004 branch without updating tests, README, or the other kits, so this reverts the code side rather than the tests. Full suite is green: 95 tests passing, tsc -b clean. Fixes #3048

### DIFF
--- a/kits/delete-user-data/src/config.ts
+++ b/kits/delete-user-data/src/config.ts
@@ -26,7 +26,7 @@ import {
 } from "firebase-functions/params";
 import type { DeleteUserDataConfig } from "./export-config";
 
-const instanceId = defineString("FIREBASE_KIT_INSTANCE_ID");
+const instanceId = defineString("INSTANCE_ID");
 
 const params = {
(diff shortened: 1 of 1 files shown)


cabljac:
Reworked in #3060: the revert direction was wrong. firebase-tools >= 15.27.0 injects FIREBASE_KIT_INSTANCE_ID for kit instances, but as a reserved env var the params machinery can never satisfy, so the correct fix is a direct process.env read rather than either param name.

PR #3054, opened 2026-09-01
test(firestore-counter): port the active legacy emulator tests

cabljac:
Issue #2974 flags the kit as having no emulator tests. The kit's FakeFirestore suites already cover more controller and worker logic than the legacy extension ever did, so this is a narrow port rather than a new test layer.

The legacy `controller.emulator.test.ts` had four tests, two of them `test.skip`'d upstream. Both active ones are ported to `tests/integration/controller.emulator.test.ts` and run against a real Firestore emulator: internal-state document creation on first run, and resharding. Resharding goes through `rescheduleWorkers` rather than the static `balanceWorkers` helper - the pure function is already covered by the fake, so the emulator only earns its keep by exercising the surrounding transaction and worker-document writes. `worker.emulator.test.ts` is 100% skipped upstream and is not ported. `e2e.test.ts` needs a deployed functions environment and is out of scope.

The suite skips cleanly when `FIRESTORE_EMULATOR_HOST` is unset. Verified locally against the emulator: 79 passed, 12 files. Both assertions were checked to fail when the expected values are altered. Test-only, no src changes.

Part of #2974.

### DIFF
--- a/kits/firestore-counter/tests/integration/controller.emulator.test.ts
+++ b/kits/firestore-counter/tests/integration/controller.emulator.test.ts
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Controller tests against a real Firestore emulator. The rest of the suite
+ * runs on FakeFirestore, so these cover what a fake cannot: real transaction
+ * read/write ordering and server-resolved timestamps.
+ *
+ * They are skipped unless FIRESTORE_EMULATOR_HOST is set.
+ *
+ * Run locally from kits/firestore-counter:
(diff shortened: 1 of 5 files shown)


kits/firestore-counter/tests/integration/controller.emulator.test.ts
```diff
@@ -0,0 +1,150 @@
...
+} from "firebase-admin/firestore";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+
+import {
+  ControllerStatus,
+  ShardedCounterController,
+} from "../../src/controller";
+
+const runEmulator = !!process.env.FIRESTORE_EMULATOR_HOST;
+const describeEmulator = runEmulator ? describe : describe.skip;
```
IzaakGough, line 44: This guard means the file never executes in the kit's CI job. `.github/workflows/test-kits.yml` runs `npm ci && npm run build && npm test` with no Firestore emulator and no firebase-tools installed, so the green check on this PR is `describe.skip`, not a pass. Locally it all works: `npx vitest run` gives 77 passed / 2 skipped with no emulator, and the `emulators:exec` command in the header comment gives 2 passed.

The legacy equivalents do run in CI. `firestore-counter/functions/__tests__/controller.emulator.test.ts` has no guard, it sets `FIRESTORE_EMULATOR_HOST` to `localhost:8080` directly (matching the firestore port in `_emulator/firebase.json`), and root `jest --listTests` picks it up, so `npm run test:ci` (`firebase emulators:exec ... jest -P demo-test`) exercises it on every PR.

So a regression in `aggregateOnce`'s first-run `t.set(EMPTY_CONTROLLER_DATA)` branch, or in `rescheduleWorkers`' `serverTimestamp()` writes, would merge green here while failing on the legacy side. Could the kit test job get a `firebase emulators:exec --only firestore` step?

IzaakGough (approved):

PR #3056, opened 2026-09-01
test(bigquery-firestore-export): port legacy unit coverage into the kit

cabljac:
The kit carried 19 unit tests against roughly 1,330 lines of unit coverage in the legacy `bigquery-firestore-export` extension, and `ensureNotificationTopic` - code with no legacy counterpart - had none at all. This ports the missing coverage: 63 tests now, all in four new files so nothing existing is edited and no src changes are made.

`notification-topic.test.ts` drives `ensureNotificationTopic` through `handleUpsertTransferConfig` and pins the concurrent-create race (gRPC ALREADY_EXISTS swallowed, every other error rethrown before the transfer config is touched). `dts-transfer-config.test.ts` backfills the legacy update-mask change-detection cases and the get/create/update wrappers. `run-results.test.ts` replaces the legacy emulator setup with an in-memory Firestore fake against the injected `HandlerContext`, pinning row counts, partial write failures, non-success runs, latest-document ordering and Pub/Sub redelivery. `helper-values.test.ts` ports the granular value-conversion and name-parsing cases.

Deliberately skipped: the live reconfiguration e2e suite and the `xdescribe`'d `e2e/functions.test.ts`, neither of which runs without live GCP resources. Assertions follow the kit's current behavior, not the legacy soft-failure paths.

Part of #2974.

### DIFF
--- a/kits/bigquery-firestore-export/tests/dts-transfer-config.test.ts
+++ b/kits/bigquery-firestore-export/tests/dts-transfer-config.test.ts
@@ -74,6 +74,18 @@ function unchangedTransferConfig(): TransferConfig {
   } as TransferConfig;
 }
 
+/**
+ * The stored config with only `delta` applied, so a change case can assert the
+ * whole updated config and catch edits the update mask does not mention.
+ */
+function transferConfigWithDelta(
+  delta: (transferConfig: TransferConfig) => void
+): TransferConfig {
+  const expected = unchangedTransferConfig();
+  delta(expected);
+  return expected;
+}
+
 function clientReturning(transferConfig: TransferConfig | null) {
   return {
     getTransferConfig: vi.fn().mockResolvedValue([transferConfig]),
@@ -128,10 +140,11 @@ describe("constructUpdateTransferConfigRequest change detection", () => {
     );
 
     expect(request.updateMask?.paths).toEqual(["schedule"]);
-    expect(request.transferConfig?.schedule).toBe("every 15 minutes");
-    expect(
-      request.transferConfig?.params?.fields?.partitioning_field?.stringValue
-    ).toBe("");
+    expect(request.transferConfig).toEqual(
+      transferConfigWithDelta((expected) => {
+        expected.schedule = "every 15 minutes";
+      })
+    );
   });
 
   test("masks params when the destination table name changed", async () => {
@@ -142,10 +155,12 @@ describe("constructUpdateTransferConfigRequest change detection", () => {
     );
 
     expect(request.updateMask?.paths).toEqual(["params"]);
-    expect(
-      request.transferConfig?.params?.fields?.destination_table_name_template
-        ?.stringValue
-    ).toBe('different_table_{run_time|"%H%M%S"}');
+    expect(request.transferConfig).toEqual(
+      transferConfigWithDelta((expected) => {
+        expected.params.fields.destination_table_name_template.stringValue =
+          'different_table_{run_time|"%H%M%S"}';
+      })
+    );
   });
 
   test("masks params when the query changed", async () => {
@@ -156,8 +171,11 @@ describe("constructUpdateTransferConfigRequest change detection", () => {
     );
 
     expect(request.updateMask?.paths).toEqual(["params"]);
-    expect(request.transferConfig?.params?.fields?.query?.stringValue).toBe(
-      "SELECT * FROM source.accounts"
+    expect(request.transferConfig).toEqual(
+      transferConfigWithDelta((expected) => {
+        expected.params.fields.query.stringValue =
+          "SELECT * FROM source.accounts";
+      })
     );
   });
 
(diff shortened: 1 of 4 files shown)


kits/bigquery-firestore-export/tests/helper-values.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    expect(convertUnsupportedDataTypes(123)).toBe(123);
+    expect(convertUnsupportedDataTypes(true)).toBe(true);
+  });
+
+  test("converts every BigQuery temporal type to a Firestore Timestamp", () => {
+    const converted = convertUnsupportedDataTypes({
+      timestamp: new BigQueryTimestamp("2023-01-15T10:30:00Z"),
+      date: new BigQueryDate("2023-01-15"),
+      datetime: new BigQueryDatetime("2023-01-15T10:30:00"),
+      time: new BigQueryTime("1970-01-01T10:30:00Z"),
```
IzaakGough: `BigQueryTime` never holds an ISO datetime string in practice, so this case asserts a path that cannot happen. A `TIME` column is decoded through `BigQuery.time(value)`, and the constructor stores the raw API string verbatim, so `.value` is `"10:30:00"`.

With the real value the conversion throws rather than returning a `Timestamp`:

```
> new Date("10:30:00")
Invalid Date
> Timestamp.fromDate(new Date("10:30:00"))
Error: Value for argument "seconds" is not a valid integer.
```

That matters because the conversion is called synchronously inside the write loop rather than inside a settled promise, so it throws straight out of `writeRunResultsToFirestore` and `handleTransferRunMessage`. One `TIME` column fails the whole run's export and Pub/Sub keeps redelivering.

https://github.com/firebase/extensions/blob/87cf77620168bedc091346d52427b56ced9ff6cb/kits/bigquery-firestore-export/src/helper.ts#L183-L188

Switching the fixture to `new BigQueryTime("10:30:00")` turns this test red, which looks like the right outcome to surface before the kit ships.

kits/bigquery-firestore-export/tests/helper-values.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    });
+
+    expect(converted.timestamp).toBeInstanceOf(Timestamp);
+    expect(converted.date).toBeInstanceOf(Timestamp);
+    expect(converted.datetime).toBeInstanceOf(Timestamp);
```
IzaakGough: Smaller version of the same thing. `datetime` only gets `toBeInstanceOf`, and `"2023-01-15T10:30:00"` is the real `DATETIME` shape, so the value assertion is worth adding here too.

A date-time literal with no offset is parsed as local time, so `convertUnsupportedDataTypes` shifts `DATETIME` values by whatever offset the function runs in. Only `timestamp` gets a value assertion (line 105), so this passes on every machine and the shift stays unpinned.

kits/bigquery-firestore-export/tests/dts-transfer-config.test.ts
```diff
@@ -0,0 +1,359 @@
...
+    const client = clientReturning(unchangedTransferConfig());
+    client.updateTransferConfig.mockResolvedValue([updated]);
+
+    const result = await updateTransferConfig(client, TRANSFER_CONFIG_NAME, {
+      ...config,
+      schedule: "every 15 minutes",
+    });
+
+    expect(result).toEqual(updated);
+    expect(client.updateTransferConfig).toHaveBeenCalledOnce();
```
IzaakGough: `toHaveBeenCalledOnce()` does not look at the request, so a regression that drops or widens the update mask still passes here. Asserting the object handed to `client.updateTransferConfig` would pin it.

Worth covering the no-change case through this wrapper as well. The empty mask is only asserted at the builder level, and `UpdateTransferConfigRequest.fromObject` turns `paths: []` into an empty `FieldMask`:

```
fromObject({transferConfig: {...}, updateMask: {paths: []}}).toJSON()
// {"transferConfig":{...},"updateMask":{}}
```

`handleUpsertTransferConfig` calls `updateTransferConfig` on every upsert, so a redeploy with no config change sends that empty mask. I have not exercised it against the live DTS API, so I am only flagging the missing coverage rather than claiming the API rejects it.

kits/bigquery-firestore-export/tests/run-results.test.ts
```diff
@@ -0,0 +1,373 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { BigQuery } from "@google-cloud/bigquery";
+import type { Firestore } from "firebase-admin/firestore";
```
IzaakGough: `Firestore` is not used in this file. The fake is typed through `as unknown as ResultHandlerContext`, so this import can go.

Nothing catches it today because `tsconfig.json` excludes `**/*.test.ts` and the kit has no linter.

IzaakGough (commented):
### Code review

Four comments, all on the new test files. The first one is the substantive one: the `BigQueryTime` fixture is a value shape the BigQuery client never produces, so that assertion passes over a real crash in `convertUnsupportedDataTypes`. The other three are smaller coverage notes.

I ran the suite locally to check the port: 10 files, 63 tests, all passing.

PR #3040, opened 2026-09-01
fix(firestore-send-email): restore the invalid_string branch in formatZodError

cabljac:
The kit's formatZodError dropped the invalid_string branch the legacy extension has, so string-validation failures (e.g. a failed .email() check) fell through to zod's raw default message instead of the extension's "Field 'x' must be a valid email address" / "Field 'x' is invalid" shape. This restores the branch; the kit pins zod ^3.25.76 (same major as the extension's ^3.24.4), so the issue code and its validation field carry over unchanged. formatZodError is now exported so the new vitest tests can drive it with real zod errors; note the kit's current schemas do not yet use string validators, so this path is only reachable if a schema adds one, matching the extension's defensive intent. Full kit suite, tsc, and prettier pass; no live/e2e run.

Part of #3036.

Issue #3009, opened 2026-09-01
fix(firestore-send-email): SendGrid sends fail with AUTH_TYPE=OAuth2


cabljac:
Secret gating forces `apiKey: undefined` so `setApiKey` is never called and every send fails for this supported config combination.

Parity ledger: #2974, firestore-send-email.

IzaakGough:
Closed by #3089

PR #3057, opened 2026-09-01
test(firestore-send-email): port the legacy unit suites into the kit

cabljac:
The `firestore-send-email` kit shipped with 14 tests against roughly 1,900 lines of source, while the legacy extension it was ported from carries around 2,000 lines of unit tests. This PR ports the four pure unit suites that transfer without an emulator: `validatePayload` (new `tests/validation.test.ts`), `setSmtpCredentials` / `isSendGrid` (new `tests/helpers.test.ts`), `SendGridTransport` (new `tests/nodemailer-sendgrid.test.ts`), and template merging appended to the existing `tests/prepare-payload.test.ts`. Jest goes to vitest, `it` to `test`, and the legacy `setDependencies` mocking is replaced by the kit's `PreparePayloadDependencies` seam. The kit test count goes from 14 to 115, all green, `tsc -b` clean, prettier clean. Only test files change; no source or manifest is touched.

Three tests are new rather than ported, each mutation-verified by breaking the branch it covers and confirming it was the only failure. Two pin guards the kit's transport refactor added and the legacy transport lacked: an empty `watchHtml` emits no content entry, an empty `messageId` emits no header. The third covers the template merge falling back to the message's own attachments when the render supplies none.

Deliberately not ported: everything gated on the `TESTING` env flag (the kit has no `process.env.TESTING`), the `describe.skip`'d SMTP e2e, and the `E2E_SENDGRID` live suite. Also dropped are the legacy helper tests' `regex.test(...)` assertions, which only exercised a regex literal declared inside the test file rather than the kit's own `SMTP_CONNECTION_URI` `validationRegex`.

Two notes. #3040 also adds `tests/validation.test.ts` with `formatZodError` tests; the two files do not overlap in coverage but will need a trivial merge, whichever lands second. Its `invalid_string` branch is also the one real difference between the kit's `validation.ts` and the legacy file, which are otherwise logic-identical on every reachable path. And no test here touches the SendGrid OAuth2 secret gating in #3009, so nothing pins that bug.

Part of #2974.

### DIFF
--- a/kits/firestore-send-email/tests/config.test.ts
+++ b/kits/firestore-send-email/tests/config.test.ts
@@ -16,10 +16,22 @@
 
 import { describe, expect, test, vi } from "vitest";
 
+interface StringParamOpts {
+  default?: string;
+  input?: { text?: { validationRegex?: RegExp } };
+}
+
+const { stringParamOpts } = vi.hoisted(() => ({
+  stringParamOpts: new Map<string, StringParamOpts | undefined>(),
+}));
+
 vi.mock("firebase-functions/params", () => ({
-  defineString: (_name: string, opts?: { default?: string }) => ({
-    value: () => opts?.default ?? "",
-  }),
+  defineString: (name: string, opts?: StringParamOpts) => {
+    stringParamOpts.set(name, opts);
+    return {
+      value: () => opts?.default ?? "",
+    };
+  },
   defineInt: (_name: string, opts?: { default?: number }) => ({
     value: () => opts?.default ?? 0,
   }),
@@ -87,6 +99,43 @@ describe("secretParamsForAuthType", () => {
     expect(() => secretParamsForAuthType("Unknown")).toThrow(
       "Unsupported AUTH_TYPE for firestore-send-email: Unknown"
     );
+  });
+});
+
+describe("SMTP_CONNECTION_URI validationRegex", () => {
+  function connectionUriRegex(): RegExp {
+    const regex = stringParamOpts.get("SMTP_CONNECTION_URI")?.input?.text
+      ?.validationRegex;
+    if (!regex) {
+      throw new Error("SMTP_CONNECTION_URI declares no validationRegex");
+    }
+    return regex;
+  }
+
+  test("accepts the documented URI forms and a blank value", () => {
+    for (const uri of [
+      "smtps://username@smtp.hostname.com:465",
+      "smtps://smtp.gmail.com:465",
+      "smtps://username@gmail.com:password@smtp.gmail.com:465",
+      "smtp://smtp.gmail.com:587?pool=true",
+      "",
+    ]) {
+      expect(connectionUriRegex().test(uri)).toBe(true);
+    }
+  });
+
+  test("rejects a URI without a scheme or without a port", () => {
+    expect(connectionUriRegex().test("smtp.gmail.com:465")).toBe(false);
+    expect(connectionUriRegex().test("smtp://smtp.gmail.com")).toBe(false);
+  });
+
+  test("accepts trailing garbage after a valid prefix because the first alternative is unanchored", () => {
+    for (const uri of [
+      "smtp://fakeemail@gmail.com:4,h?dhuNTbv9zMrP4&7&7%*3:smtp.gmail.com:465?pool=true&service=gmail",
+      "smtps://smtp.gmail.com:465 and then total garbage",
+    ]) {
+      expect(connectionUriRegex().test(uri)).toBe(true);
+    }
   });
 });
 
--- a/kits/firestore-send-email/tests/config.test.ts
+++ b/kits/firestore-send-email/tests/config.test.ts
@@ -129,6 +129,7 @@ describe("SMTP_CONNECTION_URI validationRegex", () => {
     expect(connectionUriRegex().test("smtp://smtp.gmail.com")).toBe(false);
   });
 
+  // Inherited from the legacy extension.yaml regex; anchoring is tracked in #3067.
   test("accepts trailing garbage after a valid prefix because the first alternative is unanchored", () => {
     for (const uri of [
       "smtp://fakeemail@gmail.com:4,h?dhuNTbv9zMrP4&7&7%*3:smtp.gmail.com:465?pool=true&service=gmail",
--- a/kits/firestore-send-email/tests/helpers.test.ts
+++ b/kits/firestore-send-email/tests/helpers.test.ts
@@ -1,0 +1,239 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { logger } from "firebase-functions";
+import Mail from "nodemailer/lib/mailer";
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import type { ResolvedSendEmailConfig } from "../src/export-config";
+import { isSendGrid, setSmtpCredentials } from "../src/helpers";
+import { AuthenticatonType } from "../src/types";
+
+const warnSpy = vi.spyOn(logger, "warn").mockImplementation(() => undefined);
+
+function makeConfig(
+  overrides: Partial<ResolvedSendEmailConfig>
+): ResolvedSendEmailConfig {
+  return {
+    databaseId: "(default)",
+    databaseRegion: "us-central1",
+    mailCollection: "mail",
+    defaultFrom: "",
+    testing: false,
+    ttlExpireType: "never",
+    ttlExpireValue: 1,
+    tlsOptions: "{}",
+    oauthSecure: true,
+    authType: AuthenticatonType.UsernamePassword,
+    ...overrides,
+  };
+}
+
+describe("setSmtpCredentials", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  test("prefers the smtpPassword secret over the password in the URI", () => {
+    const config = makeConfig({
+      smtpConnectionUri:
+        "smtps://fakeemail@gmail.com:secret-password@smtp.gmail.com:465",
+      smtpPassword: "fakepassword",
+    });
+
+    const credentials = setSmtpCredentials(config);
+
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.port).toBe(465);
+    expect(credentials.options.host).toBe("smtp.gmail.com");
+    expect(credentials.options.auth.pass).toBe("fakepassword");
+    expect(credentials.options.secure).toBe(true);
+  });
+
+  test("falls back to the password embedded in the URI", () => {
+    const config = makeConfig({
+      smtpConnectionUri:
+        "smtps://fakeemail@gmail.com:secret-password@smtp.gmail.com:465",
+    });
+
+    const credentials = setSmtpCredentials(config);
+
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.port).toBe(465);
+    expect(credentials.options.host).toBe("smtp.gmail.com");
+    expect(credentials.options.auth.user).toBe("fakeemail@gmail.com");
+    expect(credentials.options.auth.pass).toBe("secret-password");
+    expect(credentials.options.secure).toBe(true);
+  });
+
+  test("accepts a URI with a username but no password", () => {
+    const config = makeConfig({
+      smtpConnectionUri: "smtps://fakeemail@gmail.com@smtp.gmail.com:465",
+    });
+
+    const credentials = setSmtpCredentials(config);
+
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.port).toBe(465);
+    expect(credentials.options.host).toBe("smtp.gmail.com");
+    expect(credentials.options.auth.user).toBe("fakeemail@gmail.com");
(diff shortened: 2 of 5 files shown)


kits/firestore-send-email/tests/helpers.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    expect(
+      isSendGrid(
+        makeConfig({
+          smtpConnectionUri: "smtps://apikey@fake-sendgrid.net:465",
+        })
+      )
+    ).toBe(false);
+  });
+
+  test("returns false when no URI is configured", () => {
```
IzaakGough, line 252: Worth a second look on dropping the `regex.test(...)` assertions, because those two cases are the ones the kit's own regex now gets wrong.

`src/config.ts` defines the `SMTP_CONNECTION_URI` `validationRegex` as `/^(smtp[s]*:\/\/(.*?(:[^:@]*)?@)?[^:@]+:[0-9]+(\?[^ ]*)?)|^$/`. Adding `|^$` for the blank case took the place of the trailing `$`, and since `|` binds loosest the first alternative is no longer anchored at the end, so it matches on a prefix:

```
"smtp://fakeemail@gmail.com:4,h?dh...:smtp.gmail.com:465?pool=true"  legacy=false  kit=true
"smtps://smtp.gmail.com:465 and then total garbage"                  legacy=false  kit=true
"smtp://host:25junk"                                                 legacy=false  kit=true
```

The first is exactly what the legacy `invalid separator` test pinned. `...(\?[^ ]*)?)$|^$/` restores the anchor. That is a source change rather than something for a test-only PR, so a follow-up seems like the right home for it, but it does mean the regex ends up with no coverage anywhere.

kits/firestore-send-email/tests/helpers.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        "smtp://fakeemail@gmail.com:4,h?dhuNTbv9zMrP4&7&7%*3@smtp.gmail.com:465?pool=true&service=gmail",
+    });
+
+    expect(() => setSmtpCredentials(config)).toThrow(Error);
+    expect(warnSpy).toHaveBeenCalledWith(
+      "Invalid URI: please reconfigure with a valid SMTP connection URI"
+    );
+  });
+
+  test("builds an OAuth2 transport when the auth type is OAuth2", () => {
```
IzaakGough, line 193: `checkMicrosoftServer` is the one branch of `setSmtpCredentials` left uncovered. An outlook, office365, or hotmail hostname returns a different transport entirely (`service: "hotmail"`, decoded user and pass, no tls options), and nothing here reaches it. The legacy suite did not cover it either, so this is inherited rather than introduced, but it is the largest gap left in a file this PR otherwise pins thoroughly.

kits/firestore-send-email/tests/helpers.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    expect(credentials.options.port).toBe(465);
+    expect(credentials.options.host).toBe("smtp.gmail.com");
+    expect(credentials.options.auth.user).toBe("fakeemail@gmail.com");
+    expect(credentials.options.auth.pass).toBe("secret-password");
+    expect(credentials.options.secure).toBe(false);
+    expect(credentials.options.pool).toBe(true);
+    expect(credentials.options.service).toBe("gmail");
+  });
+
+  test("percent-encodes special characters in the smtpPassword secret", () => {
```
IzaakGough: Small naming point: the assertion reads the decoded `auth.pass`, so what this pins is the encode/decode round trip rather than the encoding. Something like "round-trips special characters in the smtpPassword secret" would match what it checks.

kits/firestore-send-email/tests/prepare-payload.test.ts
```diff
@@ -96,3 +98,346 @@ describe("preparePayload", () => {
...
+    const result = await prepare({
+      to: "test@example.com",
+      template: { name: "html-only-template", data: {} },
+      message: { attachments: [{ filename: "original.doc" }] },
+    });
+
+    expect(result.message.attachments).toEqual([{ filename: "original.doc" }]);
+  });
+
+  test("a template that renders nothing leaves the message intact", async () => {
```
IzaakGough: The message here has no `attachments`, and the assertion says it comes back with `[]`, so the merge does change the message. The assertion is the valuable part and worth keeping as-is. The name is the bit that overpromises, maybe "a template that renders nothing only adds an empty attachments array".

IzaakGough (commented):
Checked this out and ran it: 115 pass, prettier clean. I re-ran the mutation check on the three new tests (the `watchHtml` guard, the `messageId` guard, and the `templateRender.attachments` fallback) and each break produced exactly one failure, the intended one. The port lines up too: 13 of the 15 legacy helper tests plus all 19 transport tests, and the two dropped ones are the `regex.test(...)` pair as described.

One note on the verification: `tsconfig.json` has `include: ["src"]` and excludes `**/*.test.ts`, so `tsc -b` never reads these files, and vitest transpiles without typechecking. I typechecked them separately with `tests` added to `include` and they come out clean, so the claim holds, just not by that route.

Nothing blocking. Four inline notes below, the first is the only one with real substance.

CorieW:
### Test report — no problems found in the ported suites

This is a test-only change, so I verified it by running and then deliberately breaking the code under test.

**Suite runs green** — `npx vitest run` in `kits/firestore-send-email`: **119/119 pass across 8 files** (up from 16 on the base).

**The ported tests are not vacuous.** Mutation check — inverted the return values of `isSendGrid` in `src/helpers.ts`:

```ts
return !(url.hostname === "smtp.sendgrid.net");
...
return !(false);
```

```
Tests  4 failed | 115 passed (119)
```

So the ported cases do bind to real behaviour rather than passing regardless.

### One problem: merge conflict with #3040

This PR adds `kits/firestore-send-email/tests/validation.test.ts` as a **new file** (599 lines, `describe("validatePayload")`). #3040 adds a file at the **same path** as a new file too (47 lines, `describe("formatZodError")`). `git merge-tree` against the merge base reports it as `added in both` with a whole-file conflict:

```
added in both
  our    100644 4f0b2f44... kits/firestore-send-email/tests/validation.test.ts
  their  100644 53fe1a1b... kits/firestore-send-email/tests/validation.test.ts
```

The two files' contents are disjoint — this PR's suite has no `formatZodError` coverage at all — so the resolution is a concatenation, not a choice. Whichever lands second will need it done by hand.

Everything else here looks good.

kits/firestore-send-email/tests/nodemailer-sendgrid.test.ts
```diff
@@ -0,0 +1,437 @@
...
+}
+
+/** Yields to the microtask queue so the transport's normalize/send chain settles. */
+function flush(): Promise<void> {
+  return new Promise((resolve) => setImmediate(resolve));
+}
+
+function sentMessage(): any {
+  return send.mock.calls[0][0] as any;
+}
```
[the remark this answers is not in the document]
cabljac: Applied in 38bfe418: sentMessage now uses send.mock.calls[0]?.[0], so an assertion on an unsent message fails with an expect diff instead of a TypeError inside the helper. Suite still 119/119.

kits/firestore-send-email/tests/nodemailer-sendgrid.test.ts
```diff
@@ -0,0 +1,437 @@
...
+function mailFrom(source: Partial<MailSource>): MailSource {
+  return {
+    normalize: (cb) => cb(null, source as MailSource),
+  } as MailSource;
+}
+
+/** Yields to the microtask queue so the transport's normalize/send chain settles. */
+function flush(): Promise<void> {
+  return new Promise((resolve) => setImmediate(resolve));
+}
```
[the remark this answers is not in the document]
cabljac: The docstring was already corrected in 6fb9ecc7; it now says "Defers to the check phase", which is exactly where setImmediate lands. Keeping setImmediate: these tests only run under Node (vitest node environment), and setTimeout(0) would add a timer clamp for no benefit.

kits/firestore-send-email/tests/prepare-payload.test.ts
```diff
@@ -96,3 +98,337 @@ describe("preparePayload", () => {
...
+      to: "test@example.com",
+      template: { name: "html-only-template", data: {} },
+      message: { text: "Original text content", subject: "Original Subject" },
+      attachments: [{ filename: "original.doc" }],
+    });
+
+    expect(result.message.html).toBe("<h1>Template HTML</h1>");
+    expect(result.message.text).toBe("Original text content");
+    expect(result.message.subject).toBe("Template Subject");
+  });
```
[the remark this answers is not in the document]
cabljac: Addressed in 9d146ae4. The top-level attachments key was an inert leftover from the legacy test - preparePayload never reads it - so it was dropped rather than moved, and a dedicated test ("message attachments survive a template that renders none") pins attachments inside message surviving a template that renders none.

kits/firestore-send-email/tests/helpers.test.ts
cabljac: Applied in b8cbd0b7. config.test.ts now captures the SMTP_CONNECTION_URI validationRegex via the defineString mock and pins it: documented URI forms plus blank accepted, missing scheme or port rejected, and an explicit test pinning the unanchored-alternative behaviour on your two trailing-garbage URIs, so the anchor fix will flip a named test when it lands. The regex fix itself stays a src follow-up as you suggested.

kits/firestore-send-email/tests/helpers.test.ts
cabljac: Applied in b8cbd0b7: "builds a hotmail service transport for Microsoft SMTP hosts" now covers the checkMicrosoftServer branch - service hotmail, decoded user and pass, and tls undefined.

kits/firestore-send-email/tests/helpers.test.ts
cabljac: Renamed in b8cbd0b7 to "round-trips special characters in the smtpPassword secret" (and the URI-password sibling to match).

kits/firestore-send-email/tests/prepare-payload.test.ts
cabljac: Agreed, assertion kept as-is. Renamed in b8cbd0b7 to "a template that renders nothing only adds an empty attachments array".

kits/firestore-send-email/tests/config.test.ts
```diff
@@ -90,6 +102,43 @@ describe("secretParamsForAuthType", () => {
...
+      expect(connectionUriRegex().test(uri)).toBe(true);
+    }
+  });
+
+  test("rejects a URI without a scheme or without a port", () => {
+    expect(connectionUriRegex().test("smtp.gmail.com:465")).toBe(false);
+    expect(connectionUriRegex().test("smtp://smtp.gmail.com")).toBe(false);
+  });
+
+  test("accepts trailing garbage after a valid prefix because the first alternative is unanchored", () => {
```
IzaakGough, line 133: The shipped `validationRegex` in `src/config.ts:171` splits on `|` into two whole alternatives, and the first one has no `$`, so anything after a valid prefix is accepted. Both URIs asserted here pass it and both throw from `new URL()`, for example:

```
smtp://fakeemail@gmail.com:4,h?dhuNTbv9zMrP4&7&7%*3:smtp.gmail.com:465?pool=true&service=gmail
```

For a user, that is a `:` typed where the `@` before the host belongs: install-time validation accepts the URI, and then every queued email fails in `setSmtpCredentials` with "Invalid URI: please reconfigure with a valid SMTP connection URI".

The legacy suite asserted `false` for that same URI, but against a separate anchored regex literal declared in the test file (`...(\?[^ ]*)?)$`) rather than the one in `config.ts`, so the gap was not visible there.

Anchoring the first alternative would fix it:

```js
/^(?:smtp[s]?:\/\/(?:.*?(?::[^:@]*)?@)?[^:@]+:[0-9]+(?:\?[^ ]*)?)$|^$/
```

If you would rather keep this PR test-only, could the assertion reference a tracked issue instead? As written, a passing test defends the missing anchor, so fixing it later reads as a test break.

kits/firestore-send-email/tests/helpers.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      smtpPassword: "secret/password",
+    });
+
+    const credentials = setSmtpCredentials(config);
+
+    expect(credentials).toBeInstanceOf(Mail);
+    expect(credentials.options.service).toBe("hotmail");
+    expect(credentials.options.auth.user).toBe("fakeemail@outlook.com");
+    expect(credentials.options.auth.pass).toBe("secret/password");
+    expect(credentials.options.tls).toBeUndefined();
```
IzaakGough, line 178: This pins that the Microsoft branch of `setSmtpCredentials` drops `config.tlsOptions`. Every other host path goes through `createTransport(url.href, { tls: parseTlsOptions(config.tlsOptions) })`, but when `checkMicrosoftServer` matches (hostnames containing `outlook`, `office365`, or `hotmail`) the branch builds `createTransport({ service: "hotmail", auth })` with no `tls` at all.

So a user on `smtp-mail.outlook.com` or an Office 365 relay who sets `TLS_OPTIONS` to a non-default value, `{"rejectUnauthorized":false}` for a self-signed corporate relay being the common one, has it silently ignored.

This test is also the only coverage on that branch, so it would be good for it not to certify the dropped setting. Passing `tls: parseTlsOptions(config.tlsOptions)` in the hotmail branch and asserting it here would cover both. If dropping it is deliberate, a one-line note saying why would help the next reader.

kits/firestore-send-email/tests/config.test.ts
cabljac: Pushed 13462450: the test now carries a reference to #3067, which tracks anchoring the regex in both the kit and the legacy extension.yaml.

kits/firestore-send-email/tests/helpers.test.ts
cabljac: Checked the legacy source: origin/master firestore-send-email/functions/src/helpers.ts (lines 71-78) builds the Microsoft transport as createTransport({ service: "hotmail", auth }) with no tls either, so the drop is inherited and this test pins parity. Restoring TLS_OPTIONS on that branch would be a behavior change beyond the extension, so I'd keep this PR as-is and track the src fix separately - happy to open an issue for it.

PR #3058, opened 2026-09-01
test(firestore-genai-chatbot): port missing coverage from the legacy extension

cabljac:
The kit migration dropped the legacy extension's logger and overrides suites, and `handlers.ts` arrived with no coverage at all. This restores them, test-only, with no source changes.

`logger.test.ts` and `overrides.test.ts` are ports. The logger suite asserts behaviour instead of reaching into the private level field. The overrides suite pins what the kit's schema actually does with numeric strings: `parseInt(arg, 10) || undefined` drops a non-numeric value and also drops a legitimate `"0"`, and a non-numeric float string yields `NaN`, which passes `z.number()` and reaches the generation options unchecked. The `NaN` pass-through is inherited from the legacy extension and deserves a small src follow-up on the #2974 ledger; these tests only pin it, they do not fix it.

`handlers.test.ts` drives `handleDocumentWrite` with the fake-snapshot style the onwrite-processor tests already use, mocking only the generative client and Firestore reads, so it covers config through `createProcessor` to the status writes and the error mapper.

`tests/integration/chat-flow.emulator.test.ts` runs the same flow against a real Firestore emulator, covering the dotted-path status update, the order-field backfill, and the history query. It is gated on `FIRESTORE_EMULATOR_HOST`, so `npm test` skips it; `npm run test:emulator` runs it. It passed three consecutive runs locally, but no CI job runs kit emulator tests today.

Legacy `config.test.ts` was deliberately not ported: its `EXT_INSTANCE_ID`/env shapes are extension-runtime specific and the export-config and deploy-options suites already cover the kit-native equivalents.

25 new tests (47 passing, 3 emulator-gated). `tsc -b` clean, test files typecheck, prettier clean.

Part of #2974.

### DIFF
--- a/kits/firestore-genai-chatbot/tests/handlers.test.ts
+++ b/kits/firestore-genai-chatbot/tests/handlers.test.ts
@@ -1,0 +1,159 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { beforeEach, describe, expect, test, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  fetchHistory: vi.fn(),
+  fetchDiscussionOptions: vi.fn(),
+  getGenerativeClient: vi.fn(),
+  send: vi.fn(),
+}));
+
+vi.mock("../src/firestore", () => ({
+  fetchHistory: mocks.fetchHistory,
+  fetchDiscussionOptions: mocks.fetchDiscussionOptions,
+}));
+
+vi.mock("../src/generative-client", () => ({
+  getGenerativeClient: mocks.getGenerativeClient,
+}));
+
+import { resolveConfig } from "../src/export-config";
+import { State } from "../src/firestore-onwrite-processor/common";
+import type { DocumentWriteEvent, HandlerContext } from "../src/handlers";
+import { createProcessor, handleDocumentWrite } from "../src/handlers";
+
+function config(overrides: Record<string, unknown> = {}) {
+  return resolveConfig({
+    projectId: "project",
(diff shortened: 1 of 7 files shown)


IzaakGough (approved):

IzaakGough (commented):
Ran both suites locally: `npx vitest run` passes (47 passed, 3 skipped) and `npm run test:emulator` passes 3/3. The new assertions genuinely pin behaviour rather than passing vacuously, so nothing here blocks. Two non-blocking notes on coverage that reads a little broader than it is.

kits/firestore-genai-chatbot/tests/logger.test.ts
```diff
@@ -0,0 +1,149 @@
...
+const mocks = vi.hoisted(() => ({
+  debug: vi.fn(),
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+}));
+
+vi.mock("firebase-functions", () => ({ logger: mocks }));
+
+import { Logger, LogLevel, logger } from "../src/logger";
```
IzaakGough, line 28: Most of this file exercises `setLogLevel` filtering, but `logger`, `Logger`, and `LogLevel` are not exported from `src/lib.ts` (which `src/index.ts` re-exports wholesale), and nothing in `src/` calls `setLogLevel` on the kit's own logger. So the level is effectively fixed at `INFO` for anyone consuming the package, and `logger.debug("Generating response with Genkit")` in `src/generative-client/genkit.ts` can never be emitted.

Not a problem with the tests themselves, but worth deciding which way to resolve it: export the logger from `src/lib.ts` so the level really is configurable, or trim this part of the coverage so it does not imply a surface that is not reachable.

kits/firestore-genai-chatbot/tests/handlers.test.ts
```diff
@@ -0,0 +1,159 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { beforeEach, describe, expect, test, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  fetchHistory: vi.fn(),
+  fetchDiscussionOptions: vi.fn(),
```
IzaakGough, line 21: `fetchDiscussionOptions` is mocked and gets a `mockResolvedValue` in `beforeEach`, but `resolveConfig` defaults `enableOverrides` to `false` and no test in this file sets it, so `createGenerateChatResponse` never reaches that branch. The mock makes the per-discussion overrides path look covered when it is not, including the case where `extractOverrides` throws and the processor maps it to an `ERROR` status.

Either add a case with `enableOverrides: true`, or drop the mock so the gap is visible.

PR #3064, opened 2026-09-01
fix(storage-resize-images): map the Off content-filter option to OFF

### LINKED ISSUE #3047: fix(storage-resize-images): the "Off" content-filter selection stores "False" and fails every event
The select option "Off (No filtering)" in `kits/storage-resize-images/src/config.ts:233` maps to the value `"False"`, not `"OFF"`. A user who picks Off deploys with `CONTENT_FILTER_LEVEL=False`, and `convertHarmBlockThreshold("False")` throws `Invalid HarmBlockThreshold: False` on every event, so the kit is unusable with filtering off. The dead const `CONTENT_FILTER_OPTIONS` at `config.ts:44` shows the intended values and is used nowhere.

Fix: change the select value to `"OFF"`; use or delete `CONTENT_FILTER_OPTIONS`. Found during the adversarial review of #3044 (the bug predates it). Part of the #2974 stabilisation work.

cabljac:
Picking "Off (No filtering)" for the content filter breaks the kit completely: the select option stored "False", the deployed environment carried CONTENT_FILTER_LEVEL=False, and convertHarmBlockThreshold threw "Invalid HarmBlockThreshold: False" on every storage event. The fix maps the option to "OFF", which the resolver converts to a null threshold and content-filter.ts treats as filtering disabled. The dead CONTENT_FILTER_OPTIONS const, which recorded the intended values but was used nowhere, is deleted.

New tests in tests/config.test.ts read the CONTENT_FILTER_LEVEL param from declaredParams and pin the full option list, resolve every option through resolveResizeImagesConfig (OFF yields a null threshold), and pin that "False" is rejected. The option-list and resolve tests fail on the previous code. Full kit suite, tsc -b, and prettier are clean. Verified in unit tests only; no live deploy was run.

Fixes #3047


kits/storage-resize-images/tests/config.test.ts
```diff
@@ -214,6 +217,70 @@ describe("configFromEnv", () => {
...
+
+  test('the retired "False" option value is rejected', async () => {
+    const { resolveResizeImagesConfig } = await import("../src/export-config");
+    expect(() =>
+      resolveResizeImagesConfig({
+        ...baseConfig,
+        contentFilterLevel: "False" as never,
+      })
+    ).toThrow("Invalid HarmBlockThreshold: False");
+  });
```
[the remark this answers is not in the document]
cabljac: "False" was never functional, so there is no working deployment to protect: with CONTENT_FILTER_LEVEL=False the kit already crashes on every storage event, which is exactly the bug this PR fixes. After upgrading, such a deployment behaves identically until the user re-configures, at which point the select writes "OFF" and they are fixed. Mapping "False" to null would silently disable a safety filter on an invalid value and grandfather a never-valid sentinel into the config surface, and the kits are pre-stable with no meaningful install base. Keeping the loud error is deliberate.

kits/storage-resize-images/src/config.ts
```diff
@@ -41,12 +41,6 @@ const IMAGE_TYPE_OPTIONS = [
   "false",
```
IzaakGough: The `imageTypes` param further down has the same mismatch this PR fixes, and it is still there: the multiSelect maps `original` to `"False"`, while every consumer compares the lowercase `"false"` listed here. `resize-image.ts` does `const shouldFormatImage = format !== "false"`, `export-config.ts` has `DEFAULT_IMAGE_TYPE = "false"`, the param's own default is `["false"]`, and `extension.yaml` uses `value: false` for that option.

So picking "original" in the UI gives `IMAGE_TYPE=["False"]`, `shouldFormatImage` stays true, `SUPPORTED_IMAGE_CONTENT_TYPE_MAP["False"]` is undefined, and sharp is asked to convert to a format called `False`. Every resize fails, and only for the option that is meant to do nothing. Since this PR is already auditing select values against what the runtime accepts, it seems worth fixing in the same change.

kits/storage-resize-images/src/config.ts
```diff
@@ -41,12 +41,6 @@ const IMAGE_TYPE_OPTIONS = [
...
 ] as const;
 const MEMORY_OPTIONS = [512, 1024, 2048, 4096, 8192] as const;
```
IzaakGough: `IMAGE_TYPE_OPTIONS` just above and `MEMORY_OPTIONS` here are unreferenced for the same reason `CONTENT_FILTER_OPTIONS` was, so it might be worth removing all three together rather than leaving two behind.

kits/storage-resize-images/src/config.ts
```diff
@@ -232,7 +226,7 @@ const params = {
 
     default: "OFF",
     input: select({
-      "Off (No filtering)": "False",
+      "Off (No filtering)": "OFF",
```
IzaakGough, line 219: One case this does not reach: a param value already present in `.env.<project>` is reused rather than re-prompted, so an install that previously picked "Off" keeps `CONTENT_FILTER_LEVEL=False` and still throws `Invalid HarmBlockThreshold: False` after upgrading and redeploying. The new test pins `"False"` as rejected, so the options are accepting it as an alias for `OFF` in `convertHarmBlockThreshold`, or a CHANGELOG line telling people to edit the stored value. At `0.0.2-rc.0` that may well not be worth doing, just flagging it. This rests on reading the param code rather than a real deploy.

CorieW:

**Unit tests + typecheck** — `kits/storage-resize-images`: **179 passed / 6 skipped (187)**, `tsc -b --force` clean. The typecheck matters here because the PR deletes the `CONTENT_FILTER_OPTIONS` const outright; nothing else referenced it.

Regression-proved by reverting `src/config.ts` to the base — both new cases go red:

```
× offers OFF and the three block thresholds
× every option resolves, and OFF disables filtering
Tests  2 failed | 12 passed (14)
```

**Parity is exact.** The extension's `storage-resize-images/extension.yaml:370-371` maps the same label to `OFF`:

```yaml
- param: CONTENT_FILTER_LEVEL
  options:
    - label: Off (No filtering)
      value: OFF
```

So the kit's `"False"` was a straight typo against the extension, not an intentional deviation — confirms the fix direction.

Also worth saying: reading the declared param back out of `declaredParams` and asserting the full `select.options` array is a better test than checking the one changed string. It pins the invariant the bug violated (every option the select offers must be a value the resolver accepts) rather than just the symptom.

**Live deploy** — repacked the kit into a local functions-kit project and deployed all three resize instances to a real project:

```
3 x "Successful create operation."
+  Deploy complete!
```

This is a real exercise of the fixed path rather than a formality — the deployed instance configs are `CONTENT_FILTER_LEVEL=OFF` (x2) and `BLOCK_ONLY_HIGH` (x1), so deploy-time resolution ran against the corrected `OFF` value on two of the three.

**Not verified:** no image uploaded to confirm filtering is actually disabled at runtime. Application Default Credentials on this machine need reauth (`invalid_grant` / `invalid_rapt`), so I could not write to the bucket to trigger `generateResizedImage`.

No problems found.

kits/storage-resize-images/src/config.ts
cabljac: Confirmed and applied in da26ef4e: the multiSelect now stores lowercase "false" for original, matching the param default, DEFAULT_IMAGE_TYPE, and the format !== "false" check in resize-image.ts. Added tests pinning the IMAGE_TYPE option list and that every conversion value is a key of SUPPORTED_IMAGE_CONTENT_TYPE_MAP.

kits/storage-resize-images/src/config.ts
cabljac: Applied in da26ef4e: IMAGE_TYPE_OPTIONS and MEMORY_OPTIONS in config.ts are gone (the MEMORY_OPTIONS lookup map in export-config.ts is a different, referenced const and stays).

kits/storage-resize-images/src/config.ts
cabljac: Right about the reuse path. Keeping the loud error - "False" never worked, so such a deploy was already failing on every event, and aliasing it would grandfather a never-valid sentinel - but da26ef4e adds CHANGELOG lines telling upgraders to edit a stored CONTENT_FILTER_LEVEL=False to OFF (and IMAGE_TYPE ["False"] to ["false"], which has the same reuse caveat).

kits/storage-resize-images/CHANGELOG.md
```diff
@@ -1 +1,3 @@
+- The "Off (No filtering)" content-filter option now stores `OFF` instead of `False`, which the resolver rejected, so every storage event threw `Invalid HarmBlockThreshold: False`. A `.env` written by an earlier deploy keeps the stored value on upgrade: if it carries `CONTENT_FILTER_LEVEL=False`, edit it to `OFF` (or re-select the option).
+- The "original" image-type option now stores `false` instead of `False`, which the resize path treated as a target format named `False`, failing every resize for that option. A stored `IMAGE_TYPE=["False"]` likewise needs editing to `["false"]`.
```
IzaakGough: A stored `IMAGE_TYPE=["False"]` doesn't actually fail the resize, so this line may send people looking for errors that aren't there.

I ran `modifyImage` with `format: "False"` against the kit's own bucket stub: `convertType` has no branch for it, so the buffer passes through unchanged, the call returns `{ success: true }`, and it uploads `images/test_200x200.False` with `contentType: undefined` (GCS then serves `application/octet-stream`). The finish event reports success.

Worth describing the real symptom instead: resized files named `*.False` with no content type.

kits/storage-resize-images/src/config.ts
```diff
@@ -166,7 +150,7 @@ const params = {
       tiff: "tiff",
       gif: "gif",
       avif: "avif",
-      original: "False",
+      original: "false",
```
IzaakGough, line 153: This fixes what new deploys store, but nothing at runtime normalizes or rejects a stale `IMAGE_TYPE=["False"]`. Combined with the silent pass-through described in the changelog comment, an instance that upgrades keeps writing mis-named, mis-typed resized images indefinitely with no signal in the logs. `CONTENT_FILTER_LEVEL=False` at least throws `Invalid HarmBlockThreshold: False`.

Two options: normalize the value where `imageTypes` is resolved in `resolveResizeImagesConfig` (no legitimate format is named `False`), or make an unrecognized format throw in `convertType`/`modifyImage` so a stale config fails loudly instead of corrupting output.

kits/storage-resize-images/src/config.ts
IzaakGough, line 1: Minor, and outside the diff so I couldn't anchor it to the line: the `isAnimated` select (line 189) is the last artifact of the same YAML-boolean coercion this PR fixes. `extension.yaml` has `label: Yes` / `label: No (1st frame only)`, and `Yes` was coerced to `True` when the param was ported, the same way `value: false` became `"False"`. So the deploy prompt reads "True" instead of "Yes".

Grep says these three were the only occurrences across all kits, so changing this label closes the class. The new option-audit tests pin values but not labels, so nothing would catch it.

kits/storage-resize-images/CHANGELOG.md
cabljac: Agreed, reworded to the observed behaviour in 60532811: a stale IMAGE_TYPE=["False"] was uploaded as <name>_<size>.False with no content type and reported as a success.

kits/storage-resize-images/src/config.ts
cabljac: Agreed, but rejecting unknown image types at resolve time is stricter than the extension, so it is split out as #3124 to keep this PR scoped to the OFF mapping.

kits/storage-resize-images/src/config.ts
cabljac: Agreed, split out as #3123.

IzaakGough (approved):
lgtm

Issue #3013, opened 2026-09-01
fix(firestore-vector-search): validate queryOnWrite prefilters


cabljac:
`prefilters` is cast unvalidated; the extension zod-parses it. Restore the parse.

Parity ledger: #2974, firestore-vector-search §7b.

cabljac:
Fixed by #3065, merged into kits.

PR #3065, opened 2026-09-01
fix(firestore-vector-search): validate prefilters on the query onWrite path

cabljac:
On the queryOnWrite path, `handleQueryOnWrite` cast `data.prefilters as Prefilter[]` without validation, so malformed prefilters on a watched query document (a non-array, or non-object entries) either failed downstream inside the Firestore query or silently misbehaved. The legacy extension (firestore-vector-search v0.1.3) zod-parses prefilters with `z.array(prefilterSchema)` on this path (parity ledger #2974, item 7b).

This reuses the kit's existing `prefilterSchema` via a new `parsePrefilters` helper and replaces the cast at the point of use, throwing `Invalid prefilters: ...` before any embedding work. The legacy processor additionally caught the error and wrote an ERROR status to the document; the kit throws out of this handler today, and status parity is deferred to ledger item 7c.

Tests: a new `handleQueryOnWrite prefilters validation` describe block, confirmed red before the fix; full suite, `tsc -b`, and prettier pass. Note #3038 touches the head of the same function and appends its own onWrite tests; whichever lands second rebases trivially.

Fixes #3013

Following review (050ce5d5): an explicit `prefilters: null` is treated as no prefilters, preserving the `?? []` behavior the parse replaced, and validation errors include each issue's path so a mixed array names the index of the bad entry. Note the parse checks the top-level shape only (an array of objects), matching legacy's `z.record(z.any())` prefilterSchema; a typo'd key inside a prefilter is not caught here and still surfaces from the Firestore query downstream.

### DIFF
--- a/kits/firestore-vector-search/src/queries/index.ts
+++ b/kits/firestore-vector-search/src/queries/index.ts
@@ -26,10 +26,16 @@ const prefiltersSchema = z.array(prefilterSchema);
 
 /** Validates raw prefilters from a watched query document. */
 export function parsePrefilters(data: unknown): Prefilter[] {
-  if (data === undefined) return [];
+  if (data == null) return [];
   const parsed = prefiltersSchema.safeParse(data);
   if (!parsed.success) {
-    const issues = parsed.error.issues.map((issue) => issue.message).join("; ");
+    const issues = parsed.error.issues
+      .map((issue) =>
+        issue.path.length
+          ? `${issue.path.join(".")}: ${issue.message}`
+          : issue.message
+      )
+      .join("; ");
     throw new Error(`Invalid prefilters: ${issues}`);
   }
   return parsed.data;
--- a/kits/firestore-vector-search/src/queries/index.ts
+++ b/kits/firestore-vector-search/src/queries/index.ts
@@ -22,6 +22,19 @@ import type { FirestoreVectorStoreClient } from "../vector-store";
 export const prefilterSchema = z.record(z.any());
 export type Prefilter = z.infer<typeof prefilterSchema>;
 
+const prefiltersSchema = z.array(prefilterSchema);
+
+/** Validates raw prefilters from a watched query document. */
+export function parsePrefilters(data: unknown): Prefilter[] {
+  if (data === undefined) return [];
+  const parsed = prefiltersSchema.safeParse(data);
+  if (!parsed.success) {
+    const issues = parsed.error.issues.map((issue) => issue.message).join("; ");
+    throw new Error(`Invalid prefilters: ${issues}`);
+  }
+  return parsed.data;
+}
+
 export interface ParsedQueryRequest {
   query: string;
   limit?: string | number;
--- a/kits/firestore-vector-search/src/handlers.ts
+++ b/kits/firestore-vector-search/src/handlers.ts
@@ -26,8 +26,8 @@ import type { ResolvedVectorSearchConfig } from "./export-config";
 import * as logs from "./logs";
 import {
   FirestoreVectorStoreClient,
-  type Prefilter,
   parseLimit,
+  parsePrefilters,
   parseQuerySchema,
   performTextQuery,
 } from "./queries";
@@ -129,7 +129,7 @@ export async function handleQueryOnWrite(
   const result = await performTextQuery({
     query,
     limit: data.limit ? parseLimit(data.limit) : ctx.config.defaultQueryLimit,
-    prefilters: (data.prefilters as Prefilter[] | undefined) ?? [],
+    prefilters: parsePrefilters(data.prefilters),
     embedClient: embedClient(ctx),
     vectorStore: vectorStore(ctx),
     config: ctx.config,
--- a/kits/firestore-vector-search/tests/handlers.test.ts
+++ b/kits/firestore-vector-search/tests/handlers.test.ts
@@ -284,4 +284,29 @@ describe("handleQueryOnWrite prefilters validation", () => {
     expect(chain.where).not.toHaveBeenCalled();
     expect(set).toHaveBeenCalledWith({ result: { ids: IDS } }, { merge: true });
   });
+
+  test("treats an explicit null prefilters as no prefilters", async () => {
+    const { ctx, chain } = makeCtx();
+    const { event, set } = queryDocEvent({
+      query: "test query",
+      prefilters: null,
+    });
+
+    await handleQueryOnWrite(event, ctx);
+
+    expect(chain.where).not.toHaveBeenCalled();
+    expect(set).toHaveBeenCalledWith({ result: { ids: IDS } }, { merge: true });
+  });
+
+  test("names the offending entry's index in the error", async () => {
+    const { ctx } = makeCtx();
+    const { event } = queryDocEvent({
+      query: "test query",
+      prefilters: [{ field: "category", operator: "==", value: "test" }, 42],
+    });
+
+    await expect(handleQueryOnWrite(event, ctx)).rejects.toThrow(
+      "Invalid prefilters: 1: Expected object, received number"
+    );
+  });
 });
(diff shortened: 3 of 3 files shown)


kits/firestore-vector-search/src/queries/index.ts
```diff
@@ -22,6 +22,19 @@ import type { FirestoreVectorStoreClient } from "../vector-store";
 export const prefilterSchema = z.record(z.any());
 export type Prefilter = z.infer<typeof prefilterSchema>;
 
+const prefiltersSchema = z.array(prefilterSchema);
+
+/** Validates raw prefilters from a watched query document. */
+export function parsePrefilters(data: unknown): Prefilter[] {
+  if (data === undefined) return [];
```
IzaakGough: This returns early only for `undefined`, so an explicit `null` now throws. The expression it replaces was `(data.prefilters as Prefilter[] | undefined) ?? []`, and `??` also covered `null`.

A query doc written as `{ query: "cats", prefilters: null }` (for example a caller doing `prefilters: filters ?? null`) used to run with no prefilters and get its `result` written back. Now `safeParse(null)` fails with "Expected array, received null", `handleQueryOnWrite` throws, and nothing is written to the doc, so a client waiting on it sees neither a result nor a status.

If rejecting `null` is intended that seems reasonable, just worth being explicit about. Otherwise `if (data == null) return [];` covers both cases.

kits/firestore-vector-search/src/queries/index.ts
```diff
@@ -22,6 +22,19 @@ import type { FirestoreVectorStoreClient } from "../vector-store";
 export const prefilterSchema = z.record(z.any());
```
IzaakGough, line 22: `z.record(z.any())` accepts any object, so `[{}]` and `[{ field: "a" }]` both pass (checked against the kit's zod 3). A prefilter with a typo'd key, say `[{ fields: "category", operator: "==", value: "x" }]`, gets through validation and then reaches `where(undefined, undefined, undefined)` in `FirestoreVectorStoreClient.query`, which surfaces as `HttpsError("unknown", 'Value for argument "fieldPath" is not a valid field path...')`.

That matches the legacy schema, so it may well be deliberate for parity. If so, it might help to note in the PR body that this validates the top-level shape rather than the individual prefilter fields, since those are the malformations users are most likely to hit.

kits/firestore-vector-search/src/queries/index.ts
```diff
@@ -22,6 +22,19 @@ import type { FirestoreVectorStoreClient } from "../vector-store";
...
 export type Prefilter = z.infer<typeof prefilterSchema>;
 
+const prefiltersSchema = z.array(prefilterSchema);
+
+/** Validates raw prefilters from a watched query document. */
+export function parsePrefilters(data: unknown): Prefilter[] {
+  if (data === undefined) return [];
+  const parsed = prefiltersSchema.safeParse(data);
+  if (!parsed.success) {
+    const issues = parsed.error.issues.map((issue) => issue.message).join("; ");
```
IzaakGough: Dropping `issue.path` loses the only thing identifying which entry is bad. With `prefilters: [{ field: "a", operator: "==", value: 1 }, "category == test", 42]` the thrown message reads `Invalid prefilters: Expected object, received string; Expected object, received number`, with no indices. Since this handler writes no status field, that log line is the only place the error shows up.

Including the path, or just using `parsed.error.message`, keeps the index.

CorieW:
### Test report

**Unit tests** — `npx vitest run` in `kits/firestore-vector-search`: **53/53 pass**, with all four new cases running:

```
✓ handleQueryOnWrite prefilters validation > runs the query and applies valid prefilters
✓ handleQueryOnWrite prefilters validation > rejects a non-array prefilters value before embedding
✓ handleQueryOnWrite prefilters validation > rejects a prefilter entry that is not an object
✓ handleQueryOnWrite prefilters validation > treats missing prefilters as no prefilters
```

Regression-proved by reverting `src/handlers.ts` and `src/queries/index.ts` to the base:

```
× rejects a non-array prefilters value before embedding
× rejects a prefilter entry that is not an object
Tests  2 failed | 12 passed (14)
```

The two "rejects" cases are the ones the blind `as Prefilter[]` cast let through, and the other two staying green confirms the happy path and the `undefined` path are unchanged. Rejecting *before* embedding is the right ordering too — it avoids burning an embedding call on a request that can't be served.

**Live deploy** — repacked the kit into a local functions-kit project and deployed all three vector-search instances (`-a91f6c2e`, `-euclid`, `-openai`) to a real project:

```
24 x "Successful create operation."
+  Deploy complete!
```

### One problem: merge conflict with #3038

Both this branch and #3038 append a new `describe` block (plus the same `handleQueryOnWrite` / `VectorWriteEvent` imports) to `kits/firestore-vector-search/tests/handlers.test.ts`. `git merge-tree` against the merge base reports **4 conflict hunks, all in that one test file**:

```
kits/firestore-vector-search/tests/handlers.test.ts
+<<<<<<< .our   (x4)
```

`src/handlers.ts` is touched by both but auto-merges cleanly, and the merged result is semantically right — #3038's echo guard sits above `performTextQuery`, and this PR's `parsePrefilters` replaces an argument inside the call. So only the test file needs hand-resolving, by whichever of the two lands second.

### Not verified

No live query document written against the deployed functions: Application Default Credentials on this machine need reauth (`invalid_grant` / `invalid_rapt`), so the prefilter rejection is confirmed at unit level only, not against a real `queryOnWrite` invocation.

kits/firestore-vector-search/src/queries/index.ts
```diff
@@ -22,6 +22,19 @@ import type { FirestoreVectorStoreClient } from "../vector-store";
 export const prefilterSchema = z.record(z.any());
 export type Prefilter = z.infer<typeof prefilterSchema>;
 
+const prefiltersSchema = z.array(prefilterSchema);
```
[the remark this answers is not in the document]
cabljac: Keeping the loose schema deliberately: it reuses the kit's existing prefilterSchema, which matches the legacy z.record(z.any()) exactly, and this PR's scope is restoring the legacy parse, not tightening it. prefilterSchema also backs the callable query path, so tightening it here would change behavior beyond this PR. The PR body now states explicitly that only the top-level shape is validated.

kits/firestore-vector-search/src/queries/index.ts
```diff
@@ -22,6 +22,19 @@ import type { FirestoreVectorStoreClient } from "../vector-store";
...
+/** Validates raw prefilters from a watched query document. */
+export function parsePrefilters(data: unknown): Prefilter[] {
+  if (data === undefined) return [];
+  const parsed = prefiltersSchema.safeParse(data);
+  if (!parsed.success) {
+    const issues = parsed.error.issues.map((issue) => issue.message).join("; ");
+    throw new Error(`Invalid prefilters: ${issues}`);
+  }
+  return parsed.data;
+}
```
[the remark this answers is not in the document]
cabljac: Both points applied in 050ce5d5 (they were raised independently in review): parsePrefilters returns [] for null as well as undefined, and error messages include each issue's path. Tests cover the null case and the indexed message.

kits/firestore-vector-search/src/queries/index.ts
cabljac: Applied in 050ce5d5: data == null returns [], so prefilters: null runs with no prefilters as the ?? [] it replaced did. Pinned by a new test.

kits/firestore-vector-search/src/queries/index.ts
cabljac: Applied in 050ce5d5: issues now render as path: message when a path exists, so the mixed-array case reads "Invalid prefilters: 1: Expected object, received number". A test pins the indexed message.

kits/firestore-vector-search/src/queries/index.ts
cabljac: Kept as-is for parity - prefilterSchema matches legacy's z.record(z.any()) and also backs the callable query path. The PR body now notes explicitly that validation covers the top-level shape only and a typo'd key still surfaces from the Firestore query downstream.

IzaakGough (approved):

Issue #3026, opened 2026-09-01
fix(speech-to-text): remove the retained replace("tmp/", ...) in the .txt path


cabljac:
The `.txt` output path still carries the extension's `replace("tmp/", …)` transform, wrong regardless of where the transcoded `.wav` location decision (in #2974) lands.

Parity ledger: #2974, speech-to-text §1.

PR #3073, opened 2026-09-01
fix(speech-to-text): stop stripping tmp/ from the .txt transcription path

cabljac:
The `.txt` transcription output URI retained `replace("tmp/", "")` from the legacy extension, where the transcoded `.wav` was uploaded under an object name derived from its local `/tmp` path. The kit uploads the `.wav` bucket-relative, so the remnant only mangled object paths that legitimately contain `tmp/` (e.g. `audio/tmp/clip.mp3` produced a transcript at `audio/clip.mp3.wav_transcription.txt`). The transform is removed; the `.txt` now lands at `<transcoded object>_transcription.txt` exactly, as the README states.

Added a unit test on `transcribeAndUpload` pinning the `gcsUri` for a `tmp/`-containing path; it was red against the remnant and is green after. Full vitest suite, `tsc -b` and prettier pass. Live transcription was not run.

The open decision on where the transcoded `.wav` lands (#2974 Migration section) is out of scope and unchanged.

Fixes #3026

### DIFF
--- a/kits/speech-to-text/CHANGELOG.md
+++ b/kits/speech-to-text/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Initial release of kit, see README for differences between the legacy extension and this kit
+- The `.txt` transcription output no longer has `tmp/` stripped from its path, a remnant of the legacy extension's temp-file handling; it lands at `<transcoded object>_transcription.txt` exactly ([#3026](https://github.com/firebase/extensions/issues/3026))
--- a/kits/speech-to-text/README.md
+++ b/kits/speech-to-text/README.md
@@ -150,8 +150,11 @@ The transcript itself is written to the same place as before
 (`<original path>.wav_transcription.txt`, under `OUTPUT_STORAGE_PATH` when set),
 so only the intermediate audio moves. If you have lifecycle rules, cleanup jobs
 or client code that expect the WAV under a `tmp/` prefix, point them at the new
-path. Both files still carry the `isTranscodeOutput` metadata flag that stops the
-function from processing its own output.
+path. The transcoded `.wav` still carries the `isTranscodeOutput` metadata flag
+that stops the function from processing its own output. The transcript `.txt` is
+written directly by the Speech-to-Text API and carries no metadata, so its
+finalize event runs the function again; that run creates a transcript document
+for the `.txt` object and marks it `FAILED` with "Invalid content type.".
 
 ### The function may now run for nine minutes
 
--- a/kits/speech-to-text/src/transcribe.ts
+++ b/kits/speech-to-text/src/transcribe.ts
@@ -65,10 +65,7 @@ export async function transcribeAndUpload({
   options: SpeechOptions;
 }): Promise<TranscribeAudioResult> {
   const inputUri = `gs://${bucket.name}/${name}`;
-  const outputUri = `gs://${bucket.name}/${name.replace(
-    "tmp/",
-    ""
-  )}_transcription.txt`;
+  const outputUri = `gs://${bucket.name}/${name}_transcription.txt`;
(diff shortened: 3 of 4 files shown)


IzaakGough (approved):

IzaakGough:
One thing I noticed while reading around this change, not a request on this PR.

The transcript `.txt` is written by the Speech API via `outputConfig.gcsUri`, so it does not get the `isTranscodeOutput` metadata that `uploadTranscodedFile` sets on the `.wav`. `transcribeAudio` is registered with `onObjectFinalized` and only a bucket filter, so the `.txt` finalize event re-enters `handleObjectFinalized`, which creates a transcript document and then marks it `FAILED` with "Invalid content type."

That is unchanged by this diff (the `.txt` already landed in the same bucket), but `README.md` line 153 says "Both files still carry the `isTranscodeOutput` metadata flag", which does not match the code. So either the README line needs a correction, or the handler needs another way to recognise its own transcript output.

CorieW:

**Unit tests + typecheck** — `kits/speech-to-text`: **34/34 pass**, `tsc -b --force` clean.

Regression-proved by reverting `src/transcribe.ts` to the base — the new case goes red with exactly the mangled path the issue describes:

```
× writes the .txt next to the uploaded object, even when its path contains tmp/

-  "gcsUri": "gs://my-bucket/audio/tmp/clip.mp3.wav_transcription.txt",   (expected)
+  "gcsUri": "gs://my-bucket/audio/clip.mp3.wav_transcription.txt",       (received)
```

Good choice of fixture — `audio/tmp/clip.mp3.wav` puts `tmp/` in the middle of the path, which is precisely where the old `String.replace` (first-occurrence, unanchored) did the damage. A fixture with a leading `tmp/` would have passed either way.

**I independently confirmed the "legacy remnant" reasoning**, since removing a `replace` is the kind of change that can quietly relocate output. `transcribeAndUpload` is only ever handed the *uploaded storage object*, and that object's name is derived from the input object path — never from the local temp dir:

```ts
// src/handlers.ts
const localCopyPath: string = path.join(os.tmpdir(), filePath);   // local only, never passed on
...
const transcodedObjectName = `${filePath}.wav`;                   // bucket-relative
const transcodedUploadResult = await ctx.fns.uploadTranscodedFile({ storagePath: ..., bucket });
const [file] = transcodedUploadResult.uploadResponse;
await ctx.fns.transcribeAndUpload({ client, file, ... });         // <- storage object
```

So `name` reaching `transcribeAndUpload` can never contain the function's `/tmp` working path. The only strings the old `replace` could ever match were user-chosen paths that happen to contain `tmp/` — i.e. it was strictly a bug, and dropping it cannot relocate anything that was previously correct.

**Live deploy** — repacked the kit into a local functions-kit project and deployed both speech-to-text instances (`-a91f6c2e`, `-video`) to a real project:

```
2 x "Successful create operation."
+  Deploy complete!
```

Verified the vendored build actually carries the change (`replace("tmp/"` → 0 occurrences, `_transcription.txt` → 1) rather than trusting the repack.

**Not verified:** no audio file uploaded to confirm the `.txt` lands at the new path in the bucket. Application Default Credentials on this machine need reauth (`invalid_grant` / `invalid_rapt`), so I could not write to Storage to trigger a transcription.

No problems found.

cabljac:
Confirmed against the code: only uploadTranscodedFile sets isTranscodeOutput, and the .txt finalize re-enters handleObjectFinalized, creating a transcript doc marked FAILED with "Invalid content type.". 71d4e489 corrects the README to describe that behavior instead of claiming both files carry the flag. Teaching the handler to recognise its own transcript output is left as a follow-up, since it changes runtime behavior and this PR is a path fix.

PR #3089, opened 2026-09-02
fix(firestore-send-email): keep SMTP_PASSWORD available so SendGrid works with AUTH_TYPE=OAuth2

### LINKED ISSUE #3009: fix(firestore-send-email): SendGrid sends fail with AUTH_TYPE=OAuth2

cabljac:
SendGrid sends fail in the kit when AUTH_TYPE=OAuth2. configFromEnv gates secrets on the auth type and forces smtpPassword to undefined under OAuth2, but transportLayer selects the SendGrid transport purely on the SMTP URI host (smtp.sendgrid.net) and passes smtpPassword as the SendGrid API key. The transport therefore got apiKey: undefined, sgMail.setApiKey was never called, and every send failed. The legacy extension reads SMTP_PASSWORD unconditionally (firestore-send-email/functions/src/config.ts and index.ts:86-90 on master), so SendGrid + OAuth2 worked there.

The fix passes the SMTP password secret through for every auth type, and secretParamsForAuthType now keeps SMTP_PASSWORD bound under OAuth2 so a future gated deploy binding cannot reintroduce the bug. New tests pin configFromEnv keeping the secret under OAuth2 and transportLayer handing the key to setApiKey; the changelog gets an entry. Rebased onto kits after #3057 merged, folding the new tests into the ported suites. Suite: 123 tests pass, tsc --noEmit clean. An adversarial review mutation-tested both halves of the fix (reverting config.ts fails the config tests, reverting the transport apiKey fails the transportLayer test) and found no regression in any consumer of smtpPassword; the obfuscated-config log path still redacts it. Caveat: no live SendGrid send was verified; the behavior is pinned at unit level against a mocked @sendgrid/mail.

Fixes #3009

### DIFF
--- a/kits/firestore-send-email/src/config.ts
+++ b/kits/firestore-send-email/src/config.ts
@@ -311,7 +311,14 @@ export const secretParams = [
 export function secretParamsForAuthType(authType?: string) {
   switch (authType || AuthenticatonType.UsernamePassword) {
     case AuthenticatonType.OAuth2:
-      return [params.clientId, params.clientSecret, params.refreshToken];
+      // The SendGrid transport reads SMTP_PASSWORD as its API key regardless
+      // of auth type, so it must stay bound under OAuth2.
+      return [
+        params.smtpPassword,
+        params.clientId,
+        params.clientSecret,
+        params.refreshToken,
+      ];
     case AuthenticatonType.UsernamePassword:
     case AuthenticatonType.ApiKey:
       return [params.smtpPassword];
@@ -336,10 +343,9 @@ export function configFromEnv(): SendEmailConfig {
     databaseRegion: params.databaseRegion.value(),
     mailCollection: params.mailCollection.value(),
     smtpConnectionUri: params.smtpConnectionUri.value(),
-    smtpPassword:
-      authType === AuthenticatonType.OAuth2
-        ? undefined
-        : optionalSecret(params.smtpPassword),
+    // Not gated on auth type: the SendGrid transport reads it as its API key
+    // even when AUTH_TYPE is OAuth2.
+    smtpPassword: optionalSecret(params.smtpPassword),
     defaultFrom: params.defaultFrom.value(),
     defaultReplyTo: params.defaultReplyTo.value(),
     usersCollection: params.usersCollection.value(),
--- a/kits/firestore-send-email/tests/config.test.ts
+++ b/kits/firestore-send-email/tests/config.test.ts
@@ -26,10 +26,15 @@ const { stringParamOpts } = vi.hoisted(() => ({
 }));
 
 vi.mock("firebase-functions/params", () => ({
+  // Only stubbed names may read the env: ambient shell vars (USER, HOST)
+  // collide with real param names and would make results machine-dependent.
   defineString: (name: string, opts?: StringParamOpts) => {
     stringParamOpts.set(name, opts);
     return {
-      value: () => process.env[name] ?? opts?.default ?? "",
+      value: () =>
+        (name === "AUTH_TYPE" ? process.env[name] : undefined) ??
+        opts?.default ??
+        "",
     };
   },
   defineInt: (_name: string, opts?: { default?: number }) => ({
--- a/kits/firestore-send-email/tests/config.test.ts
+++ b/kits/firestore-send-email/tests/config.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { describe, expect, test, vi } from "vitest";
+import { afterEach, describe, expect, test, vi } from "vitest";
 
 interface StringParamOpts {
   default?: string;
@@ -29,7 +29,7 @@ vi.mock("firebase-functions/params", () => ({
   defineString: (name: string, opts?: StringParamOpts) => {
     stringParamOpts.set(name, opts);
     return {
-      value: () => opts?.default ?? "",
+      value: () => process.env[name] ?? opts?.default ?? "",
     };
   },
   defineInt: (_name: string, opts?: { default?: number }) => ({
@@ -60,6 +60,10 @@ import { resolveConfig } from "../src/export-config";
 import { AuthenticatonType } from "../src/types";
 
 describe("configFromEnv", () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
   test("maps params and keeps secret-backed values deferred", () => {
     const config = configFromEnv();
     expect(config.mailCollection).toBe("mail");
@@ -67,6 +71,16 @@ describe("configFromEnv", () => {
     expect(config.defaultReplyTo).toBe("");
     expect(typeof config.smtpPassword).toBe("object");
     expect(config.clientId).toBeUndefined();
+  });
+
+  test("keeps the SMTP password for OAuth2 so SendGrid can use it as API key", () => {
+    vi.stubEnv("AUTH_TYPE", AuthenticatonType.OAuth2);
+    const config = configFromEnv();
+    expect(config.authType).toBe(AuthenticatonType.OAuth2);
+    expect(typeof config.smtpPassword).toBe("object");
+    expect(typeof config.clientId).toBe("object");
+    expect(typeof config.clientSecret).toBe("object");
+    expect(typeof config.refreshToken).toBe("object");
   });
 });
 
@@ -79,12 +93,12 @@ describe("secretParamsForAuthType", () => {
     ).toEqual(["SMTP_PASSWORD"]);
   });
 
-  test("binds only OAuth secrets for OAuth2 auth", () => {
+  test("keeps the SMTP password bound alongside OAuth secrets for OAuth2 auth", () => {
     expect(
       secretParamsForAuthType(AuthenticatonType.OAuth2).map(
         (secret) => (secret as { name: string }).name
       )
-    ).toEqual(["CLIENT_ID", "CLIENT_SECRET", "REFRESH_TOKEN"]);
+    ).toEqual(["SMTP_PASSWORD", "CLIENT_ID", "CLIENT_SECRET", "REFRESH_TOKEN"]);
   });
 
   test("uses username/password secret binding by default", () => {
--- a/kits/firestore-send-email/tests/config.test.ts
+++ b/kits/firestore-send-email/tests/config.test.ts
@@ -21,20 +21,19 @@ interface StringParamOpts {
   input?: { text?: { validationRegex?: RegExp } };
 }
 
-const { stringParamOpts } = vi.hoisted(() => ({
+const { stringParamOpts, paramEnv } = vi.hoisted(() => ({
   stringParamOpts: new Map<string, StringParamOpts | undefined>(),
+  paramEnv: new Map<string, string>(),
 }));
 
 vi.mock("firebase-functions/params", () => ({
-  // Only stubbed names may read the env: ambient shell vars (USER, HOST)
-  // collide with real param names and would make results machine-dependent.
+  // Values come from paramEnv rather than process.env: AUTH_TYPE, USER and
+  // HOST are all real param names that collide with ambient shell vars, which
+  // would otherwise make results machine-dependent.
   defineString: (name: string, opts?: StringParamOpts) => {
     stringParamOpts.set(name, opts);
     return {
-      value: () =>
-        (name === "AUTH_TYPE" ? process.env[name] : undefined) ??
-        opts?.default ??
-        "",
+      value: () => paramEnv.get(name) ?? opts?.default ?? "",
     };
   },
   defineInt: (_name: string, opts?: { default?: number }) => ({
@@ -66,7 +65,7 @@ import { AuthenticatonType } from "../src/types";
 
 describe("configFromEnv", () => {
   afterEach(() => {
-    vi.unstubAllEnvs();
+    paramEnv.clear();
   });
 
   test("maps params and keeps secret-backed values deferred", () => {
@@ -79,7 +78,7 @@ describe("configFromEnv", () => {
   });
 
   test("keeps the SMTP password for OAuth2 so SendGrid can use it as API key", () => {
-    vi.stubEnv("AUTH_TYPE", AuthenticatonType.OAuth2);
+    paramEnv.set("AUTH_TYPE", AuthenticatonType.OAuth2);
     const config = configFromEnv();
     expect(config.authType).toBe(AuthenticatonType.OAuth2);
     expect(typeof config.smtpPassword).toBe("object");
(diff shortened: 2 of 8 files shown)


kits/firestore-send-email/src/config.ts
```diff
@@ -311,7 +311,14 @@ export const secretParams = [
...
     case AuthenticatonType.OAuth2:
-      return [params.clientId, params.clientSecret, params.refreshToken];
+      // The SendGrid transport reads SMTP_PASSWORD as its API key regardless
+      // of auth type, so it must stay bound under OAuth2.
+      return [
+        params.smtpPassword,
+        params.clientId,
+        params.clientSecret,
+        params.refreshToken,
+      ];
```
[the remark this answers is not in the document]
cabljac: secretParamsForAuthType has no production consumer: index.ts binds secretParams (all four secrets) unconditionally, so this hunk changes no deploy-time binding - it only pins the config-level invariant. Gating it on the URI would also not work as suggested: params.smtpConnectionUri.value() reads an empty process.env at deploy/manifest time, so the conditional would silently evaluate false. And URI-conditional binding would diverge from the legacy extension, which mounts SMTP_PASSWORD unconditionally (extension.yaml declares it type: secret, always bound). Parity is the bar here.

kits/firestore-send-email/src/config.ts
```diff
@@ -336,10 +343,9 @@ export function configFromEnv(): SendEmailConfig {
...
     mailCollection: params.mailCollection.value(),
     smtpConnectionUri: params.smtpConnectionUri.value(),
-    smtpPassword:
-      authType === AuthenticatonType.OAuth2
-        ? undefined
-        : optionalSecret(params.smtpPassword),
+    // Not gated on auth type: the SendGrid transport reads it as its API key
+    // even when AUTH_TYPE is OAuth2.
+    smtpPassword: optionalSecret(params.smtpPassword),
     defaultFrom: params.defaultFrom.value(),
```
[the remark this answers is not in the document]
cabljac: SecretParam.value() does not throw at runtime when the secret is unset - it warns and returns "", which normalizeOptional maps to undefined, so there is no error path to guard. The deploy-time throw is unreachable because resolveConfig runs lazily inside ensureInitialized(), never at module load. SMTP_PASSWORD is also always bound via the unconditional secretParams in index.ts.

kits/firestore-send-email/tests/config.test.ts
```diff
@@ -56,6 +60,16 @@ describe("configFromEnv", () => {
...
+
+  test("keeps the SMTP password for OAuth2 so SendGrid can use it as API key", () => {
+    vi.stubEnv("AUTH_TYPE", AuthenticatonType.OAuth2);
+    const config = configFromEnv();
+    expect(config.authType).toBe(AuthenticatonType.OAuth2);
+    expect(typeof config.smtpPassword).toBe("object");
+    expect(typeof config.clientId).toBe("object");
+    expect(typeof config.clientSecret).toBe("object");
+    expect(typeof config.refreshToken).toBe("object");
+  });
```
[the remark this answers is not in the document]
cabljac: This presupposes the URI-gated binding from the other thread; a test expecting the SMTP password to be dropped for OAuth2 without SendGrid would pin the legacy-divergent bug this PR removes. The legacy extension reads SMTP_PASSWORD unconditionally for every auth type, and the kit now matches it.

kits/firestore-send-email/tests/config.test.ts
```diff
@@ -67,12 +81,12 @@ describe("secretParamsForAuthType", () => {
...
 
-  test("binds only OAuth secrets for OAuth2 auth", () => {
+  test("keeps the SMTP password bound alongside OAuth secrets for OAuth2 auth", () => {
     expect(
       secretParamsForAuthType(AuthenticatonType.OAuth2).map(
         (secret) => (secret as { name: string }).name
       )
-    ).toEqual(["CLIENT_ID", "CLIENT_SECRET", "REFRESH_TOKEN"]);
+    ).toEqual(["SMTP_PASSWORD", "CLIENT_ID", "CLIENT_SECRET", "REFRESH_TOKEN"]);
   });
```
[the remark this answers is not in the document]
cabljac: Same as the sibling thread: this presupposes URI-gated binding, which would diverge from the legacy extension's unconditional SMTP_PASSWORD mount. Keeping the password for every auth type is the parity behavior this PR restores.

kits/firestore-send-email/tests/config.test.ts
```diff
@@ -48,6 +48,10 @@ import { resolveConfig } from "../src/export-config";
 import { AuthenticatonType } from "../src/types";
 
 describe("configFromEnv", () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
```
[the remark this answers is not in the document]
cabljac: Kept local: configFromEnv is the only describe that stubs envs, so scoping the afterEach there keeps the cleanup next to the stubbing it undoes.

kits/firestore-send-email/tests/config.test.ts
```diff
@@ -60,6 +65,10 @@ import { resolveConfig } from "../src/export-config";
 import { AuthenticatonType } from "../src/types";
 
 describe("configFromEnv", () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
```
IzaakGough: `vi.unstubAllEnvs()` restores stubs but does not clear a value that was already in the environment, and the mock above now reads `process.env.AUTH_TYPE`. Anyone who sources the kit's `.env` into their shell (`AUTH_TYPE=OAuth2` is a real key in it) gets `maps params and keeps secret-backed values deferred` failing on `expect(config.clientId).toBeUndefined()`. A `beforeEach` that stubs `AUTH_TYPE` to `undefined` would close the one variable the new comment's machine-dependence warning leaves open.

kits/firestore-send-email/tests/helpers.test.ts
```diff
@@ -253,3 +260,18 @@ describe("isSendGrid", () => {
     expect(isSendGrid(makeConfig({}))).toBe(false);
   });
 });
+
+describe("transportLayer", () => {
```
IzaakGough, line 266: This block has no `beforeEach(() => vi.clearAllMocks())`, unlike the `setSmtpCredentials` describe above, so `setApiKey` calls carry over across the file. It passes today only because no earlier test builds a SendGrid transport. The first one that does would let the `setApiKey` assertion match a stale call, and the test would stop pinning the fix.

IzaakGough (commented):
The fix looks right. `transportLayer` checks `isSendGrid` before `setSmtpCredentials`, so the SendGrid branch is reachable under `AUTH_TYPE=OAuth2` and now gets a real API key, and un-gating `smtpPassword` is safe for the OAuth2 non-SendGrid path because `setSmtpCredentials` returns `setupOAuth2` before touching it. `logs.ts` still redacts it, and nothing else reads it. Verified locally at this head: 123 tests pass, typecheck and prettier clean, and restoring the old OAuth2 gate makes the new config test fail, so it does pin the behaviour.

One thing outside the diff that I think is worth including here: `kits/firestore-send-email/README.md` around line 156 still tells OAuth2 installs to create `SMTP_PASSWORD` with a placeholder value. Someone on SendGrid who follows that ends up at `sgMail.setApiKey("<placeholder>")` and every send fails with a 401, so the setup path the CHANGELOG entry describes as fixed still breaks. A short carve-out saying the real API key goes in `SMTP_PASSWORD` when the connection URI is `smtp.sendgrid.net`, whatever `AUTH_TYPE` is set to, would cover it.

The two inline notes are both about test durability rather than current behaviour, so neither needs to block.

IzaakGough (approved):

kits/firestore-send-email/src/config.ts
```diff
@@ -336,10 +343,9 @@ export function configFromEnv(): SendEmailConfig {
     databaseRegion: params.databaseRegion.value(),
     mailCollection: params.mailCollection.value(),
     smtpConnectionUri: params.smtpConnectionUri.value(),
-    smtpPassword:
-      authType === AuthenticatonType.OAuth2
-        ? undefined
-        : optionalSecret(params.smtpPassword),
+    // Not gated on auth type: the SendGrid transport reads it as its API key
+    // even when AUTH_TYPE is OAuth2.
+    smtpPassword: optionalSecret(params.smtpPassword),
```
IzaakGough, line 348: Raising this from testing the branch rather than from the diff. This change does what it says, and running it end to end surfaces a second problem one layer down.

I ran the kit in the functions emulator with `AUTH_TYPE=OAuth2` and `SMTP_CONNECTION_URI=smtps://apikey@smtp.sendgrid.net:465`, then wrote a doc to the mail collection. Before this change the logged config had no `smtpPassword` and the send died with `TypeError: sgMail.send is not a function`. With this change `smtpPassword` reaches the transport, and it now dies earlier, at init, with `TypeError: sgMail.setApiKey is not a function`.

The cause is `src/nodemailer-sendgrid/index.ts:17`, `import * as sgMail from "@sendgrid/mail"`. That package exports a `MailService` instance, so `setApiKey` and `send` are prototype methods. The kit's `tsconfig.json` sets `esModuleInterop: true`, so tsc emits `__importStar`, which copies own properties only:

```
own props: [ 'client', 'substitutionWrappers', 'secretRules', 'MailService' ]
```

Both methods are lost. The legacy extension has the same import line but no `esModuleInterop` in its tsconfig, so it compiles to a plain `require` and works. It is not OAuth2 specific either: I reproduced the same `setApiKey` crash with `AUTH_TYPE=UsernamePassword`. The unit tests do not catch it because they mock `@sendgrid/mail`.

Switching to `import sgMail from "@sendgrid/mail"` fixes it. With both changes the same emulator run reaches SendGrid and writes a real result, `delivery.state: ERROR` with `Unauthorized (401)` for my deliberately fake key, and the logs show `API key does not start with "SG."` coming from `setApiKey`.

No strong view on whether it belongs in this PR or a follow-up, but as it stands merging this alone moves SendGrid from a send-time crash to an init-time one.

IzaakGough: Pushed this as 2c2a3258 so the branch is testable end to end.

Switching to a default import needed the two `@sendgrid/mail` mocks updated as well: they exposed named exports only, so 20 tests went to zero recorded calls until the factories returned the object as both the default and the named exports, which is the real module's shape.

The guard runs against `lib/` rather than `src/`. Vitest's transform does not reproduce the tsc `__importStar` emit, so a normal unit test passes with the bug present. I checked that both ways: with the namespace import restored it is the only one of the 124 tests that fails, and it passes once the import is fixed. CI already builds before it tests, so the ordering holds there.

One thing I left alone: `src/nodemailer-sendgrid/index.ts` is not prettier clean at the union type near the end of `send`, and it was that way before this change, so it seemed better not to mix a reformat into the diff.

IzaakGough:
Closed out the testing on a live SendGrid key, running the kit in the functions emulator with `AUTH_TYPE=OAuth2` and the SendGrid connection URI.

The config change does what it says. `smtpPassword` now reaches the transport, where before it was absent from the resolved config entirely. With a real key the request authenticates and SendGrid answers at the account level, so `setApiKey` and `send` both complete the round trip. That last part only became reachable with 2c2a3258, since before it the transport threw on `setApiKey` regardless of auth type.

What is not proven: the send stopped at `Maximum credits exceeded`, so `delivery.state: SUCCESS` and the `sendgridQueueId` parsing never ran. That path is untouched here and covered by the existing unit tests, so it did not seem worth chasing a credited account. The control is useful though, a deliberately fake key on the same rig returns `The provided authorization grant is invalid, expired, or revoked`, so the credits response is SendGrid resolving the key to an account rather than rejecting it.

All of this was the emulator rather than a real deploy, so deploy-time secret binding onto the service is still untested.

IzaakGough:
Took this to a live deploy on a test project to cover the one thing the emulator cannot reach, deploy-time secret binding.

The deployed service binds all four secrets as env vars, `SMTP_PASSWORD` included, under `AUTH_TYPE=OAuth2`. A document written to the real `mail` collection was picked up, the function read the key out of Secret Manager, built the SendGrid transport and authenticated, landing on the same `Maximum credits exceeded` as the emulator run. The transport constructing at all is 2c2a3258 working in a deployed build, since the previous one died at `setApiKey` before any network call.

One thing found along the way, separate from this PR's subject. Secret params resolve to two different resources depending on which layer of the deploy is asking. `applyPrefix` rewrites each endpoint's secret env vars to `kit-<instance id>-<PARAM>`, but it does not set `resourceId` on the params, so the params layer falls through to the bare name. Both halves are observable:

- The params layer reaches for the bare name. An instance with no secrets of its own bound onto an unrelated project level `SMTP_PASSWORD` that happened to exist, and a missing one reports `In non-interactive mode but have no value for the secret CLIENT_ID: CLIENT_ID`, followed by `Set this secret before deploying: firebase functions:secrets:set CLIENT_ID`.
- Endpoint validation then asks for the prefixed name: `Secret [projects/.../secrets/kit-default-SMTP_PASSWORD] not found or has no versions`.

So following the CLI's own suggestion creates a secret the deploy then ignores, and `functions:secrets:set` cannot create the prefixed name either, since it uppercases the argument and `KIT_` is a reserved prefix. Creation uses that same bare `resourceId`, so the interactive prompt should land in the same place, though I did not drive the prompt itself.

What does work today is `FIREBASE_SECRET_REF_<PARAM>=projects/<p>/secrets/<name>/versions/latest` in the instance's `.env`. It overrides both layers, takes any resource name, and needs no prompt, so it is also the only CI safe route. That one I confirmed with a dry run rather than a completed deploy. Creating `kit-<instance>-<PARAM>` by hand with gcloud works too, and is what got the deploy above through.

For anyone coming from an extension none of this should be manual. `firebase ext:export --mode functions` writes those same `FIREBASE_SECRET_REF_` entries pointing at the live `ext-<instance>-<PARAM>` resources, with a secret ejection prompt alongside. That sits behind the private `extMigrationFeatures` experiment and `ext:migrate` is currently a stub, so it is not usable yet. That paragraph is from reading firebase-tools 15.29.0 rather than running it.

This looks like a firebase-tools issue rather than anything to change in this PR, and the kit README wording is downstream of whatever it settles on.

kits/firestore-send-email/tests/config.test.ts
IzaakGough: Fixed in 7e8d39fc. The param mock now reads a map the tests control instead of process.env, so AUTH_TYPE, USER and HOST can no longer leak in from the shell. Verified with `AUTH_TYPE=OAuth2 USER=someone HOST=example.com` in the environment, which used to fail and now passes, and reverting the config change still fails the OAuth2 test, so it is still pinning the behaviour.

kits/firestore-send-email/tests/helpers.test.ts
IzaakGough: Added in dcb0f004.

Issue #3023, opened 2026-09-01
fix(storage-resize-images): restore the us-central1 content-filter fallback


cabljac:
The content-filter region reads `FUNCTION_REGION` and throws when absent; the extension fell back to `us-central1`. Check first whether #3002 already covers this path.

Parity ledger: #2974, storage-resize-images §5.

cabljac:
Fixed by #3090, merged into kits (10ccf945). Note: kits-base PRs do not auto-close issues (auto-close only fires on the default branch), hence the manual close.

Issue #3028, opened 2026-09-01
decision(kits): AI-provider location fallback after the LOCATION removal


cabljac:
`VERTEX_AI_MODEL_LOCATION=null` no longer means "function region" (firestore-genai-chatbot §3), firestore-translate-text takes the vertex region from `FUNCTION_REGION`/genkit default (§2), firestore-vector-search same class (§5). Decide one fallback rule for all three kits.

we're sticking with an explicit DATABASE_REGION param, see firestore-bigquery-export for now.

Ledger: #2974.

IzaakGough:
Closing as fixed by #3162

PR #3090, opened 2026-09-02
fix(storage-resize-images): restore the us-central1 content-filter fallback

cabljac:
The kit's `checkImageContent` threw `FUNCTION_REGION is required for Vertex AI filtering.` when no region was passed; the extension fell back to `us-central1` (master, storage-resize-images/functions/src/content-filter.ts:121). This restores the fallback at the same point of use: the Vertex AI plugin now receives `location ?? "us-central1"` and the throw is gone. `config.ts` and `export-config.ts` are untouched, so there is no overlap with #3002 and the resolved config still reports `FUNCTION_REGION` faithfully.

Normal CLI deploys never hit the throw: firebase-tools injects `FUNCTION_REGION` into every gen2 function's env (verified in 15.26.0, lib/gcp/cloudfunctionsv2.js:197). It was reachable for library consumers of `checkImageContent`, emulator runs, and hand-rolled envs.

The throw test is replaced by a fallback test (verified to fail against the old code) plus an explicit-location passthrough test; README, CHANGELOG, and a stale test comment are updated. 182 tests and `tsc --noEmit` pass; the live Vertex filtering path was not exercised. This instantiates a "FUNCTION_REGION if set, else us-central1" rule relevant to the open decision in #3028.

Fixes #3023


IzaakGough (approved):
Looks good. Reviewed the change and ran the kit suite locally (182 passed, 3 skipped).

A few things I checked:

- `location` was already optional, and the only caller (`src/handlers.ts`) passes `config.region`, so nothing downstream changes.
- Matches the extension's behaviour, which resolves `LOCATION ?? "us-central1"`.
- An empty-string region skips the `??` fallback, but `@genkit-ai/vertexai` resolves `options?.location || "us-central1"` internally, so it still lands on the same default.
- The new fallback test is doing real work: putting the old throw back makes exactly that test fail.

Not exercised: the live Vertex filtering path, same as the description notes.

PR #3092, opened 2026-09-02
fix(firestore-vector-search): restore terminal-state skip on embedOnWrite

### LINKED ISSUE #3011: fix(firestore-vector-search): restore re-embed and ERROR-retry semantics
The extension never re-embeds documents in a terminal state and does not retry on ERROR; the kit flips both. Restore both together, they share the status machinery.

Parity ledger: #2974, firestore-vector-search §6a/6b.

CorieW:
Fixes #3011.

The extension's `FirestoreOnWriteProcessor` embedded each document once. It skipped any document whose status state was `PROCESSING`, `COMPLETED`, `ERROR` or `BACKFILLED`, so editing the input after a successful embed did nothing (§6a) and a failed embed was never retried (§6b). The kit instead skipped on `beforeInput === input && output exists`, so it re-embedded on every input edit and retried every `ERROR` document.

This PR restores the extension's rule. The guard reads the kit's flat `status.state`; the nested-vs-flat status shape (§6) stays tracked in #2974.

## Changes

- `handleEmbedOnWrite` skips a document whose status state is `PROCESSING`, `COMPLETED`, `ERROR` or `BACKFILLED`, replacing the input-comparison guard.
- `handleEmbedOnWrite` and `embedPath` skip an empty-string input, as the extension's `shouldProcess`, `shouldBackfill` and `shouldUpdate` did. Without this, an empty document is marked terminal and the new guard never releases it (review follow-up, 4f5e04b5e).
- README: removed the "editing the input re-embeds" difference, and added the embed-once rule, the empty-input rule and the way to re-embed (delete `status`, write again) under `Unchanged`.
- Tests: `handleEmbedOnWrite` coverage for embed-and-`COMPLETED`, `ERROR` write plus rethrow, deleted document, non-string and empty input, no re-embed for each terminal state, embed when the state is absent or unrecognised, and a custom `statusFieldName`.

## Testing

`vitest run` in `kits/firestore-vector-search`: 141 tests pass. `tsc -b` and `prettier` clean.

Deployed to `corie-testing` in a throwaway codebase (`fvs-pr3011`, own instance, collection, secrets; since deleted). Gemini, `us-central1`:

- **Fresh document**: embedded, `status: { state: "COMPLETED" }`, 768-dim vector.
- **§6a**: with the document at `COMPLETED`, changed `input` to unrelated text. After 120s the vector was byte-identical and the status untouched.
- **§6b**: seeded `status: { state: "ERROR", message: "SEEDED-TRACER" }`, then edited. After 60s the message was unchanged and no embedding was written.
- **`BACKFILLED`, `PROCESSING`**: seeded then edited; both untouched.
- **Removed guard**: a document with an all-zero embedding and no status was re-embedded, so the old input-comparison skip is gone.
- **No self-retrigger**: the status write re-enters `embedOnWrite` once and the guard absorbs it; the document was unchanged 120s later.

### Out of scope

With `UPDATE_ON_CONFIGURE: true`, every `updateTask` failed with a missing `GEMINI_API_KEY`: `DEFAULT_TASK_OPTIONS` and `EMBEDDING_TASK_OPTIONS` bind no secrets. Pre-existing on `kits`; tracked on `fix/kits-vector-search-task-secrets`.


CorieW:
Review pass 1 — two gaps in the test coverage added by this PR.

**1. `kits/firestore-vector-search/tests/handlers.test.ts` — the custom `statusFieldName` test was not discriminating.** It set a terminal state on *both* `status` and `embedStatus`, so it passed whichever field the guard actually read, and would not have caught a hardcoded `status`. Split into two tests: a terminal state on the configured field skips, and a terminal state on the default `status` field is ignored when `statusFieldName` is `embedStatus`.

**2. `kits/firestore-vector-search/tests/handlers.test.ts` — no test covered the removal of the old input-comparison guard.** Every case exercised a changed input or a fresh document, so the suite would still pass with `beforeInput === input && output exists → return` left in place. The extension's skip rule was the status state alone, so an unchanged document that already carries an embedding but no status is still processed. Added that case.

kits/firestore-vector-search/tests/handlers.test.ts
```diff
@@ -205,3 +211,184 @@ describe("handleQueryCall", () => {
...
+
+    await handleEmbedOnWrite(event, embedCtx());
+
+    expect(getSingleEmbedding).toHaveBeenCalledWith("hello");
+  });
+
+  describe("with a custom status field name", () => {
+    const customConfig = resolveVectorSearchConfig({
+      projectId: "test-project",
+      instanceId: "test-instance",
```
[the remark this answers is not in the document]
CorieW: Already addressed in 1c95e33f, which landed just after this review ran against 6c0fb441. Same finding, same two cases — split into separate `skips on the configured field` and `ignores a terminal state on the default field` tests under a `with a custom status field name` describe, rather than one test with a mid-body `vi.clearAllMocks()`, so a failure names which direction broke.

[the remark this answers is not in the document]
CorieW: Verified by mutation: changing the guard to `isInTerminalState(data, "status")` — the exact bug this comment describes — fails both tests (`skips on the configured field`, `ignores a terminal state on the default field`); both pass on HEAD. No further change needed.

CorieW:
Some similarities to https://github.com/firebase/extensions/pull/3095

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -78,12 +99,9 @@ export async function handleEmbedOnWrite(
   logs.start("embedOnWrite");
 
   const data = event.data.after.data() ?? {};
+  if (isInTerminalState(data, ctx.config.statusFieldName)) return;
   const input = data[ctx.config.inputFieldName];
   if (typeof input !== "string") return;
```
IzaakGough: The extension's `shouldProcess` was `!!newData[inputField] && typeof newData[inputField] === 'string'`, so an empty string was skipped. Here an empty string passes, which was harmless while the input-comparison guard could recover the document, but the terminal-state guard makes it permanent:

1. A document is created as `{ input: "" }`, say a form document created first and filled in a moment later.
2. `embedOnWrite` embeds the empty string and writes `status: { state: "COMPLETED" }`, or the provider rejects it and it writes `status: { state: "ERROR" }`.
3. The user sets the real input. The status is terminal, so the write is skipped and the document never gets a real embedding. No error, no log.

The extension writes no status at all at step 2, so step 3 embeds normally. I ran both variants (`COMPLETED` and `ERROR`) against this branch and neither re-embeds. The same two tests fail against `kits`, where the document is still embedded once the input is filled in, so this is new here rather than pre-existing.

Matching `shouldProcess` covers it:

```ts
if (typeof input !== "string" || !input) return;
```

A test for empty-then-filled would be worth having alongside the terminal-state cases.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -47,6 +47,27 @@ export type VectorWriteEvent = FirestoreEvent<
   Record<string, string>
 >;
 
+/**
+ * States the extension's `FirestoreOnWriteProcessor` treated as final. A
+ * document that has reached one of these is never processed again, so each
+ * document is embedded once and a failure is never retried.
+ */
```
IzaakGough, line 59: Minor, on the comment rather than the code. In the extension `PROCESSING` was written by `writeStartEvent` before the embed ran, so it doubled as an in-flight marker. The kit only ever writes `COMPLETED` and `ERROR`, so `PROCESSING` is reachable here only on documents an installed instance left behind. Worth a line saying so, since the comment currently reads as though the kit drives all four states itself.

kits/firestore-vector-search/README.md
```diff
@@ -251,23 +251,14 @@ longer written, only `COMPLETED` and `ERROR`. Anything reading
 `status.<instance id>.state`, or a security rule or index keyed to it, needs
 updating. The field name is still `STATUS_FIELD_NAME`, defaulting to `status`.
 
+`embedOnWrite` still reads the same four states the extension treated as final —
+`PROCESSING`, `COMPLETED`, `ERROR` and `BACKFILLED` — so documents an installed
+instance already embedded are still skipped once you flatten their status field.
```
IzaakGough: The section this replaces was the only place the README said that a document is embedded once and that editing `input` afterwards does not refresh its embedding. Removing it from the differences list is right, since it is no longer a difference, but a reader who never ran the extension now has nowhere to learn it, and a stale embedding fails silently.

A line under `### Unchanged` would cover it, along with the way out: clearing the document's `status` field and writing it again does re-embed, which I confirmed against this branch.

IzaakGough (commented):
I checked the parity claim against the real `FirestoreOnWriteProcessor` source (`@invertase/firebase-extension-utilities@0.1.3`) and it holds: it skips `[PROCESSING, COMPLETED, ERROR, BACKFILLED]` read off the status map, and it checks state before `shouldProcess`, which is the ordering the new guard uses. `FAILED_BACKFILL`, the fifth state the backfill helper writes, is correctly left out of the set. Built the branch and ran the suite: 62 pass, `tsc -b` clean.

One regression I think is worth fixing before this lands, plus two smaller notes, all inline.

CorieW:
Merged `kits` in and resolved the conflicts (ab6f1b82d).

The only behavioral resolution: `tests/events.test.ts` on `kits` had a skipped-write case built on the old input-comparison guard (unchanged input plus an existing embedding), which this PR removes. It now uses a `COMPLETED` document instead. The rest was mechanical: keep both hunks in `src/handlers.ts`, rebuild `tests/handlers.test.ts` on the kits `writeEvent`/`makeCtx` helpers with the `handleEmbedOnWrite` suite on top, and keep the new query-document status section in the README while correcting its closing sentence, which pointed at the section this PR deletes.

`vitest run` in `kits/firestore-vector-search`: 137 tests pass. `tsc -b` and prettier clean.

kits/firestore-vector-search/src/handlers.ts
CorieW: Added in 4f5e04b5e. The comment now says the kit writes only `COMPLETED` and `ERROR`, and that `PROCESSING` and `BACKFILLED` reach the guard only on documents an installed instance left behind.

kits/firestore-vector-search/README.md
CorieW: Added a bullet under `### Unchanged` in 4f5e04b5e covering the embed-once rule and the way out, clearing the `status` field and writing the document again. It also records that an empty input is skipped and gets no status, which the same commit restores.

kits/firestore-vector-search/src/handlers.ts
CorieW: Agreed, fixed in 4f5e04b5e with `typeof input !== "string" || input === ""`, plus a skip test and an empty-then-filled test that both fail on the old guard. `shouldBackfill` and `shouldUpdate` carried the same check, so `embedPath` gets it too: a backfill over an empty document left it terminal and stuck in exactly the same way.

CorieW: Reviewed and confirmed. [4f5e04b5e](https://github.com/firebase/extensions/commit/4f5e04b5ed9884f5a6484e03550d446d758b6654) restores the extension's behavior.

The guard now reads `typeof input !== "string" || input === ""` in both `handleEmbedOnWrite` and `embedPath`, which matches the extension's `shouldProcess`, `shouldBackfill`, and `shouldUpdate`. An empty input is skipped and gets no status, so the document embeds once the input is filled in. Two new tests (skip on empty, empty-then-filled) cover the path; both fail on the old guard. Resolving.

kits/firestore-vector-search/src/handlers.ts
CorieW: Reviewed and confirmed. [4f5e04b5e](https://github.com/firebase/extensions/commit/4f5e04b5ed9884f5a6484e03550d446d758b6654) adds the missing line: the kit writes only `COMPLETED` and `ERROR`, and `PROCESSING` and `BACKFILLED` reach the guard only on documents an installed extension instance left behind. Resolving.

kits/firestore-vector-search/README.md
CorieW: Reviewed and confirmed. [4f5e04b5e](https://github.com/firebase/extensions/commit/4f5e04b5ed9884f5a6484e03550d446d758b6654) adds a bullet under `### Unchanged` that states the embed-once rule, the way out (delete the `status` field and write the document again), and that an empty input is skipped and gets no status. Resolving.

Issue #3016, opened 2026-09-01
fix(firestore-vector-search): publish events from all paths


cabljac:
Events are only published from `handleEmbedOnWrite`; the query, backfill, update, and init paths publish nothing.

Parity ledger: #2974, firestore-vector-search.

PR #3094, opened 2026-09-02
fix(firestore-vector-search): stop publishing events the extension never sent

CorieW:
Closes #3016.

The issue asks for events to be published from the query, backfill, update and
init paths, on the basis that only `handleEmbedOnWrite` publishes today.

Checked the extension first. `firestore-vector-search` v0.1.3 declares four
event types in `extension.yaml` — `onStart`, `onSuccess`, `onError`,
`onCompletion` under `firebase.extensions.firestore-vector-search.v1.*` — and
publishes **none** of them. There is no events module, no Eventarc channel
setup, and no `publish()` call anywhere in `functions/src`;
`@invertase/firebase-extension-utilities` does not publish on the extension's
behalf either. Every other extension with an `events:` block (e.g.
`storage-resize-images`, `firestore-translate-text`) does have an `events.ts`
that publishes, so this one is an outlier.

So parity here means publishing from **no** path, not from every path. The kit
was the diverging side: it published all four types from `embedOnWrite`, which
matched neither the extension's behaviour nor any other path in the kit. The
kit README documented this as a deliberate improvement ("Events are actually
published now"), which is an upgrade rather than a parity justification.

## Changes

- Delete `src/events.ts` (channel setup and the four `record*Event` helpers).
- Remove the `recordStartEvent`, `recordSuccessEvent`, `recordErrorEvent` and
  `recordCompletionEvent` calls from `handleEmbedOnWrite`, dropping the `finally`
  block that only existed to publish `onCompletion`.
- Remove `events.setupEventChannel()` from the memoized `getContext()` in
  `src/index.ts`.
- README: drop the `## Events` section and the "Events are actually published
  now" difference; note under **Unchanged** that the four declared event types
  are never published and `EVENTARC_CHANNEL` is not read, as in the extension.

## Testing

`vitest run` in `kits/firestore-vector-search`: 5 files, 49 tests, all pass. No test
changed — the extension has no event tests, so there is nothing to port. `tsc -b`
clean, `prettier --list-different` clean, all CI checks green.

Behaviour was deployed and smoke-tested in `corie-testing` from an isolated
single-instance codebase (instance `vs3094`, collection `pr3094_vs`), packed from
this branch. `EVENTARC_CHANNEL` was deliberately set to
`locations/us-central1/channels/firebase`, a channel that does not exist in the
project, so any surviving publish attempt would surface as an error.

- `embedOnWrite` — writing a document produced a 768-dimension native vector in
  `embedding` and `status: { state: "COMPLETED" }`.
- `queryOnWrite` — writing `{ query, limit }` to `_vs3094/index/queries/q1` wrote
  back `result: { ids: ["onwrite-1"] }`.
- `initVectorSearch` — ran on `afterFirstDeploy`, logged its config and started
  Firestore vector index creation.
- No events published — 0 references to Eventarc publishing (`eventarc.publish`,
  the configured channel, `getEventarc`, or any of the four event type names)
  across the logs of all eight deployed functions, despite `EVENTARC_CHANNEL`
  being set. This matches the extension, which publishes nothing.

Backfill could not be exercised: `backfillTask` and `updateTask` fail with
`GenkitError: FAILED_PRECONDITION: Please pass in the API key or set the
GEMINI_API_KEY or GOOGLE_API_KEY environment variable`, because
`EMBEDDING_TASK_OPTIONS` in `src/index.ts` omits `secrets`. This is pre-existing
on `kits` and untouched by this PR, which changes no task options or secrets; it
needs a separate issue.


cabljac (approved):
Good investigation - the upstream verification is thorough (the copy-pasted onSuccess description from storage-resize-images is a nice smoking gun), and resolving #3016 by removal rather than addition is the right parity read.

Two asks before merge:

1. Add a CHANGELOG entry. The rc.0/rc.1 packages on npm do publish these events when EVENTARC_CHANNEL is set, so an rc user upgrading loses behavior with no note - and your #3096/#3097 both carry CHANGELOG entries for behavior changes, so this one should too.

2. Pin the absence with one test. I tried re-adding a publish call in handleEmbedOnWrite and the suite stayed green, which means the exact thing this PR removes can silently come back. A single test asserting no event publish happens on the embed path (e.g. mock the channel and assert it's never created/called) turns the removal into a guarantee. General rule worth keeping: a removal PR should leave behind a test that fails if the removal is reverted, same as a fix PR leaves a test that fails if the fix is reverted.

CorieW:
Both addressed in 8075fb08.

**1. CHANGELOG entry** — added, calling out that the `0.0.2-rc` packages on npm do publish all four events when `EVENTARC_CHANNEL` is set, so an rc user upgrading loses that behaviour and any existing subscription now receives nothing.

**2. Test pinning the absence** — you were right that the suite didn't catch this, and the obvious version of the test doesn't either. Asserting `publish` is never called stays green on a revert of `handlers.ts`, because `record*Event` returns early when `eventChannel` is undefined and `setupEventChannel()` is called from `index.ts`, not from the handler. So a handler-level test never has a live channel to publish through.

`tests/events.test.ts` instead records whether `firebase-admin/eventarc` enters the module graph at all — Vitest only evaluates a `vi.mock` factory if something actually imports the module, so a reintroduced `events.ts` trips it regardless of channel configuration. It also sets `EVENTARC_CHANNEL` and `EXT_SELECTED_EVENTS` and keeps the `publish` assertions, so both the structural and behavioural routes are covered.

Verified the way you asked: restoring the old `events.ts` and `handlers.ts` fails all four tests; against this branch the suite is 53/53.

One note for anyone reading the test — `eventarc.imported` is deliberately never reset in `beforeEach`. The import it records happens once at module load, before any test body runs, so resetting it wipes the evidence. My first draft did exactly that and passed against a full revert.

PR #3097, opened 2026-09-02
fix(firestore-vector-search): restore batched backfill

### LINKED ISSUE #3012: fix(firestore-vector-search): restore batched backfill
Backfill is now per-document tasks and per-document embedding calls, the metadata-doc gate is gone, and `collection().get()` loads every document into memory, so every redeploy re-embeds the whole collection.

Parity ledger: #2974, firestore-vector-search §9.

CorieW:
Closes #3012. Parity ledger: #2974, `firestore-vector-search` §9.

The kit's backfill and update triggers loaded the whole collection with `collection().get()`, enqueued one task per document, embedded one document per call, and never read the metadata document, so every deploy re-embedded the collection. This restores the extension's task-thread backfill (§9a–9g), ported from `@invertase/firebase-extension-utilities`.

## Changes

- `src/backfill.ts` (new): metadata gate, task-thread enqueue, task runner, chunk runner, eligibility rules.
- Enumerate with `listDocuments()` (§9a). 50 ids per task, one task in flight; each task records its counters on `_<instance id>/index`, then enqueues its successor. One enqueue document per chunk under `_<instance id>/index/enqueues` (§9b).
- Backfill embeds each chunk in provider-sized batches; update embeds one document per call (§9c).
- Both passes gate on `embeddingProvider`, `dimension`, `inputField` and `outputField` in the metadata document (§9d).
- Eligibility: non-empty string input; skip when `<status field>.<instance id>.state` is anything but `BACKFILLED`; update also requires an existing embedding (§9e).
- One commit per chunk writes `BACKFILLED` or `FAILED_BACKFILL` with `completeTime`; a failed batch marks its documents failed and the task succeeds (§9f, §9g).
- `handleInit` enqueues one trigger per deploy: backfill when `DO_BACKFILL`, else update when `UPDATE_ON_CONFIGURE`.
- README: `UPDATE_ON_CONFIGURE`, gate and first-deploy, and deviations sections. CHANGELOG updated.
- Tests: `tests/backfill.test.ts` (41), `tests/task-queues.test.ts`. 200 pass, `tsc -b` clean.

### Deliberate deviations

Each avoids an extension bug (README, "Where the backfill pass deliberately does not match the extension"):

- Progress writes merge. The extension's `set()` wiped the comparison fields, so its gate never held (#3012).
- Every chunk document, including the trailing partial batch, commits before the first dispatch. The extension committed each 50th, so 51–99 chunks stalled on `Next task document … does not exist`.
- Per-document failures stay aligned. The extension dropped rejected results and wrote one document's embedding onto another.
- A short batch result marks trailing documents `FAILED_BACKFILL` (extension: `BACKFILLED`, no vector).
- Phantom parents from `listDocuments()` are skipped (extension: threw and retried).
- The update pass records all four comparison fields (extension: only `embeddingProvider`).
- Eligible documents are collected inside the transaction (extension: outside, so retries double-counted).

Kits have no `setProcessingState`; progress is logs plus the progress document (#2974 Notes). Task ids are `kit-<instance id>-task-N` (extension: `ext-`).

Out of scope: §10 lifecycle wiring. Pre-existing, documented in the README, pending a ledger entry: `DO_BACKFILL=false, UPDATE_ON_CONFIGURE=true` runs the update pass where the extension ran nothing.

## Testing

**Unit.** `backfill.test.ts` covers the gate (absent, unchanged, each field changed, held after counters), the thread (120 ids → 3 chunks; 2,600 → 52 chunks, all committed before dispatch; empty list), batch and single-document chunks, every skip rule, both failure paths, aligned failures, the short batch, counters and `DONE`, `Invalid task document`, a missing successor, and trigger gating.

**Deployed**, pre-merge branch (flat `status.state`; the status-path change is unit-tested). Codebase `fvs-pr3097`, `EMBEDDING_PROVIDER=vertex`, 60 seeded documents (55 plain; one each `BACKFILLED`, `COMPLETED`, `ERROR`, non-string, empty). Deleted afterwards.

- First deploy: 60 enumerated; both enqueue documents committed before `task-1` dispatched; `task-1` (50) → `task-2` (10), both `DONE`. 56 `BACKFILLED` with `completeTime` and a 768-dimension vector; 4 skipped with the extension's log lines. Progress `{total 60, processed 56, skipped 4, failed 0, DONE}`, comparison fields intact.
- Unchanged redeploy: `no pass required`; nothing enumerated or embedded (the #3012 case).
- Update pass (`DO_BACKFILL=false`, `UPDATE_ON_CONFIGURE=true`, metadata cleared): 61 documents in chunks of 50 and 11; 56 processed, 5 skipped, `DONE`.
- `embedOnWrite` unchanged. Pre-existing instance: gate held across two unchanged redeploys, reopened on an `INPUT_FIELD_NAME` change.

**Side-by-side parity, deployed** (run `e420bfab`, `corie-testing`, since torn down). Kit at `ea66370f0`, packed as `0.0.2-rc.2` and installed as `vske420bfab`, beside the published extension `googlecloud/firestore-vector-search@0.1.3` as `vsxe420bfab`. Both configured identically against one stub embedder (`EMBEDDING_PROVIDER=custom`, 16 dimensions, batch size 7, input `"BOOM"` returns 500) over isolated collections, so embeddings were deterministic and shared.

- Backfill over an identical 20-document fixture: 20/20 documents identical — same state, vector presence, `startTime`/`completeTime`, no extra status fields. Identical counters `{total 20, processed 7, skipped 9, failed 4}`, one chunk of 20 ids, task `<prefix>-task-1`, `DONE` on both.
- Every skip rule matched (`COMPLETED`, `ERROR`, `PROCESSING`, `FAILED_BACKFILL`, missing, empty and non-string input), as did the batch-failure grouping: `"BOOM"` fell in the second batch of 7, so documents 17–20 failed together on both sides.
- `embedOnWrite`: 5/5 identical.
- One divergence, the documented merge deviation: the extension's task thread replaced `_<instance id>/index` with the counters alone, so its gate reopened and re-ran the whole pass on every later trigger, while the kit merged, kept the comparison fields and logged `no pass required`.


CorieW:
Review of the changes in this PR against the extension. Three problems, all in the path this PR touches.

**1. The OpenAI client's batch size is 1, so `openai` still makes one API call per document.**

`OpenAiEmbedClient` calls `super(1)`, while the extension's `OpenAIEmbedClient` passes `batchSize: 16`. The new chunk runner batches at `EmbedClient.batchSize`, so with `EMBEDDING_PROVIDER: openai` a 50-document chunk still becomes 50 `embeddings.create` calls — the exact cost this issue is about, left in place for one provider. The kit's `getEmbeddings` already sends `input: [...inputs]`, so the array is handled; only the declared size was wrong. This is the §2 row's `batchSize 16 → 1` half; the model and dimension change stays with its own tracked item. It also makes the README's "16 documents per OpenAI call" claim false as written.

**2. The enqueue documents are committed after the first task is dispatched, so the first task can race its own successor.**

`enqueueTaskThread` dispatches `task-1` inside the loop and only commits the enqueue documents at 50-chunk boundaries and after the loop. For any thread of 2 to 50 chunks, every enqueue document is now committed *after* `task-1` is already running, so `task-1` finishing quickly hits `Next task document … does not exist` and burns retries. The extension committed on every iteration when there were fewer than 50 chunks, which made the window smaller but did not close it. Recording every chunk before dispatching closes it.

**3. The `enqueues` subcollection is not documented.**

The README enumerates the internal Firestore paths this kit owns, and the restored task thread adds `_<instance id>/index/enqueues/<task id>`. The extension wrote the same documents, so this is a documentation gap rather than a behaviour difference, but the paths list should be complete.

CorieW:
Second review pass. One problem, introduced by this PR.

**With `DO_BACKFILL: true` and `UPDATE_ON_CONFIGURE: true`, the two passes collide on one progress document.**

`handleInit` enqueues both triggers when both settings are on, and it runs on `afterFirstDeploy` and on `afterRedeploy`. Before this PR that was harmless: each trigger enqueued its own per-document tasks and the update pass skipped documents with no existing embedding. Now both triggers share `_<instance id>/index` as their task thread, and both mint the same task ids (`kit-<instance id>-task-N`), so two concurrent threads overwrite each other's enqueue documents and interleave their counter increments. The counters then never reach `backfillJobsTotal`, and both threads keep dispatching until one of them dies on a chunk the other rewrote.

The extension never hit this because it ran the backfill on install and the update on configure, never both at once, so a single metadata document was enough.

`handleInit` now enqueues the update trigger only when it has not already enqueued the backfill trigger. The backfill pass is a strict superset of the update pass — the update pass adds "and already has an embedding" to the same eligibility rule and writes the same fields — so nothing is lost by letting the backfill win, and either setting on its own behaves exactly as before.

Also worth stating plainly in the description, since it is a difference a reviewer comparing against the extension will look for: **both passes record the full metadata set.** The extension's `updateOptions.metadata` carried only `embeddingProvider`, so an update pass wrote a document missing three of the four fields the gate compares, and the next gate check could never hold. The same four fields are written by both passes here, for the same reason the write merges.

CorieW:
Third review pass. Two small problems, both in the surface this PR adds; no behaviour differences left against the extension.

**1. `./lib` exports `BackfillProcess`, whose signatures reference an unexported type.**

`BackfillProcess.shouldBackfill` / `processFn` / `batchFn` are all typed in terms of `DocumentData`, declared in `src/backfill.ts` and not re-exported from `lib.ts`, so a consumer implementing the interface cannot name its own parameter types. The name is also the same as Firestore's own `DocumentData`, which would collide on import in exactly the code that needs both. Renamed to `BackfillDocumentData` and exported.

**2. "Delete `_<instance id>/index` to force a full re-embed" reads as broader than it is.**

That document is also the parent of the `queries` subcollection your clients write to. Deleting a document does not delete its subcollections, so the advice is safe, but it should say so rather than leave the reader to work it out.

Also folded in: `"PROCESSING"` is written to the enqueue documents but was missing from the `BackfillJobStatus` union, so that one write was the only status write in the module not checked against it.

kits/firestore-vector-search/src/backfill.ts
```diff
@@ -0,0 +1,506 @@
...
+      pendingWrites = 0;
+    }
+
+    counter++;
+  }
+
+  if (pendingWrites > 0) {
+    logger.info("Committing the batch...");
+    await writer.commit();
+  }
```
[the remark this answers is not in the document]
CorieW: Already fixed. This review ran against the first push (81c15a0f); the enqueue documents have been committed before the first dispatch since 437cce5c, and it is covered by `records every chunk before dispatching the first task`.

kits/firestore-vector-search/src/backfill.ts
```diff
@@ -0,0 +1,506 @@
...
+}
+
+export function getNextTaskId(prevId: string, instanceId: string): string {
+  const pattern = new RegExp(`^kit-${instanceId}-task-\\d+$`);
+  if (!pattern.test(prevId)) {
+    throw new Error(`Invalid task ID format: ${prevId}`);
+  }
+  const taskNum = prevId.split("task-")[1];
+  return taskIdFor(instanceId, Number.parseInt(taskNum, 10) + 1);
+}
```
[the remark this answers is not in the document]
CorieW: Fixed in cb1a886f. Captures the digits from the same anchored pattern instead of splitting on `task-`, with a test using instance id `my-task-force`.

kits/firestore-vector-search/src/backfill.ts
```diff
@@ -0,0 +1,506 @@
...
+          `Document ${doc.ref.path} is not in the correct state to be backfilled`
+        );
+        continue;
+      }
+
+      validDocuments.push(doc);
+    }
+  });
+
+  return { validDocuments, skippedDocuments };
```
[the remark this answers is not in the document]
CorieW: Fixed in cb1a886f, collected inside the callback and returned from it. Worth noting the extension has the same shape, but a retry here would overstate the chunk's counts and stall the thread, so this one is worth diverging on.

kits/firestore-vector-search/src/backfill.ts
```diff
@@ -0,0 +1,506 @@
...
+  logger.info(
+    `Current state: ${processed} processed, ${totalSkipped} skipped, ${totalFailed} failed out of ${totalTasks} total tasks`
+  );
+
+  if (processed + totalSkipped + totalFailed === totalTasks) {
+    await firestore.doc(tasksDoc).update({
+      backfillStatus: "DONE" satisfies BackfillJobStatus,
+    });
+    return;
+  }
```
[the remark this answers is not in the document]
CorieW: Not taking this one: it would reduce parity. The extension reads the counters, writes `FieldValue.increment`, and compares locally; the suggestion replaces the increments with absolute writes in a transaction. This PR is a parity restoration (#2974 §9), and duplicate Cloud Tasks delivery overshoots the counters upstream too, so it is not a regression introduced here. Worth an issue against both if we want it fixed.

cabljac:
Rebased onto `kits` at d97ce5ee with Corie's permission. Five original commits kept with their authorship; one commit appended at the end.

What changed in the rebase:

- `CHANGELOG.md` and `README.md`: textual conflicts with #3094 (events bullet) resolved by keeping both sides. The intro sentence now reads "Multimodal embedding and the shape of the status field ... changed", which is the `kits` wording minus "the backfill", since this PR removes that divergence.
- `src/embeddings/client/text/open_ai.ts` and `tests/embeddings.test.ts`: dropped from the branch. #3096 already landed batch size 16 and `text-embedding-ada-002` on `kits`, and its OpenAI suite covers the three tests this branch added, one of which asserted the old `text-embedding-3-small` at 512. The CHANGELOG line about raising the batch size went with it.
- `tests/task-queues.test.ts` (new commit 395d7463): this test landed on `kits` after the branch was cut. Its Firestore stub had no `doc()`, `listDocuments()` or `batch()`, and it expected `handleInit` to enqueue both triggers when both settings are on, which the one-pass rule in this PR changes. The stub is extended, the init case asserts the backfill trigger alone, and a second case pins the update trigger on its own.

Verified on the rebased head: 159 tests pass (113 on `kits` plus the 45 in `backfill.test.ts` and the one new case), `tsc --noEmit` clean, prettier 2.8.8 clean on every changed file. The PR body's description of the OpenAI change and its test counts are now stale by that much; the deployed run it describes was against the pre-rebase code and was not repeated.

kits/firestore-vector-search/README.md
```diff
@@ -198,35 +199,40 @@ whatever `EMBEDDING_PROVIDER` is set to. If either does not exist, `firebase
...
-*n* embedding calls rather than *n*/batch size.
-
-There is also no install-time progress reporting, since there is no extension
-install UI to report into. Watch the function logs instead.
+### `UPDATE_ON_CONFIGURE` is read, and the backfill gate is stricter than the extension's
+
+This setting was declared by the extension but never read: its update pass was
+gated on `DO_BACKFILL` instead. The kit reads `UPDATE_ON_CONFIGURE`, so the two
+passes are controlled independently — `DO_BACKFILL` after the first deploy,
+`UPDATE_ON_CONFIGURE` after every redeploy.
```
cabljac, line 207: I think this sentence doesn't match what the code does. `initVectorSearch` runs on both `afterFirstDeploy` and `afterRedeploy` (`src/index.ts:118-123`), and `handleInit` returns after enqueuing the backfill whenever `doBackfill` is on (`src/handlers.ts:207-212`), so with `DO_BACKFILL=true` it is the backfill trigger that gets enqueued on every redeploy too, and `UPDATE_ON_CONFIGURE` never gets a look in. Lines 301-302 already say "after your first deploy and after every redeploy", so the two paragraphs contradict each other.

That also makes the one-pass rule a deviation worth naming. The extension maps install to `backfillTrigger` and configure to `updateTrigger` (`extension.yaml:400-406`), and the update pass only touches documents that already have an embedding, so a reconfigure that changed `INPUT_FIELD_NAME` left never-embedded documents alone. Here it backfills them as well. Probably the right call given the shared thread, but it changes which documents get written on a redeploy, so it belongs in the Differences list rather than only in the `handleInit` comment.

One more thing this section should say, because "redeploying without changing anything enqueues nothing" is only true from the second kit deploy. On a migrated instance the first deploy re-embeds the whole collection: the extension's `_<instance id>/index` document holds only the progress counters after any pass that enqueued tasks (its `set()` replaced the config fields), so the gate opens, and the pass then treats every extension-written document as unprocessed because it reads `status.state` (`src/backfill.ts:470-472`) where the extension wrote `status.<instance id>.state`. That is a bigger bill than the extension's reconfigure, which skipped documents the write trigger had marked `COMPLETED` and re-embedded only the ones its backfill had marked `BACKFILLED`. With `DO_BACKFILL=true` it lands on the first kit deploy, so people should expect it. Worth checking my reading of the status path there, since that is the #3011 shape and I may be behind on where that landed.

Could you reword this paragraph to what actually happens (both settings act on every deploy, backfill wins when both are on), and add the install/configure mapping and the first-deploy re-embed to the Differences list?

kits/firestore-vector-search/README.md
```diff
@@ -198,35 +199,40 @@ whatever `EMBEDDING_PROVIDER` is set to. If either does not exist, `firebase
...
+The two passes share that document as their task thread, so only one of them
+runs per deploy: with both settings on, the backfill pass runs, which covers
+every document the update pass would have (the update pass is the same
+eligibility rule plus "and already has an embedding").
+
+The extension's gate did not survive its own first pass, because the progress
+counters it wrote to the same document replaced the recorded configuration. The
+kit merges instead, so the comparison fields persist and the gate holds on every
+later deploy. To force a full re-embed without changing any setting, delete the
+`_<instance id>/index` document; its `queries` subcollection is untouched, so
```
cabljac, line 224: This section names two deviations (the merge, and the one-pass rule), but reading `backfill.ts` against the helper package I count more, and the rule is that every one is documented here so a reader can decide whether it matters to them. The ones I can see:

- Every enqueue document is committed before the first task is dispatched, and the remainder is always committed (`backfill.ts:185-216`). The extension dispatched task-1 before its own enqueue document was in the batch (`trigger.ts:70-83`) and never committed the chunks past the last multiple of 50 (`trigger.ts:89`), so 51 to 99 chunks lost the tail.
- Update-pass failures stay on their own document (`backfill.ts:431-438`, `392-398`). The extension dropped rejected results and shifted the rest, so one document could receive another's embedding.
- A batch result shorter than its input marks the trailing documents `FAILED_BACKFILL`. The extension marked them `BACKFILLED` with no new vector. Side effect of the fix above, but observable.
- A document id with no document (a subcollection parent) is skipped (`backfill.ts:462`). The extension's `shouldBackfill` threw a `TypeError` there and the task retried until the queue gave up.
- The update pass records all four comparison fields on the metadata document (`handlers.ts:324-333`). The extension recorded only `embeddingProvider` for that pass, so even without the `set()` wipe the other three comparisons would have read `undefined` and its gate would have opened on every reconfigure.
- `getValidDocs` collects inside the transaction callback (`backfill.ts:453-484`). The extension accumulated outside it, so a transaction retry double-counted.

All of these look justified to me under "matching would reproduce a bug", and the PR body already argues most of them. Could you add each to this list with one line on what the extension did? The `handleInit` comment and the PR body are not where a user will look.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -197,6 +206,10 @@ export async function handleInit(ctx: HandlerContext): Promise<void> {
 
   if (ctx.config.doBackfill) {
     await enqueueBackfillTrigger(ctx);
+    // The two passes share one task thread on the index metadata document, and
+    // the backfill pass covers every document the update pass would. Running
+    // both at once would have them overwrite each other's progress.
+    return;
   }
   if (ctx.config.updateOnConfigure) {
     await enqueueUpdateTrigger(ctx);
```
cabljac, line 215: One deviation here that I think is unclaimed, and it only half holds. The extension's install always ran `backfillTrigger`, and `updateOrCreateMetadataDoc` created the document with the full config before checking `DO_BACKFILL` (`metadata_document.ts:64-69`, called from `firestore_process_backfill_trigger.ts:14-23`), so with `DO_BACKFILL=false` the document still existed. Flipping the setting to true later and reconfiguring then found a matching document and did not backfill. Here nothing is enqueued when both settings are off, so no document is written and flipping `DO_BACKFILL` on later does run the pass. With `UPDATE_ON_CONFIGURE=true` though, the update trigger writes the document on the first deploy and the flip is gated out exactly as on the extension. Worth a sentence in the Differences list either way, and the "delete the metadata document to force a pass" advice covers the trap. Worth double-checking my reading of the extension's ordering there.

cabljac (commented):
Rebased this onto `kits` earlier today (see the comment above for what changed). The code side looked good to me and I have no code changes to ask for; the three inline comments are all about the README describing what the code actually does, since the Differences section is where a migrating user will look.

Fair warning: the parity comparison against the extension and the helper package was AI-led, and I have only spot-checked it. If any of the extension behaviour I describe does not match what you saw when you ported it, push back and I'll go and look properly.

kits/firestore-vector-search/README.md
CorieW: Agreed, the README was wrong. The kit registers `initVectorSearch` for both `afterFirstDeploy` and `afterRedeploy`, and `handleInit` runs the backfill whenever `DO_BACKFILL=true`, so the update pass never runs while backfill is on. The extension mapped install to `backfillTrigger` and configure to `updateTrigger` (`extension.yaml:400-406`). Fixed in 4bc459d32: the `UPDATE_ON_CONFIGURE` section now says both settings act on every deploy and names the extension mapping, and a new section covers the first deploy over a migrated instance (the kit reads `status.state`, the extension wrote `status.<instance id>.state`, so the pass re-embeds everything once). CHANGELOG gets one line for that re-embed.

kits/firestore-vector-search/README.md
CorieW: Agreed. The PR body argued these but the README listed only two. 4bc459d32 adds a section listing all six: commit-before-dispatch with the remainder commit, per-document failure alignment, short batch results marked `FAILED_BACKFILL` (the extension spread `undefined` and marked them `BACKFILLED`), the phantom-parent skip, all four comparison fields on the update pass (the extension's `updateOptions.metadata` carried only `embeddingProvider`), and `getValidDocs` collecting inside the transaction. Each entry states what the extension did.

kits/firestore-vector-search/src/handlers.ts
CorieW: Confirmed against the extension: `updateOrCreateMetadataDoc` calls `createMetadataDoc` before `shouldRunBackfill()`, so the extension wrote the document on install whatever `DO_BACKFILL` was. In the kit nothing is enqueued when both settings are off, so no document exists and flipping `DO_BACKFILL` on later runs a full pass; with `UPDATE_ON_CONFIGURE=true` the update pass writes the document and the flip does nothing, matching the extension. Documented in 4bc459d32 (`UPDATE_ON_CONFIGURE` section), code left as is.

kits/firestore-vector-search/README.md
CorieW: Correction to my reply above, after merging `origin/kits` in c274e342b: the kit now reads and writes `status.<instance id>.state`, the same path as the extension and #3180, so the status-path mismatch no longer exists. The first kit deploy over an installed instance still runs a pass, because the extension's `set()` replaced the comparison fields, but the pass follows the extension's rules: it skips `COMPLETED` and `ERROR` documents and re-embeds `BACKFILLED` or status-less ones. README and CHANGELOG say so at ea66370f0. Resolving.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -297,94 +310,175 @@ export async function handleBackfillTrigger(
...
-  const input = data[ctx.config.inputFieldName];
-  // Matches the extension's `shouldBackfill` and `shouldUpdate`, which both
-  // required a truthy string, and keeps an empty document out of the terminal
-  // status that `handleEmbedOnWrite` would then skip forever.
-  if (typeof input !== "string" || input === "") return;
-  if (requireExistingEmbedding && !data[ctx.config.outputFieldName]) {
+  // Resolved first so a missing region fails the trigger rather than being
+  // logged and swallowed below.
+  const queue = taskQueue(ctx, taskQueueName);
+  const { path, shouldBackfill } = await updateOrCreateMetadataDoc(
```
cabljac: I think the order here means a failed enqueue is never retried. Walking it through:

1. :360 `updateOrCreateMetadataDoc` compares `embeddingProvider`, `dimension`, `inputField`, `outputField` on `_<instance id>/index` with the current config. They differ, so it writes the current values onto that document.
2. :375-396 `listDocuments` and `enqueueTaskThread` run. Say `queue.enqueue` throws (queue missing, IAM).
3. :397 the catch logs and returns. No tasks ran.
4. Next `firebase deploy`: step 1 reads the document, the four fields now equal the current config, so it logs "no pass required" and returns.

From here the pass only runs if the user changes one of those four settings or deletes `_<instance id>/index` by hand, and the only hint is one error line in the logs.

The extension had the same ordering, but its behaviour on this path was different. `taskThreadTrigger` (`trigger.js:39`) did `set()` with only the five counter fields, which deleted the four comparison fields, success or failure. So on the next reconfigure `shouldDoBackfill` read `embeddingProvider` as undefined and the pass ran again. That is the #3012 bug on the success path, but on the failure path it meant a failed enqueue was always retried on the next reconfigure. The merge in this PR keeps the fields, which fixes #3012, but it also removes that retry: the extension re-ran a failed pass, the kit never does. Worth confirming my reading of `trigger.js`.

What do you think the right move is? Two options I can see: write the four fields after `enqueueTaskThread` resolves so a failed enqueue leaves the old values in place, or rethrow from the catch and let Cloud Tasks retry (though a failure after `task-1` is dispatched would then start a second thread on the same task ids). With the first option every path is either identical to today or better than the extension, except one: if the `update({backfillStatus: "RUNNING"})` at `backfill.ts:217` fails after `task-1` is already dispatched, the next deploy repeats the pass once, which is what the extension did on every reconfigure anyway.

kits/firestore-vector-search/src/backfill.ts
```diff
@@ -0,0 +1,551 @@
...
+}
+
+function taskIdFor(instanceId: string, counter: number): string {
+  return `kit-${instanceId}-task-${counter}`;
+}
+
+export function getNextTaskId(prevId: string, instanceId: string): string {
+  // Captured rather than split on "task-", because an instance id may contain
+  // that substring itself.
+  const match = prevId.match(new RegExp(`^kit-${instanceId}-task-([0-9]+)$`));
```
cabljac, line 244: Nit: `instanceId` goes into the `RegExp` unescaped. `instanceIdFromEnv` does not validate the value, and I have not checked what characters the CLI allows in the `instances` key, so a `+` or `.` in an instance id may change the match. A string check avoids the question:

```ts
const prefix = `kit-${instanceId}-task-`;
const counter = prevId.startsWith(prefix) ? prevId.slice(prefix.length) : "";
if (!/^[0-9]+$/.test(counter)) {
  throw new Error(`Invalid task ID format: ${prevId}`);
}
```

Same behaviour for `my-task-force` at backfill.test.ts:267, no regex to escape.

kits/firestore-vector-search/src/backfill.ts
```diff
@@ -0,0 +1,551 @@
...
+      backfillJobsProcessed: 0,
+      backfillJobsSkipped: 0,
+      backfillJobsFailed: 0,
+      backfillStatus: "PENDING" satisfies BackfillJobStatus,
+    },
+    { merge: true }
+  );
+
+  const chunks = chunkArray(taskParams, TASK_CHUNK_SIZE);
+  if (chunks.length === 0) {
```
cabljac, line 191: Nit: with zero chunks this returns after writing `backfillStatus: "PENDING"` at :172, so the document reads as a pass that never started. `runTrigger` returns earlier for an empty collection, so I think this branch is unreachable in practice, and the extension did the same. Fine to leave. If you would rather not pin unreachable behaviour, the test at backfill.test.ts:449 could go with it, or the early return could move above the `set()`.

kits/firestore-vector-search/src/handlers.ts
CorieW: Confirmed. The comparison fields are merged before the enqueue, and `runTrigger` swallows enqueue errors, so a failed first enqueue closes the gate for good. The extension's replacing `set()` wiped the fields, so the next reconfigure retried.

Taking your first option: keep the read and comparison as they are, write the fields only after the thread is enqueued (and on the empty-collection path), and write nothing in the catch. Rethrowing would diverge from the extension and duplicate the thread on every retry once `task-1` is dispatched.

The residual race (`RUNNING` write failing after dispatch) is no worse than the extension's repeat pass and shares a root with the non-idempotent counters; tracking it in a follow-up. Adding a test for a rejected enqueue and a README line.

cabljac (commented):
(lgtm)

PR #3098, opened 2026-09-02
fix(kits): restore the extensions' event payload shapes

### LINKED ISSUE #3027: decision(kits): event payload shape {change, context} vs {data, params}
Cross-kit consumer break (firestore-counter §5, firestore-translate-text, speech-to-text error `{}` to `{message, stack}`). If this is intentional gen2 design, it closes as a Notes entry; if not, it reopens as per-kit fixes. One decision, several boxes in #2974.

### LINKED ISSUE #3125: kits: declare roles/eventarc.publisher so event publishing does not 403
Every kit that publishes Eventarc events declares `roles/eventarc.eventReceiver` and `roles/run.invoker` but not `roles/eventarc.publisher`:

- `kits/firestore-counter/src/index.ts:40`
- `kits/firestore-translate-text/src/index.ts:37`
- `kits/firestore-bigquery-export/src/index.ts:63`
- `kits/speech-to-text/src/index.ts:38`

The extensions never declared a publisher role either; they got publish rights on the channel from the Extensions install flow (the `events:` block in `extension.yaml`). Kits have no equivalent, so nothing grants it. #3098 reports every publish returning 403 in a test project with `EVENTARC_CHANNEL` set.

The consequence is worse than missing telemetry, because each handler awaits the publish inline:

- `firestore-counter` `src/handlers.ts:65` and `firestore-translate-text` `src/handlers.ts:64` await `recordStartEvent` as their first action, so with the channel set the function throws before it aggregates or translates anything.
- `firestore-bigquery-export` `src/handlers.ts:180` does the same before the BigQuery write.
- `speech-to-text` awaits `recordCompleteEvent` at `src/handlers.ts:198` after transcription, so the work is done but the invocation fails.

Fix: add `"roles/eventarc.publisher"` to `REQUIRED_ROLES` in each of the four kits (one line each, same mechanism as the roles already declared), and re-run the deploy test in #3098 to confirm a subscriber receives the payload. If the intent is for users to grant it themselves, the READMEs need to say so; none do today.

Tracked in #2974.

CorieW:
Closes #3027.
Closes #3125.

## Investigation

The decision in #3027 was whether the payload rename is intentional 2nd gen design. It is not, and nothing about 2nd gen forces it:

- A custom event's payload is whatever the kit hands `channel.publish` â€” `firebase-admin` just `JSON.stringify`s it. There is no 2nd gen wire format for custom events, so `{data, params}` is only the 2nd gen handler's local variable names leaking into the payload.
- `FirestoreEvent` carries every 1st gen `EventContext` field under a different name (`id`, `time`, `project`/`database`/`document`, `params`), so the 1st gen `context` is fully reconstructible.
- 1st gen `change` and 2nd gen `event.data` are the same `firebase-admin` `Change<DocumentSnapshot>` and serialize identically, so keeping the `change` key costs nothing.
- The kits were not just renaming: `context.eventId`, `context.timestamp`, `context.eventType` and `context.resource` were dropped from the payload entirely, so this was a regression on top of a consumer break.

For `speech-to-text`, the extension published the caught `Error` itself. The payload carries whatever enumerable fields the thrown error has: a plain `Error` gives `{"error":{}}` because `message` and `stack` are not enumerable, a Storage `ApiError` gives `code`, `errors`, `response` and `message`; the kit published `{message, stack}` instead. That is an upgrade, not parity, and it also dropped the `name` that `errorFromAny` puts on a thrown non-error. Restored to parity, with the improvement left to be tracked separately.

## Changes

- Added `src/event-context.ts` to `firestore-counter` and `firestore-translate-text`, with `toEventContext` rebuilding the 1st gen `EventContext` (`eventId`, `timestamp`, `eventType`, `resource.service`/`resource.name`, `params`) from a `FirestoreEvent`. `params` carries the trigger wildcards, and `notSupported` is republished as the empty placeholder the 1st gen body carried. Each kit's `event-context.test.ts` feeds the raw 1st gen event body the backend actually sends, so it pins the branch that runs in production. `eventType` is the 1st gen constant `google.firestore.document.write` that every 1st gen Firestore `onWrite` trigger reported.
- `firestore-counter` `handleShardWrite` publishes `onStart` as `{change, context}` and `onCompletion` as `{context}` again, instead of `{data, params}` and `{params}`.
- `firestore-translate-text` `handleDocumentWrite` does the same, across all three `onCompletion` call sites.
- `speech-to-text` `recordErrorEvent` publishes `{ error }` again instead of `{ error: { message, stack } }`.
- Dropped the "Event payloads have a different shape" note from the `firestore-counter` and `firestore-translate-text` READMEs and the "`fail` events for unexpected errors now say what went wrong" note from the `speech-to-text` README, and recorded the payloads under each kit's "Unchanged" list instead.
- Tests: added `event-context.test.ts` to both kits; pinned the rebuilt payloads in the `handlers` and `events` suites, including a JSON round-trip assertion on what a subscriber reads off the wire; added a `speech-to-text` case for the thrown-non-error payload keeping its `name`, and reworked the raw-`Error` case to assert the `{"error":{}}` the extension produced.

## Testing

Unit: 81 tests in `firestore-counter`, 94 in `firestore-translate-text`, 34 in `speech-to-text`, all passing. `tsc -b` clean for all three, and `prettier@2.8.8 --list-different` clean over every file touched. The new wire-level assertion was checked as non-vacuous by breaking `eventType` and confirming it fails.

Deployed: all three kits were `npm pack`ed into `project/function-kits/*/source/vendor` and the installed bundles confirmed to contain the new code. The four overlapping `kit-firestore-translate-text-*-fstranslate` functions in `corie-testing` were deleted and confirmed gone from `functions:list` before deploying `firestore-counter-a91f6c2e`, `firestore-translate-text-a91f6c2e` and `speech-to-text-a91f6c2e`.

Payload, from real platform events (captured before 53cccfa33, so `params` below is populated; it is now `{}`, see Changes). Eventarc publishing 403s in this project, so a subscriber cannot see the payload (see below); instead a temporary codebase held Firestore triggers on the same document patterns, running the kit's own compiled `toEventContext` on the real events. `firestore-counter`, on a shard write:

```json
{"change":{"before":{...},"after":{"_fieldsProto":{"counter":{"integerValue":"1"}},...}},
 "context":{"eventId":"84b2826b-9cbb-488c-8db3-098edb1b02b0",
            "timestamp":"2026-09-02T14:53:29.897039Z",
            "eventType":"google.firestore.document.write",
            "resource":{"service":"firestore.googleapis.com",
                        "name":"projects/corie-testing/databases/(default)/documents/fxkits_range_counters/page-1788360801506/_counter_shards_/0000"},
            "params":{"collection":"fxkits_range_counters","counter":"page-1788360801506","shardId":"0000"}}}
```

`firestore-translate-text`, on a document write: same five `context` fields, `resource.name` `projects/corie-testing/databases/(default)/documents/fxkits_range_tr_gemini/smoke-1788360897518`, `params` `{"messageId":"smoke-1788360897518"}`.

Indirectly affected behaviour, with the channel unset as it was before: the translate handler still translates end to end â€” `{"input":"hello world","translated":{"en":"hello world","es":"hola mundo"}}` â€” and the counter still aggregates, the shard write landing as `{"counter":1}` on the counter document.

`speech-to-text`: its changed path is the catch-all handler, provoked by finalizing an `audio/wav` object and deleting it so the download throws inside the `try`. `ApiError: No such object: â€¦` was logged from `logs.error(error)` and the invocation completed, so the error path is intact. The payload itself is covered only by the unit round-trip, for the reason below.

Afterwards the temporary codebase and its functions were deleted, the three other `firestore-translate-text` instances redeployed, and the smoke documents removed.

## Out of scope

- Eventarc publishing 403'd in `corie-testing` because the kits declared `roles/eventarc.eventReceiver` but no publisher role (#3125). Fixed on this branch in f69b065a1: every event-publishing kit now declares `roles/eventarc.publisher`, with the README role tables and a regression test updated.
- `firestore-vector-search` publishes `{params}`, but the extension published nothing at all from that path (it declared the event types and never used them), so there is no extension payload to be at parity with.
- The `change` snapshots serialize with `firebase-admin` internals in both the extension and the kit, but the kit is on `firebase-admin@14` where the extension was on `@12`, so the internal field set differs slightly. That is an SDK-upgrade artifact this PR does not change â€” before it, the same snapshots were serialized under `data`.

## Review follow-ups

- 53cccfa33, then reverted in c5f7ee5e3: publishing `context.params` as `{}` was wrong. A live side-by-side deploy showed the gen1 backend fills `params` from the `extension.yaml` wildcards before `_makeParams` is reached, so the extensions did publish them. Both kits publish `event.params` again.
- 3050dce7d: both Firestore kits also publish `context.notSupported: {}`, the inherited empty placeholder from the 1st gen event body.
- cd86f7871: speech-to-text README and `recordErrorEvent` docstring describe the payload as the error's enumerable fields; `ApiError`-shaped test added.
- fc1df5c96: test tidy-up (counter `makeEvent` helper, literal expected resource, missing `event.time` case, wire-test comment) and the translate README event count.


CorieW:
Review of this PR's own changes — finding 1 of 2, `firestore-counter` test fixtures misrepresent the trigger.

The event fixtures I added in `tests/handlers.test.ts` and `tests/event-context.test.ts` use `document: "_firebase_ext_/sharded_counter"` with `params: { shardId: "0000" }`. That is the internal-state path, which is what the `worker` trigger watches. `handleShardWrite` serves `onWrite`, whose document pattern is `{collection}/{counter=**}/_counter_shards_/{shardId}` (`src/index.ts:89`, matching the extension's `extension.yaml:72`), so a real event carries a shard-document path and `params: { collection, counter, shardId }`.

The assertions pass either way, so the reconstruction is not wrong — but the fixtures pin a `resource.name` and a `params` map that no subscriber ever receives, which is exactly the detail the fix is about. Replacing them with `pages/home/_counter_shards_/0000` and the three real wildcards.

CorieW:
Review of this PR's own changes — finding 2 of 2, no test asserts the serialized `firestore-counter` / `firestore-translate-text` payloads.

The whole point of the fix is what a subscriber reads off the wire, and `firebase-admin` puts the payload on the wire as `JSON.stringify(data)`. The `speech-to-text` cases I added assert that serialized form, because the empty `{"error":{}}` only shows up after a round-trip. The counter and translate cases do not: `handlers.test.ts` mocks `../src/events` entirely, and `events.test.ts` asserts the object handed to `channel.publish` before it is ever stringified.

That leaves the parity claim in the READMEs — `context.eventId`, `context.timestamp`, `context.eventType`, `context.resource`, `context.params` all reaching subscribers — unpinned at the level it is made. Adding a JSON round-trip assertion to both kits' `events.test.ts`.

CorieW:
Review of this PR's own changes — finding 3, an unrelated reformat slipped into `firestore-counter/tests/handlers.test.ts`.

The `vi.mock("../src/controller")` block picked up a hunk that has nothing to do with this PR:

```diff
-  const actual = await vi.importActual<typeof import("../src/controller")>(
-    "../src/controller"
-  );
+  const actual =
+    await vi.importActual<typeof import("../src/controller")>(
+      "../src/controller"
+    );
```

That is prettier 3 output. The repo pins `prettier` at `2.8.8`, and the pre-commit hook shells out to a bare `prettier`, which in a worktree with no root `node_modules` resolved to a global 3.7.4 instead. Reverting the hunk; `prettier@2.8.8 --list-different` over every file this PR touches now reports only that one file, and nothing after the fix.

cabljac (commented):
The investigation in the description is the right shape - checking what v1 actually emits rather than trusting the yaml, and pinning it with a deploy, is what makes this reviewable. `eventType`, `resource.name` and the `time`/`id` mapping all check out against `firebase-functions@4.9.0`.

Two things I'd want resolved before merge: the publisher role (none of these three kits can publish at all today, and the README now tells users these payloads arrive), and the `params` claim, which I think is wrong in a way the README repeats.

The rest are smaller. Detail inline.

Note on provenance: this was an AI-led review, run against the branch and the original extension sources, and I've read the findings rather than re-derived every one of them myself. The file:line references and the SDK behaviour claims are worth checking against the source before you act on them, particularly the `params` one. Push back where it's wrong.

kits/firestore-counter/README.md
```diff
@@ -201,8 +191,11 @@ carry on getting them from the extension repo.
 - The aggregation behaviour: inline aggregation up to 200 shards, workers above
   that, 45 second self-scheduling worker runs, partial shard cleanup, and
   deletion of shards once they are summed into the counter field.
-- The three functions and the event types they publish, aside from the worker
-  and payload points above.
+- The three functions, the event types they publish and their payloads:
+  `onStart` still carries `{change, context}` and `onCompletion` still carries
+  `{context}`, with `context.eventId`, `context.timestamp`, `context.eventType`,
+  `context.resource` and the trigger wildcards under `context.params`. Aside
+  from the worker point above.
```
cabljac, line 200: Anchoring here because `src/index.ts` isn't in the diff, but this is the finding I'd most want closed before merge.

All three kits declare `roles/eventarc.eventReceiver` and no `roles/eventarc.publisher` (`firestore-counter/src/index.ts:40`, `firestore-translate-text/src/index.ts:36`, `speech-to-text/src/index.ts:38`). I think that's why you saw the 403 on every publish.

It isn't only that the payload is invisible. `handleShardWrite` awaits `recordStartEvent` as its first action (`handlers.ts:67`), so a user who sets `EVENTARC_CHANNEL` gets a function that throws before it aggregates anything. Same for translate at `handlers.ts:75`.

That matters more now than it did before this PR, because this bullet tells users these payloads arrive. Right now they don't arrive at all.

I'd add the publisher role (one line per kit, same class as the roles already declared) and re-run the deploy test. That should also get you the subscriber-side view you couldn't capture. If there's a reason the Extensions install flow is meant to grant this rather than `extension.yaml`, say so and let's file the issue and caveat the README, rather than leaving it unlinked.

Small thing on the same bullet: it ends on the fragment "Aside from the worker point above."

kits/firestore-counter/src/event-context.ts
```diff
@@ -0,0 +1,64 @@
...
+): EventContext {
+  return {
+    eventId: event.id,
+    timestamp: event.time,
+    eventType: FIRESTORE_WRITE_EVENT_TYPE,
+    resource: {
+      service: FIRESTORE_SERVICE,
+      name: `projects/${event.project}/databases/${event.database}/documents/${event.document}`,
+    },
+    params: event.params,
```
cabljac, line 80: I don't think the extension published any params here, so "still ... under `context.params`" in the README isn't parity, it's a superset.

In v1, `context.params` isn't supplied by the platform. The SDK computes it from the *code-side* trigger path (`_makeParams`, `firebase-functions/lib/v1/cloud-functions.js:150`). The extensions registered `document(process.env.INTERNAL_STATE_PATH)` (`firestore-counter/functions/src/index.ts:80`) and `document(process.env.COLLECTION_PATH)` (`firestore-translate-text/functions/src/index.ts:47`). Neither has a `{wildcard}` segment, so `WILDCARD_REGEX` matches nothing and `params` comes out `{}`. The yaml wildcards registered the trigger, they never reached the SDK.

Worth verifying yourself before you change anything, since you already have the deploy set up. Harmless either way, a superset breaks nobody, but the README and the description both claim a parity I don't think holds.

The more general point: the deploy test ran the kit's own `toEventContext` over 2nd gen events, so it can only confirm the kit is self-consistent. Nothing in it observes a real 1st gen context. If you can get one out of the emulator against the old extension, commit it as a fixture and assert `event-context.test.ts` against it. Then the parity claim is pinned rather than argued.

kits/firestore-counter/src/event-context.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Rebuilds the 1st gen `EventContext` from a 2nd gen `FirestoreEvent`.
+ *
+ * The extension handed the 1st gen handler's `context` straight to Eventarc, so
+ * subscribers read `eventId`, `timestamp`, `eventType`, `resource` and `params`
+ * off it. `FirestoreEvent` carries the same information under different names,
+ * so the published payload keeps its original shape instead of following the
+ * 2nd gen handler signature.
+ */
+export function toEventContext(
```
cabljac, line 69: On `Closes #3027`: `kits/firestore-bigquery-export/src/handlers.ts:185` still publishes the 2nd gen event under the 1st gen `context` key (`const { data, ...context } = event` at `:130`, `context.id` at `:152`). You call this out under "out of scope", which is a fair scoping call on its own, but #3027 reads as a cross-kit decision, so closing it with one kit still on the broken shape leaves it half-applied and untracked.

This helper is what that kit needs, so either pull it in here, or open the follow-up and link it from #3027 and #2974 before this merges. Either works, but not neither.

kits/speech-to-text/tests/events.test.ts
```diff
@@ -83,17 +83,26 @@ describe("events", () => {
...
       type: "firebase.extensions.storage-transcribe-audio.v1.fail",
-      data: {
-        error: { message: "kaboom", stack: err.stack },
-      },
+      data: { error: err },
+    });
+    // Parity with the extension: `message` and `stack` are not enumerable, so
+    // subscribers receive `{"error":{}}` for a genuine `Error`.
+    const payload = publish.mock.calls[0][0];
+    expect(JSON.parse(JSON.stringify(payload)).data).toEqual({ error: {} });
```
cabljac: The `{"error":{}}` framing holds for `new Error(...)`, but I don't think it's what this path actually produces.

The catch-all at `handlers.ts:199` wraps `remoteFile.download()` (`handlers.ts:133`), which throws `ApiError` from `@google-cloud/storage`. `code`, `errors`, `response` and `message` are all enumerable own properties on it, so a subscriber gets something like `{"error":{"code":404,"errors":[...],"message":"..."}}`. Your own deploy log shows `ApiError: No such object`, so that's the realistic case.

Parity with the extension is unaffected either way, it published the raw error too. But this test and the README at `:217-221` both describe the rare case as the general one. I'd reword to something like "carries whatever enumerable fields the thrown error has; a plain `Error` gives `{}`", and add an `ApiError`-shaped case so the realistic payload is pinned too.

One thing I did check: the response object isn't a leak risk, `teeny-request` gives it a `toJSON` that returns headers only.

kits/firestore-counter/tests/events.test.ts
```diff
@@ -119,12 +135,41 @@ describe("event publishing", () => {
...
+    await events.recordCompletionEvent({ context });
 
     expect(publish).toHaveBeenCalledWith({
       type: "firebase.extensions.firestore-counter.v1.onCompletion",
-      data: { params: { shardId: "0000" } },
+      data: { context },
+    });
+  });
+
+  test("puts the whole 1st gen context on the wire", async () => {
```
cabljac, line 141: A few small ones, none blocking:

- This test mocks `firebase-admin/eventarc`, so the real serializer (`toCloudEventProtoFormat`) never runs, and `recordStartEvent` is a pass-through. It ends up being the `event-context.test.ts` assertion again with a hand-rolled `JSON.stringify`. Not wrong, but the comment claims more than it proves. Either drop it, or publish something where JSON semantics actually differ from `toEqual` (an error-bearing payload would).
- No failure-mode coverage. `strict` is off, so `timestamp: undefined` compiles fine and the key just vanishes from the wire. One test for a missing `event.time` would catch that.
- `SHARD_WRITE` here (`:21-28`) and the fixture in `event-context.test.ts:25-32` are both `as any`. Translate centralises this in `makeEvent` (`tests/helpers.ts:102`), still a cast but only in one place. Worth giving counter the same.
- `expectedEventContext` (`tests/helpers.ts:122`) rebuilds the `projects/.../databases/.../documents/...` template at `:131` from the same parts the source uses, so the same mistake in both places passes. A literal expected string is stronger.
- `firestore-translate-text/README.md:216` says "The four event payloads:" and then describes two.

kits/firestore-counter/src/event-context.ts
cabljac: Withdrawing this one. I went back and checked the extension itself rather than the description: `firestore-bigquery-export` is already a 2nd gen function (`onDocumentWritten`, declared as `v2function` in its `extension.yaml`), and it builds the payload the same way the kit does, `const { data, ...context } = event` at `functions/src/index.ts:156`. So the `context` its subscribers get today is the 2nd gen event, and the kit matches it exactly.

That means there is nothing to pull in here, and applying `toEventContext` there would break parity rather than restore it. Your "out of scope" note describes it as the same class of break, but it isn't one, since the extension never published a 1st gen `EventContext` from that path. Drop that paragraph from the description so nobody picks it up as a follow-up. Closing #3027 with this PR is fine as is.

Sorry for the detour - I took the description's framing on trust instead of reading the extension first.

kits/firestore-counter/src/event-context.ts
cabljac: One precision after reading `_makeParams` properly: the only way the extension's `params` was non-empty is if a user put their own `{wildcard}` segments inside `COLLECTION_PATH` itself (the validation regex allows braces, and the code-side path is that env var verbatim). Even then `messageId` never appears, because it only exists in the yaml resource. So the exact statement is "empty by default, and never the yaml wildcards". The rest of the comment stands.

kits/firestore-counter/README.md
CorieW: Confirmed: the deploy test hit a 403 on every publish, and the cause was the missing `roles/eventarc.publisher` (kits declared only `eventReceiver`; the Extensions platform granted publish implicitly from the `events:` block). Fixed in f69b065a169871a83fe89cb68d459e63756bb2e9: the role is declared in all eight event-publishing kits, documented in each README role table, and pinned by a regression test in `firestore-translate-text/tests/index.test.ts`. This also resolves #3125. Resolving.

kits/firestore-counter/src/event-context.ts
CorieW: Update: we changed this to `params: {}` in 53cccfa33, then a real side-by-side deploy showed that was wrong.

`_makeParams` is only a fallback (`src/v1/cloud-functions.ts:434`: `context.params = context.params || _makeParams(...)`). On a live 1st gen trigger the backend has already filled `params` from the `extension.yaml` wildcards, so the fallback never runs. Deploying `firebase/firestore-counter@0.2.17` next to the kit on the same shard write: the extension published `params` with `collection`, `counter`, and `shardId`; the kit published `{}`.

Fixed in c5f7ee5e3: both kits are back to `params: event.params` (the kit triggers declare the same wildcard names as the yaml), the docstrings and READMEs are corrected, and the oracle test now feeds the raw 1st gen event body with `params` present instead of exercising the fallback.

kits/speech-to-text/tests/events.test.ts
CorieW: Confirmed: a Storage `ApiError` has enumerable `code`, `errors`, `response` and `message`, so `{}` is only the plain `Error` case, not the general one. Rewording the README and the `recordErrorEvent` docstring to "whatever enumerable fields the thrown error has; a plain `Error` gives `{}`" and adding an `ApiError`-shaped test that asserts those fields survive the JSON round-trip. Landing in the next push.

Resolved in cd86f7871fc9d4cf0b0b47655f1e56915fc74cb2.

kits/firestore-counter/tests/events.test.ts
CorieW: All taken:

- Wire test: trimming the comment to what it proves (JSON round-trip of the built context; `firebase-admin/eventarc` is mocked).
- Adding a missing `event.time` case in both kits' `event-context.test.ts` so a vanishing `timestamp` key is caught.
- Giving counter a `makeEvent` helper in `tests/helpers.ts`, replacing both `as any` fixtures.
- `expectedEventContext`: literal expected resource name instead of the rebuilt template.
- Translate README: "four" becomes `onStart` and `onCompletion`.

Resolved in fc1df5c96aee6c322a43f48e65a0a41a042b55d8.

PR #3100, opened 2026-09-02
fix(storage-resize-images): restore the extension's default for an omitted deleteOriginal

### LINKED ISSUE #3024: fix(storage-resize-images): restore programmatic defaults for deleteOriginal and isAnimated
Unset `deleteOriginal` flipped `onSuccess` to `never` and unset `isAnimated` flipped `false` to `true`, a silent behavior change for programmatic callers.

Parity ledger: #2974, storage-resize-images §1-2.

cabljac:
The kit's resolver diverged from the extension only in the programmatic path. A CLI deploy writes DELETE_ORIGINAL_FILE=false, so both resolve to never-delete; an empty env value resolves to delete-on-success in both; but an omitted deleteOriginal resolved to never-delete in the kit where the extension's parser gave delete-on-success. This removes the kit-only 'case undefined' so omission resolves to on_success, matching the extension and the kit's own empty-string path. Deploys are unaffected.

Per maintainer ruling, isAnimated keeps its default of true: the extension's unset-env false came from a shipped parser bug ('=== "true" || undefined' made the intended unset check unreachable), and the kit fixes the bug rather than reproducing it. Both behaviors are now documented in the README.

Tests pin the new resolution (undefined and empty string) and the handler consequence (success deletes, failure does not); reverting the change fails them. Verified with mocked storage, not a live bucket. 185 tests pass, tsc clean.

Fixes #3024


kits/storage-resize-images/src/export-config.ts
```diff
@@ -127,7 +127,6 @@ function deleteOriginalFile(
       return DELETE_IMAGE.always;
     case false:
     case "false":
-    case undefined:
       return DELETE_IMAGE.never;
```
IzaakGough, line 130: With `case undefined` gone, a caller who omits `deleteOriginal` resolves to `onSuccess`, and `resolveResizeImagesConfig` is exported from `src/lib.ts`, so a library consumer writing `resolveResizeImagesConfig({ bucket, sizes })` now has every original deleted after a successful resize where the previous kit release kept it. The sharpest case is the content filter: with no `failedImagesPath` set, `handleFailedImage` stores nothing, the placeholder resize succeeds, and the flagged original is deleted with no copy left anywhere, which is what the new test at tests/handlers.test.ts:418 pins. One thing worth weighing in the parity argument: `extension.yaml` declares `DELETE_ORIGINAL_FILE` as `required: true` with `default: false`, so the unset arm did not run in a real install. This needs a decision before merge, either defaulting the optional field to `"false"` so omission stays safe, or making `deleteOriginal` required so an absent field cannot delete data.

kits/storage-resize-images/src/export-config.ts
```diff
@@ -127,7 +127,6 @@ function deleteOriginalFile(
...
     default:
       return DELETE_IMAGE.onSuccess;
```
IzaakGough, line 132: Non-blocking. The `default` arm now takes both omission and every unrecognized value into the destructive branch. `DELETE_ORIGINAL_FILE` is a plain `defineString` (src/config.ts:90) with no validation, so a hand-written `.env` carrying `False`, `no`, or `0` resolves to delete-on-success. Throwing on anything that is not `true`, `false`, or `on_success` would keep whatever you decide for omission while making a typo fail loudly rather than delete originals.

kits/storage-resize-images/README.md
```diff
@@ -102,6 +102,12 @@ loads them at deploy time and prompts for any required values that are missing.
 | `customFilterPrompt`   | `CUSTOM_FILTER_PROMPT`   | no       | (empty)                | Custom filter prompt                                                                       |
 | `placeholderImagePath` | `PLACEHOLDER_IMAGE_PATH` | no       | (empty)                | Placeholder for filtered images                                                            |
 
+The `deleteOriginal` default above is what the CLI writes into `.env` at deploy
```
IzaakGough: Non-blocking, and cheap to fix in this PR. Reading firebase-tools 15.29.0 (`lib/deploy/functions/params.js`, `resolveParams`), a param absent from `.env` is prompted for with its declared default and the resolved value is injected into the function environment, and a non-interactive deploy fails outright instead. Nothing is written back into `.env`, so the deploy-safety conclusion holds but by a different route than this sentence describes. Also, the table above still lists the `deleteOriginal` default as `false`, so a reader who only scans the table takes away the opposite of the new behaviour; a footnote on that row pointing down here would close the gap. This comes from reading the dependency source, not from a live deploy.

IzaakGough (commented):
Read through the diff and ran the kit suite at head (186 pass, 3 skipped), then re-ran it with the one-line change in `src/export-config.ts` reverted (3 failures). That is how the test at tests/handlers.test.ts:438 turned up as non-discriminating: two of the three new handler tests fail on the revert, that one passes either way.

One blocking comment on `src/export-config.ts`, about an omitted field becoming an irreversible delete on the exported resolver. The other four are non-blocking, and the two README ones are cheap to fix in this PR.

kits/storage-resize-images/README.md
```diff
@@ -180,6 +186,16 @@ the extension is installed. A malformed value fails the deploy with
...
 values.` rather than being rejected by an install prompt.
 
+### An omitted `isAnimated` keeps animation
+
+The extension's config parser had a bug: `overrideIsAnimated === "true" ||
+undefined` never evaluated the intended unset check, so an unset `IS_ANIMATED`
+produced first-frame-only output even though the parameter's declared default
+was `true`. The kit deliberately fixes this rather than reproducing it: an
+omitted `isAnimated` resolves to `true`, the default the extension intended.
+Deploys are unaffected either way, since the CLI writes `IS_ANIMATED=true`
```
IzaakGough: Non-blocking, and also cheap here. `isAnimated` is a `defineBoolean`, and in firebase-functions `BooleanParam.runtimeValue()` reads `!!process.env[name] && process.env[name] === "true"`, so when `IS_ANIMATED` is not present in the runtime environment `configFromEnv()` yields `false` and animation is dropped. The fix applies to callers who omit the field programmatically, so scoping the section to that path rather than to omission in general would match what the code does. Read from the dependency source, not exercised against a deploy.

kits/storage-resize-images/tests/handlers.test.ts
```diff
@@ -390,6 +397,62 @@ describe("generateResizedImageHandler", () => {
...
+    await generateResizedImageHandler(mockObject, ctx, false);
+
+    expect(deleteRemoteFile).toHaveBeenCalledWith(
+      remoteFile,
+      "images/test.jpg"
+    );
+    expect(deleteRemoteFile).toHaveBeenCalledTimes(1);
+  });
+
+  test("an omitted deleteOriginal keeps the original on a failed run", async () => {
```
IzaakGough: Non-blocking: this one passes with the source change reverted, because an omitted `deleteOriginal` then resolves to `never` and `deleteRemoteFile` is not called either way. It also lands on the same handler branch as the test at line 357, which reaches `on_success` through an explicit override. Dropping it, or asserting something only the resolved `on_success` path can produce, would keep the suite discriminating. Separately, the comment at line 420 mentions `failedImagesPath` storing a copy first, and that path is not exercised there since `handleFailedImage` is mocked and no `failedImagesPath` is configured.

kits/storage-resize-images/src/export-config.ts
cabljac: Fair points, and the `required: true` observation is right: the unset arm never ran in a real install. The ruling on #3024 is still parity, so an omitted `deleteOriginal` resolves to `onSuccess` as the extension's parser did. The test at handlers.test.ts pins that deliberately, including the filter-blocked case, so the behaviour is visible rather than accidental. The library path is the one place this bites, and the README now says so. Making the field required on the type is a fair follow-up if we want omission to be impossible rather than documented; happy to file it.

kits/storage-resize-images/src/export-config.ts
cabljac: Pre-existing rather than new here: `False`, `no` and `0` already resolved to `onSuccess` before this PR, and that matches the extension's parser. Throwing on unrecognised values would be stricter than the extension, so it needs its own decision. I'll leave it for a follow-up if you want to file one.

kits/storage-resize-images/README.md
cabljac: Half right, thanks. The prompt and non-interactive failure are as you describe, but the value is written back: `writeResolvedParams` in `src/functions/env.ts` persists prompted params to `.env.<projectId>`, not `.env`. Fixed the note and the table footnote in 5cda761a.

kits/storage-resize-images/README.md
cabljac: Correct, `BooleanParam` reads an absent var as `false`, so the deployed function still drops animation when `IS_ANIMATED` is unset. Retitled the section to library callers and said the env path is unchanged, in 5cda761a.

kits/storage-resize-images/tests/handlers.test.ts
cabljac: Confirmed, it passed with the fix reverted. Dropped it and corrected the comment at line 420 in 9c643aba.

IzaakGough (approved):

PR #3102, opened 2026-09-02
fix(firestore-send-email): map DATABASE_REGION to a Cloud Run region for function placement

### LINKED ISSUE #3069: fix(firestore-send-email): DATABASE_REGION used as function region breaks multi-region deploys
Same conflation just fixed for firestore-bigquery-export in #3066: `kits/firestore-send-email/src/config.ts:375` sets `region: params.databaseRegion`, and the select offers the Firestore multi-regions `eur3`/`nam5`/`nam7`, which are not valid Cloud Run regions - deploys hard-fail for multi-region databases. Apply the #3066 pattern: drop the `region:` option (firebase-tools >= 15.27 resolves Firestore-triggered function regions from the database location via its dual-region mapping, and pins the Eventarc trigger region independently) and remove the param if nothing else consumes it. Found during the #3017 investigation. Ledger: #2974.

cabljac:
Fixes #3069. The kit passed the raw DATABASE_REGION value as the function's region, and the param's select offers the Firestore multi-regions eur3/nam5/nam7, which are not Cloud Run regions, so multi-region deploys hard-failed.

The issue proposed dropping the region option (the #3066 pattern). This PR deliberately deviates: by agreement with the firebase-tools team, DATABASE_REGION stays and is mapped instead, because the CLI's own inference is unreliable for param-declared databases (firebase/firebase-tools#11020) and the extension-to-kit migration exports DATABASE_REGION into the user's .env. Multi-regions map to a region inside them (nam5/nam7 to us-central1, eur3 to europe-west1, mirroring the CLI's FIRESTORE_DUAL_REGION_TO_REGION_MAPPING); regional locations pass through; unset or empty means no region option and the CLI fallback. #3101 is the sibling fix for firestore-bigquery-export.

Tests pin all mapping cases plus the unset case; breaking the mapping fails them. Caveats: no live multi-region deploy was run for this kit, and the module-load process.env read needs firebase-tools >= 15.28.0 during discovery (older CLIs degrade to the no-region fallback). The approved #3089 also touches src/config.ts; this change edits only envDeployOptions at the end of the file, so overlap is limited to a trivial CHANGELOG conflict.


kits/firestore-send-email/README.md
```diff
@@ -158,15 +158,25 @@ value, and on OAuth2 auth do the same for `SMTP_PASSWORD`.
 
 ### DATABASE_REGION now decides where the function runs
```
IzaakGough, line 162: Worth documenting the firebase-tools floor in this section. Placement reads `process.env.DATABASE_REGION` at discovery, so on 15.25.x to 15.27.x the value is not loaded, no region is declared, and the CLI falls back to the existing region or `us-central1`. On those versions the CHANGELOG line about `nam5` and `nam7` deploying to `us-central1` does not hold, and the deploy section above tells users 15.25.1 is enough for kits, so someone on that version gets the old behaviour with nothing to tell them.

The other case for the same paragraph: on a fresh interactive install the prompted value only takes effect from the second deploy, since `resolveDefaultRegionsForBuild` runs before param resolution. The first deploy lands in `us-central1` and the next one moves the function. The bigquery-export README covers both points, so mirroring that wording here would do it.

kits/firestore-send-email/src/config.ts
```diff
@@ -369,9 +370,14 @@ export function configFromEnv(): SendEmailConfig {
 }
 
 export function envDeployOptions(): DeployTimeOptions {
+  // The region option cannot be a param expression, so the value is read from
```
IzaakGough: One note on the reasoning here: the region option does accept a param expression. firebase-functions types it as `SupportedRegion | string | Expression<string> | ResetValue`, firebase-tools resolves CEL in `region` via `params.resolveList`/`resolveString` in `deploy/functions/build.js`, and the code this replaces passed `params.databaseRegion` straight in. The constraint that actually forces the env read is that the multi-region lookup cannot be expressed in CEL, so the value has to be concrete at discovery time.

Rewording it that way also makes the trade-off visible, since the env read is what makes placement a silent no-op on CLIs older than 15.28.0.

kits/firestore-send-email/README.md
```diff
@@ -158,15 +158,25 @@ value, and on OAuth2 auth do the same for `SMTP_PASSWORD`.
...
+Firestore trigger always fires in the database's own region, whatever region
+the function runs in.
+
+With `DATABASE_REGION` unset or empty, the function declares no region and the
+Firebase CLI resolves one at deploy time: it keeps the region it is already
+deployed in, and on a first deploy lands in `us-central1` unless you set the
+`FIREBASE_FUNCTIONS_DEFAULT_REGION` environment variable when running
+`firebase deploy`. Careful with that variable: it applies to every no-region
+function in the deploy, not just this kit. Note that changing an existing
+install's function region deletes and recreates the function in the new
```
IzaakGough, line 194: Could be worth spelling out what the recreate costs for this kit specifically. `processQueue` is the only function and there is no reconciliation pass, so mail documents written while the function is gone produce no Firestore event and are never delivered. They stay in the collection with no `delivery` state and nothing retries them.

Both paths above reach it: the first interactive deploy that moves the function on the second run, and upgrading the CLI past 15.28.0 with `DATABASE_REGION` already in `.env`. A sentence suggesting people drain the mail collection before a region change would help.

kits/firestore-send-email/src/region.ts
```diff
@@ -0,0 +1,37 @@
...
+ * should deploy to. Regional locations pass through unchanged; an unset or
+ * empty location returns `undefined`, meaning the functions declare no region.
+ */
+export function firestoreLocationToFunctionRegion(
+  location: string | undefined
+): string | undefined {
+  if (!location) {
+    return undefined;
+  }
+  return MULTI_REGION_TO_FUNCTION_REGION[location] ?? location;
```
IzaakGough: This lookup is case-sensitive, where the CLI equivalent it mirrors lowercases first (`getDefaultRegion` does `db.locationId.toLowerCase()` before indexing the same table). The select input keeps the prompted flow safe, but `.env` gets hand-edited and hand-copied during migration, and `DATABASE_REGION=NAM5` passes `NAM5` through as the function region, which is the same hard deploy failure this change removes.

`location.trim().toLowerCase()` before the lookup covers it. The trim is only belt and braces, values are already trimmed by the CLI's dotenv parser.

kits/firestore-send-email/tests/deploy-options.test.ts
```diff
@@ -15,32 +15,75 @@
...
+    setDatabaseRegion(undefined);
+    expect(envDeployOptions()).not.toHaveProperty("region");
+  });
+
+  test("empty DATABASE_REGION omits the region option", () => {
+    setDatabaseRegion("");
+    expect(envDeployOptions()).not.toHaveProperty("region");
   });
 
   test("serialized deploy-time options do not contain undefined", () => {
```
IzaakGough, line 79: Two coverage gaps worth closing while this is open.

Nothing imports `src/index.ts`, so the new spread on `processQueue` is unexercised: deleting `...(deploy.region ? { region: deploy.region } : {})` leaves the whole suite green, and that line is what actually places the function.

This test also now forces `DATABASE_REGION` unset, so the case it was written for, a region present in the serialized options, is no longer covered. The bigquery-export tests assert per-function options through a `triggerOptions()` helper, and the same shape would work here.

IzaakGough (approved):

Issue #3031, opened 2026-09-01
feat(firestore-bigquery-export): design write-path buffering


cabljac:
The extension buffers failed BigQuery writes in a Cloud Tasks queue (`syncBigQuery`, maxAttempts 5, dispatch throttling); on terminal failure it swallows the error, writing the row to `BACKUP_COLLECTION` only when that optional param is set. The kit has no queue: one self-heal re-attempt, then rethrow with `retry: true`, so it never loses a row but a persistent BigQuery error replays every write through Eventarc backoff for up to 24h. Aggregate cost scales with write rate x outage duration (each retried invocation performs up to 4 insert attempts plus 2 backup commits), and the flood can saturate instances. Parity ledger: #2974, firestore-bigquery-export §1.

## DECISION (2026-09-02): reinstate the Cloud Tasks buffer (option D)

Team decision: port the extension's Cloud Tasks buffer architecture into the kit rather than shipping the retry-shape design below. Rationale: migration safety on our delivery timeframe - migrating users keep exactly the failure behavior and cost profile they have today (about 5 attempts, dispatch-throttled), no billing-spike surprise, no novel classifier/gate code to debug in release candidates. A native platform solution is expected around late 2027, so the sophisticated design solves a problem the platform will eventually own; if needed we follow up in a later version.

Accepted trade-offs, recorded deliberately: the current kit's never-loses-within-24h property goes away (rows drop after the queue's attempts when no backup collection is set - which is extension parity, i.e. the bar), and the known kit queue-leak on function deletion now applies to this kit (document it).

Port-time cautions - fix these legacy bugs while porting, they are bug fixes not divergence:
1. Do NOT port `attemptToEnqueue`'s silent swallow (legacy `index.ts:290-322` dropped the event outright when the enqueue itself failed, with no visibility). Log loudly at error level and rethrow so the trigger retry covers the enqueue-failure window.
2. Do NOT reintroduce tracker 2.0.4 semantics: the always-truthy retry guard and the `!retry`-only backup are fixed in 2.1.0, which the kit consumes. The port must keep 2.1.0's back-up-on-every-terminal-failure behavior.
3. Keep the strongly-recommended `BACKUP_COLLECTION` framing (prompt description is the delivery vehicle: the CLI prompts every param absent from dotenv, defaulted or not, and hard-errors non-interactively - verified against firebase-tools 15.26.0 `lib/deploy/functions/params.js` `resolveParams`).

[...]

PR #3103, opened 2026-09-02
feat(firestore-bigquery-export): reinstate the Cloud Tasks write buffer

cabljac:
Ports the extension's Cloud Tasks write buffer into the kit, per the option D decision on #3031: migrating users keep the failure behaviour and cost profile they have today.

A failed inline write enqueues onto a new `syncBigQuery` task queue (5 attempts, 60s minimum backoff, throttled by `MAX_DISPATCHES_PER_SECOND`, default 100) after up to `MAX_ENQUEUE_ATTEMPTS` in-process attempts (default 3). The task handler runs `ensureInitialized` as a self-heal, attempts the write even if that fails, and rethrows so Cloud Tasks retries; tracker 2.1.0 writes the rows to `BACKUP_COLLECTION` on every terminal insert failure. The trigger declares no retry policy and a failed enqueue is logged, published as `onError`, and dropped, both exactly as the extension does. Enqueue targets the bare function name (firebase-admin 14.2.0+ resolves the `FIREBASE_KIT_INSTANCE_ID` prefix, Firebase CLI 15.28.0+), in the CLI-set `FUNCTION_REGION`, with a task id derived from the event id so a retried enqueue cannot buffer the same event twice. One deliberate change: the success event is published after the insert and swallowed on failure, closing a duplicate-row route the extension has.

91 unit tests pass. No live queue E2E has been run: that the deployed env carries `FIREBASE_KIT_INSTANCE_ID` and `FUNCTION_REGION`, that the gen2 task function accepts the SDK's ID token, queue creation with the resolved throttle, and the task size limit on large documents are all unverified. The tracker pins firebase-admin ^13, so a second admin copy is nested under it; a follow-up should widen that range.

Fixes #3031


cabljac:
Live E2E checklist (dev-extensions-testing), to run before marking ready for review:

- [ ] Deploy with firebase-tools >= 15.28.0; confirm `kit-<id>-syncBigQuery` queue is provisioned and `FIREBASE_KIT_INSTANCE_ID` is set on the functions
- [ ] Induce an insert failure (retype a changelog column, or revoke `roles/bigquery.dataEditor`); confirm the trigger execution succeeds and a task lands on the queue (bare-name enqueue resolves the kit prefix on admin 14)
- [ ] Observe 5 task attempts at >= 60s spacing, dispatch rate capped at `MAX_DISPATCHES_PER_SECOND`
- [ ] With `BACKUP_COLLECTION` set: confirm the backup doc exists (keyed by event id) at exhaustion; restore BigQuery, reconcile via the event_id anti-join MERGE, verify the row lands
- [ ] With `BACKUP_COLLECTION` unset: confirm the row is dropped after attempt 5 (extension parity) and the error log carries `retry_count`
- [ ] Revoke `roles/cloudtasks.enqueuer`: confirm the enqueue failure is logged at error level and the Firestore event is redelivered by Eventarc (not lost)
- [ ] Confirm `onSuccess` is published from the task handler when a buffered write lands
- [ ] Deploy once with an older CLI (< 15.28.0) to confirm the documented loud-failure degradation

kits/firestore-bigquery-export/src/logs.ts
```diff
@@ -208,7 +208,8 @@ export const logFailedEventAction = (
   document_name: string,
   event_id: string,
   operation: ChangeType,
-  error: Error
+  error: Error,
+  retry_count?: number
```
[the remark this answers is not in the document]
cabljac: The signature and the call sites do align. The full signature at logs.ts:206-213 is (action: string, document_name: string, event_id: string, operation: ChangeType, error: Error, retry_count?: number) - action is the first parameter; the diff hunk starts mid-signature, which makes it look shifted. Both call sites in handlers.ts pass the action string first, and the new retry_count argument is the optional sixth parameter passed only by handleSyncBigQueryTask. tsc --noEmit is clean, which it could not be with a real parameter mismatch. No change made.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  return `locations/${region}/functions/${functionName}`;
+}
+
+function backoffMs(attempt: number, jitter: number): number {
+  return (
+    Math.min(Math.pow(2, attempt) * BACKOFF_BASE_MS, MAX_BACKOFF_MS) + jitter
+  );
+}
```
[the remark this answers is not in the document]
cabljac: The constant per-invocation jitter is a deliberate line-for-line port of the legacy extension's attemptToEnqueue, which computes jitter once outside the retry loop; keeping the extension's exact failure behavior is this PR's explicit goal (option D on #3031). With at most 10 attempts and a 0-100ms jitter range, per-attempt jitter would make no practical difference to thundering-herd behavior here. Declining for parity; noted as a possible follow-up once the port has shipped.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const jitter = Math.random() * JITTER_MS;
+  let attempts = 0;
+
+  while (attempts < maxAttempts) {
+    if (attempts > 0) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, backoffMs(attempts, jitter))
+      );
+    }
```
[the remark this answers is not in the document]
cabljac: The constant per-invocation jitter is a deliberate line-for-line port of the legacy extension's attemptToEnqueue, which computes jitter once outside the retry loop; keeping the extension's exact failure behavior is this PR's explicit goal (option D on #3031). With at most 10 attempts and a 0-100ms jitter range, per-attempt jitter would make no practical difference to thundering-herd behavior here. Declining for parity; noted as a possible follow-up once the port has shipped.

kits/firestore-bigquery-export/src/handlers.ts
```diff
@@ -204,8 +216,65 @@ export async function handleDocumentWrite(
...
+
+  logs.logEventAction(
+    "Firestore event received by onDispatch trigger",
+    change.fullResourceName,
+    change.eventId,
+    change.changeType
+  );
+
+  try {
+    await ctx.ensureInitialized();
```
cabljac: `ensureInitialized()` runs before the insert, and the backup write only exists inside the tracker's `insertData` catch. A persistent init failure (dataset create denied, KMS key unreachable, materialized-view config rejected) throws here, the insert never runs, no row reaches `BACKUP_COLLECTION`, and all five attempts fail identically. README:246-249 and the doc comment at `src/index.ts:168` both claim the backup covers every terminal failure.

`skipInit: true` (`src/export-config.ts:210`) makes this the only init path, so nothing else covers it.

Fix: catch and log the init error, then attempt the write anyway. A missing table fails the insert terminally, which is the path that does back the row up. Test: init rejects, record still attempted.

Separately, README:232-233 oversells the self-heal. `createEnsureInitialized` memoizes success permanently (`src/init.ts:35-44`); it only retries after a failed init. A warm instance whose table was dropped will not re-provision.

kits/firestore-bigquery-export/src/handlers.ts
```diff
@@ -204,8 +216,65 @@ export async function handleDocumentWrite(
...
+    await recordEventToBigQuery(change, ctx.tracker);
+
+    await events.recordSuccessEvent({
```
cabljac, line 261: `recordSuccessEvent` is inside the same try as the insert, and the catch rethrows at :278. If the publish fails after the row lands, Cloud Tasks retries at >= 60s, outside BigQuery's insertId dedupe window, and writes a duplicate event_id row.

Fix: move the publish after the try/catch and swallow-with-log. Test: record resolves, recordSuccessEvent rejects, handler resolves.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,101 @@
...
+  while (attempts < attemptBudget) {
+    if (attempts > 0) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, backoffMs(attempts, jitter))
+      );
+    }
+
+    attempts++;
+    try {
+      await queue.enqueue(payload);
```
cabljac: No `opts.id`, so nothing dedupes an enqueue. Two paths: the enqueue lands server-side but the response is lost and the loop re-enqueues; or every attempt fails, `enqueueForSync` rethrows (`handlers.ts:126`), Eventarc redelivers, and the retry enqueues alongside anything that did land. Each extra task is an extra insert on its own schedule, past the dedupe window.

The second path is new here - the extension swallowed enqueue failures, so redelivery never happened.

Fix: pass `{ id: <eventId sanitized to [A-Za-z0-9_-]{1,500}> }` and treat `task-already-exists` as success.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @param functionName - The export name of the task function.
+ * @returns The queue resource path, `locations/<region>/functions/<name>`.
+ * @throws If no region can be resolved.
+ */
+export function syncQueuePath(
+  functionName: string = SYNC_BIGQUERY_FUNCTION
+): string {
+  const region =
+    firestoreLocationToFunctionRegion(process.env.DATABASE_REGION) ??
+    process.env.FUNCTION_REGION;
```
cabljac: `FUNCTION_REGION` is a gen1-era variable, not set on Node 22 gen2, and not referenced by firebase-functions 7.3.2 or firebase-tools 15.26.0. Unverified on a live deploy.

If that holds, then with `DATABASE_REGION` empty this throws at :49, and since `syncQueuePath()` is evaluated at :78 outside the retry loop, the throw skips every attempt and propagates through `enqueueForSync`. With `retry: true` that is 24h of redelivery per failed inline write, which is the cost profile option D exists to avoid.

Fix: drop the fallback and fail at config resolution when the region is unresolvable, rather than at first enqueue. Test: `DATABASE_REGION` as an empty string, not just unset.

kits/firestore-bigquery-export/src/config.ts
```diff
@@ -646,5 +679,7 @@ export function configFromEnv(): ExportConfig {
     transformFunction: optional(params.transformFunction.value()),
     kmsKeyName: optional(params.kmsKeyName.value()),
     logLevel: normalizeLogLevel(params.logLevel.value()),
+    maxDispatchesPerSecond: params.maxDispatchesPerSecond.value(),
+    maxEnqueueAttempts: params.maxEnqueueAttempts.value(),
```
cabljac: `IntParam.value()` returns `runtimeValue()`, which is `parseInt(process.env[name] || "0", 10) || 0` (firebase-functions 7.3.2, `lib/params/types.js:19-26` and `:411`). `default: 3` at :315 is never consulted at runtime. Unset or blank resolves to 0, clamped to 1 by `tasks.ts:80`, so one enqueue attempt instead of three.

Deployed installs are covered because the CLI writes params to .env; the emulator and a hand-written .env are not.

Fix: reuse the `optionalInt` helper from `kits/delete-user-data/src/config.ts:173-178`, which already exists for exactly this. `MAX_DISPATCHES_PER_SECOND` is unaffected, only consumed as a deploy expression at :487.

`tests/config.test.ts:48-50` mocks `defineInt` as `value: () => opts?.default ?? 0`, so the defaults test at :174 asserts against the mock and cannot catch this.

kits/firestore-bigquery-export/README.md
```diff
@@ -216,10 +228,51 @@ curl -fsS -X POST -H "Content-Type: application/json" -d '{"data":{}}' \
...
+a self-heal before re-attempting the write.
+
+## Failure handling
+
+The write path mirrors the extension's Cloud Tasks buffer:
+
+1. The trigger attempts the BigQuery insert inline. On success, done.
+2. On failure, it enqueues the serialized change onto the `syncBigQuery` queue
+   (up to `MAX_ENQUEUE_ATTEMPTS` in-process attempts with backoff) and the
+   execution succeeds - the Firestore event is not redelivered.
```
cabljac: Step 2 is wrong. `extension.yaml` declares no `retry` on the eventTrigger, so the extension never redelivered. This kit sets `retry: true` (`src/index.ts:159`) and rethrows on serialization failure (`src/handlers.ts:188`) and start-event failure (`:201`), both before the insert is attempted. A serialization failure is deterministic per document, so one poison document replays for 24h.

Fix: correct step 2 and the PR description, and stop rethrowing the serialization failure - retrying cannot help it.

kits/firestore-bigquery-export/README.md
```diff
@@ -216,10 +228,51 @@ curl -fsS -X POST -H "Content-Type: application/json" -d '{"data":{}}' \
...
+3. `syncBigQuery` re-attempts the write on the queue's schedule: 5 attempts,
+   60 seconds minimum backoff, throttled to `MAX_DISPATCHES_PER_SECOND`
+   dispatches per second (500 concurrent max).
+4. On every terminal insert failure the tracker writes the row to
+   `BACKUP_COLLECTION` (when configured), keyed by the event id, before the
+   task fails. After the fifth attempt the task is dropped. **Without a backup
+   collection, the row is dropped with it** - configure `BACKUP_COLLECTION`.
```
cabljac: Unverified against the version we actually resolve. Local source is 2.0.4 (`firestore-bigquery-export/firestore-bigquery-change-tracker/package.json`); this kit depends on published 2.1.0, which is not in the tree.

In 2.0.4 the backup is gated on `!retry` (`src/bigquery/index.ts:247`), and `retry` only flips false when `isRetryableInsertionError` is truthy at :236. That method is `async` (:162) and is called without `await`, so the condition is a Promise and always truthy. Every first failure recurses, every second failure backs up. This line is accidentally correct.

If 2.1.0 added the missing await, the backup narrows to retryable errors only and this line becomes false. Settle against 2.1.0 before merge.

cabljac (commented):
Self-review before this goes out for review. Seven items: five are code changes, two are doc corrections. The tracker one is unverified and needs settling before merge.

This started as an AI-assisted review; I've verified the findings below against the source myself, except the two I've explicitly marked unverified.

kits/firestore-bigquery-export/README.md
cabljac: Correction: withdrawn. I have now verified 2.1.0 directly. The tarball at the shrinkwrap's resolved URL matches its sha512 integrity, and the same source is on `next` (#2937, "scope insert retry to schema lag").

2.1.0 rewrote `insertData`. `isRetryableInsertionError` no longer exists; the un-awaited guard was a known bug and is the headline entry in the 2.1.0 CHANGELOG. Both replacement predicates (`schemaLagColumns`, `isTransientInsertionError`) are synchronous. The backup is unconditional on the terminal path (`lib/bigquery/index.js:294`) and wrapped so a failed backup does not replace the insert error.

So this line is correct by design. I read the tracker source on `kits`, which is still 2.0.4 because `kits` predates #2937; `next` has 2.1.0 at the same path.

Separate and not this PR: the extension's `functions/package-lock.json` on `next` still resolves the tracker to 2.0.4, so deployed extensions keep the old retry behaviour until that lockfile is bumped.

kits/firestore-bigquery-export/src/tasks.ts
cabljac: Retracted. The CLI sets `FUNCTION_REGION` on every deployed gen2 function: firebase-tools `lib/gcp/cloudfunctionsv2.js:197` assigns `environmentVariables["FUNCTION_REGION"] = endpoint.region`, present in 15.26.0, 15.27.0 and 15.28.0, and `lib/functions/env.js` reserves the name. The claim that it is never set was wrong, and 0f30873c removed the fallback on that basis.

4b1b9ba4 restores it and prefers it: it is the region the function was actually deployed to, which a `DATABASE_REGION`-derived value can disagree with on a first interactive deploy or when the variable is empty (both documented in the README). This also matches `firestore-vector-search` after #3118 (`src/handlers.ts:60`).

kits/firestore-bigquery-export/README.md
cabljac: Resolved differently from the fix proposed here. da45faf7 drops `retry: true` from the trigger and stops rethrowing on enqueue exhaustion, so the failure behaviour matches the extension: no runtime retry policy (`extension.yaml:56-73`), and a failed enqueue is logged, published as `onError`, and dropped (`functions/src/index.ts:317-327`). README steps 2 and 5, the "Failed writes" section, and the CHANGELOG are updated to say so. The redelivery surface described in this comment no longer exists.

cabljac:
Closing in favour of the same change split into a reviewable stack (stack #3132), bottom to top:

- #3127 chore: bump firebase-admin to ^14.2.0
- #3128 feat: add the sync task enqueue module
- #3129 feat: add the sync queue params
- #3130 feat: reinstate the Cloud Tasks write buffer (the behaviour change)
- #3131 docs: document failure handling and recovery

The top of the stack is identical to this branch on `kits/firestore-bigquery-export`, rebased onto current `kits`.

Issue #3106, opened 2026-09-03
fix(firestore-translate-text): stop coercing non-string input before translation


CorieW:
The kit calls `translateSingle(String(input), …)`, so a document with `input: 42` sends `"42"` to the translation API where the extension sent the raw `42`. Same for the value logged by `logs.translateInputStringToAllLanguages`.

Only reachable on the create path — the update path deletes existing translations for any non-string, non-object input before it gets here.

Parity ledger: #2974, firestore-translate-text §4b.

PR #3109, opened 2026-09-03
fix(firestore-translate-text): stop coercing non-string input before translation

CorieW:
The kit's `translateDocument` called `translateSingle(String(input), …)`, so a document whose input field is `42` sent the string `"42"` to the translation API and logged `"42"`. The extension passes the raw field value straight through (`firestore-translate-text/functions/src/translate/translateDocument.ts:68`). This drops the coercion so the value reaches `translateString` and the log line unchanged.

Only the create path can reach this: `handleUpdateDocument` deletes any existing translations and returns for a non-string, non-object input, and `handleCreateDocument`'s `if (input)` truthiness check lets `42` through. Nothing else in the kit depends on the input being a string by the time it gets here — `translateSingle` only forwards it to `service.translateString` and the log helper.

The kit's extra `input !== null` guard in the object branch (§4a) is untouched, so `null` still routes to `translateSingle` rather than throwing in `Object.entries` the way the extension does. That is a separate divergence and not in scope here; the null test now asserts the uncoerced `null` instead of `"null"`.

Two existing tests asserted the coercion and are inverted; a third covers the log line. 98 tests and `tsc --noEmit` pass. No live translation run.

Fixes #3106
Parity ledger: #2974, firestore-translate-text §4b.

### DIFF
--- a/kits/firestore-translate-text/src/translate/translateDocument.ts
+++ b/kits/firestore-translate-text/src/translate/translateDocument.ts
@@ -50,5 +50,7 @@ export const translateDocument = async (
     );
   }
 
-  await translateSingle(String(input), languages, snapshot, service);
+  // Passed through uncoerced: the extension hands the raw field value to the
+  // translation API, so a numeric input arrives as a number, not "42".
+  await translateSingle(input as string, languages, snapshot, service);
 };
--- a/kits/firestore-translate-text/tests/translate-document.test.ts
+++ b/kits/firestore-translate-text/tests/translate-document.test.ts
@@ -78,7 +78,7 @@ describe("translateDocument", () => {
     });
   });
 
-  test("coerces a non-string, non-object input to a string", async () => {
+  test("passes a non-string, non-object input through uncoerced", async () => {
     const snapshot = makeSnapshot({ input: 42 });
 
     await translateDocument(
@@ -87,10 +87,24 @@ describe("translateDocument", () => {
       makeConfig()
     );
 
-    expect(translateString).toHaveBeenCalledWith("42", "en");
+    expect(translateString).toHaveBeenCalledWith(42, "en");
   });
 
-  test("treats a null input as a single string translation", async () => {
+  test("logs a non-string input without coercing it", async () => {
+    const snapshot = makeSnapshot({ input: 42 });
+
+    await translateDocument(
+      snapshot,
+      makeService({ extractLanguages: vi.fn(() => ["en"]) }),
+      makeConfig()
+    );
+
+    expect(logger.log).toHaveBeenCalledWith(
+      messages.translateInputStringToAllLanguages(42 as never, ["en"])
+    );
+  });
+
+  test("treats a null input as a single translation, uncoerced", async () => {
     const snapshot = makeSnapshot({ input: null });
 
     await translateDocument(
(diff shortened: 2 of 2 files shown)


kits/firestore-translate-text/src/translate/translateDocument.ts
```diff
@@ -50,5 +50,7 @@ export const translateDocument = async (
     );
   }
 
-  await translateSingle(String(input), languages, snapshot, service);
+  // Passed through uncoerced: the extension hands the raw field value to the
+  // translation API, so a numeric input arrives as a number, not "42".
+  await translateSingle(input as string, languages, snapshot, service);
```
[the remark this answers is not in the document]
CorieW: The `TypeError` is real, but it is the extension's behaviour and restoring it is the point of this PR.

The extension's `GenkitTranslator.translate` has the identical sanitiser (`firestore-translate-text/functions/src/translate/common.ts:135-138`) and its `translateDocument` passes the raw field value in (`translate/translateDocument.ts:68`, `input: any`). So a document with `input: 42` on the gemini provider throws `text.replace is not a function` in the extension today. Coercing it in the kit is the divergence #2974 §4b logged, and the recorded decision there is parity.

It is not an uncaught crash either. The throw lands in `translate`'s own `catch`, which runs `logs.translateStringError` + `recordErrorEvent` and rethrows; `translateSingle` and then `handleDocumentWrite` catch it in turn, so it surfaces as a logged translation error with an `onError` event and a completion event — the same path any Translate API failure takes, on both sides.

Adding coercion inside the translators, or widening `Translator.translate` to `unknown`, would keep the kit accepting an input the extension rejects. That is a behaviour change beyond parity, so it belongs in its own decision rather than here.

On the tests masking it: the mock is deliberate. These tests pin routing and the absence of coercion at the `translateDocument` boundary, which is what changed. What the individual translators then do with a non-string is extension-inherited behaviour and unchanged by this PR.

IzaakGough, line 55: Worth noting that this changes behaviour on the Genkit path. `translateSingle` forwards the value to `service.translateString`, and `GenkitTranslator.translate` calls `text.replace(/\\/g, ...)` in `src/translate/common.ts`. `useGenkit` is set whenever the provider starts with `gemini` (`export-config.ts`), so a document written as `{ input: 42 }` now throws `TypeError: 42.replace is not a function`, which records an error event and writes no translation. Previously it translated as `"42"`. The `GoogleTranslator` path is fine, since v2 `translate()` only does `arrify` plus a regex test.

The extension behaves the same way, so this is faithful parity rather than a new divergence, and I don't think it blocks the merge. Two options if you want to tidy it: mention it in the PR body, or coerce inside `GenkitTranslator.translate`, which is where the string assumption actually lives.

Separately, the `input as string` cast is what keeps `tsc` quiet here. Widening `translateSingle`'s parameter to `unknown` (it's also exported from `src/lib.ts`) would surface the mismatch instead of asserting it away.

IzaakGough (approved):
Parity change looks right, and the suite plus `tsc --noEmit` pass locally. One non-blocking note inline about the Gemini path.

kits/firestore-translate-text/src/translate/translateDocument.ts
CorieW: Yep, definitely something to fix in the future. I'm going to create an issue to track this.

Issue #3107, opened 2026-09-03
fix(firestore-send-email): restore the TESTING env path


CorieW:
The kit has the localhost:8132 testing transport (`helpers.ts`, `if (config.testing)`) but nothing sets `testing` from the environment, so only a programmatic `SendEmailConfig.testing` can reach it. The extension read `process.env.TESTING === "true"` in `config.ts`, so an env-driven deploy or emulator run could enter testing mode; the kit cannot.

`TESTING` was never declared in the extension's `extension.yaml` either — it is a plain env read, not a param, so the fix belongs in `configFromEnv()`.

Parity ledger: #2974, firestore-send-email params table.

PR #3110, opened 2026-09-03
fix(firestore-send-email): restore the TESTING env path

CorieW:
The kit already carries the extension's testing transport — `transportLayer` returns a `127.0.0.1:8132` nodemailer transport when `config.testing` is set (`src/helpers.ts`) — but nothing ever set `testing` from the environment. It was reachable only by a programmatic caller passing `SendEmailConfig.testing`, so the env-driven path (deploy, emulator, the extension's own `_emulator` style setup) could not enter testing mode at all. The extension read it in `firestore-send-email/functions/src/config.ts:30`.

This adds `testing: process.env.TESTING === "true"` to `configFromEnv()`. Read from `process.env` rather than declared as a param, matching the extension: `TESTING` never appeared in its `extension.yaml`, only in the config read. Putting it in `configFromEnv` rather than `resolveConfig` keeps the resolver pure, so library consumers who build a `SendEmailConfig` themselves are unaffected and an explicit `testing: false` still wins.

The strict `=== "true"` comparison is the extension's, so `TRUE` / `1` / `yes` stay off; there is a test pinning that. `resolveConfig` already collapsed the optional flag to a boolean, so no type changes were needed.

3 tests added, 124 pass, `tsc --noEmit` clean. No live SMTP run.

Fixes #3107
Parity ledger: #2974, firestore-send-email params table.

### DIFF
--- a/kits/firestore-send-email/src/config.ts
+++ b/kits/firestore-send-email/src/config.ts
@@ -344,6 +344,9 @@ export function configFromEnv(): SendEmailConfig {
     defaultReplyTo: params.defaultReplyTo.value(),
     usersCollection: params.usersCollection.value(),
     templatesCollection: params.templatesCollection.value(),
+    // Read straight from the environment, not declared as a param: the
+    // extension never surfaced TESTING in extension.yaml either.
+    testing: process.env.TESTING === "true",
     ttlExpireType:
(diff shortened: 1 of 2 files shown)


IzaakGough (approved):
lgtm

PR #3111, opened 2026-09-03
fix(firestore-bigquery-export): republish the legacy firestore-counter event types

### LINKED ISSUE #3108: fix(firestore-bigquery-export): republish the legacy firestore-counter event types
The extension published every lifecycle event twice — once as `firebase.extensions.firestore-counter.v1.<name>` (a historical naming mistake kept for backwards compatibility) and once as `firebase.extensions.firestore-bigquery-export.v1.<name>`. The kit publishes only the latter, so any Eventarc trigger filtering on the `firestore-counter` types stops firing on migration.

`getEventTypes` in the kit already returns an array and every publisher already maps over it, so restoring the legacy type is one entry plus the README note that currently documents the drop as intentional.

Parity ledger: #2974, firestore-bigquery-export.

CorieW:
The extension published every lifecycle event twice: once as `firebase.extensions.firestore-counter.v1.<name>` and once as `firebase.extensions.firestore-bigquery-export.v1.<name>` (`firestore-bigquery-export/functions/src/events.ts:30-33`). The `firestore-counter` name is a historical mistake, but it is the type that shipped first, so it is the one existing Eventarc triggers are most likely filtering on. The kit dropped it, which silently stops those triggers on migration: the function still succeeds, the trigger just never fires.

`getEventTypes` already returned an array and all four publishers already `Promise.all` over it, so restoring the legacy type is one array entry. The legacy copy carries the identical `data` and `subject`; only `type` differs. `EXT_SELECTED_EVENTS` still gates what actually leaves the channel, exactly as in the extension, and the extension declares both namespaces in its `events:` block so each is ticked independently.

The README documented the drop as an intentional divergence and told users to repoint their triggers. The `## Events` section now describes the dual publish, and the differences entry is deleted: with the legacy type back and `onSuccess` restored by #3130, there is no difference left to document.

The events test asserted "the firestore-bigquery-export event type only" and is inverted; a payload-equality test is added.

## Changes

- `src/events.ts`: `getEventTypes` returns the legacy `firestore-counter` type alongside the current one, matching the extension's order.
- `tests/events.test.ts`: assert both types on all four publishers, plus a test that the legacy copy differs only by `type`.
- `README.md`: document the dual publish under `## Events`; drop the now-obsolete "Events" entry from the differences section.

## Verification

Rebased onto `kits` (`247185fa6`). 101 tests and `tsc -b` pass; `prettier@2.8.8 --list-different` clean. No live Eventarc run.

Fixes #3108
Parity ledger: #2974, firestore-bigquery-export.

### DIFF
--- a/kits/firestore-bigquery-export/README.md
+++ b/kits/firestore-bigquery-export/README.md
@@ -170,19 +170,27 @@ the instances cannot collide.
 
 ## Events
 
-When `EVENTARC_CHANNEL` is configured, the functions publish lifecycle events
-under `firebase.extensions.firestore-bigquery-export.v1.*`: `onStart` and
-`onError` from the write path, and `onSuccess` from the `syncBigQuery` task
-when a buffered write lands (matching the extension, which only emitted
-`onSuccess` from its queue handler).
+When `EVENTARC_CHANNEL` is configured, the functions publish lifecycle events:
+`onStart` and `onError` from the write path, and `onSuccess` from the
+`syncBigQuery` task when a buffered write lands (matching the extension, which
+only emitted `onSuccess` from its queue handler).
+
+Each event is published twice, exactly as the extension published it: once
+under `firebase.extensions.firestore-bigquery-export.v1.*` and once under
+`firebase.extensions.firestore-counter.v1.*`. The `firestore-counter` type is a
+historical naming mistake the extension kept for backwards compatibility, and
+the kit keeps it for the same reason: triggers listening on it survive the
+migration. The two copies carry the same `data` and `subject`, and only differ
+by `type`. Write new triggers against the `firestore-bigquery-export` types.
 
 Publishing is filtered by `EXT_SELECTED_EVENTS`: the value is split on commas
 and only exactly matching event types are published, silently. An empty value
 suppresses every event, and a value carrying only another product's types
 (the extension offered more than one namespace to tick) publishes nothing. A
 config exported from the extension brings its `EXT_SELECTED_EVENTS` along, so
-check it lists the `firebase.extensions.firestore-bigquery-export.v1.*` types
-you expect, `onSuccess` included.
+check it lists the types you expect, `onSuccess` included. It gates the legacy
+`firestore-counter` copies too, so a trigger on a legacy type only fires when
+that legacy type is listed.
 
 ## Provisioning
 
@@ -366,14 +374,6 @@ write through Eventarc redelivery for up to 24 hours and never lost a row
 inside that window. That property is gone by design - a row that exhausts the
 queue without a configured `BACKUP_COLLECTION` is dropped, exactly as in the
 extension. Set `BACKUP_COLLECTION`.
-
-### Events
-
-Events are published under `firebase.extensions.firestore-bigquery-export.v1.*`
-only. The extension also published a duplicate copy of every event under
-`firebase.extensions.firestore-counter.v1.*`, a historical naming mistake kept
-for backwards compatibility. If you have Eventarc triggers listening on those
-`firestore-counter` types, point them at the `firestore-bigquery-export` types.
 
 ### Wildcard columns include the document ID
 
--- a/kits/firestore-bigquery-export/src/events.ts
+++ b/kits/firestore-bigquery-export/src/events.ts
@@ -18,13 +18,17 @@ import * as eventArc from "firebase-admin/eventarc";
 const { getEventarc } = eventArc;
 
 /**
- * Builds the Eventarc event type for this extension.
+ * Generates both the OLD and NEW event types to maintain backward compatibility.
+ *
+ * Old Event Type: firebase.extensions.firestore-counter.v1.{eventName}
+ * New Event Type: firebase.extensions.firestore-bigquery-export.v1.{eventName}
  *
  * @param eventName The name of the event (e.g., "onStart", "onError", etc.)
- * @returns The event type string.
+ * @returns An array containing both the old and new event types
  */
 const getEventTypes = (eventName: string) => [
-  `firebase.extensions.firestore-bigquery-export.v1.${eventName}`,
+  `firebase.extensions.firestore-counter.v1.${eventName}`, // OLD Event Type for backward compatibility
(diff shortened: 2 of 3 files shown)


kits/firestore-bigquery-export/src/events.ts
```diff
@@ -18,13 +18,17 @@ import * as eventArc from "firebase-admin/eventarc";
...
  *
  * @param eventName The name of the event (e.g., "onStart", "onError", etc.)
- * @returns The event type string.
+ * @returns An array containing both the old and new event types
  */
 const getEventTypes = (eventName: string) => [
-  `firebase.extensions.firestore-bigquery-export.v1.${eventName}`,
+  `firebase.extensions.firestore-counter.v1.${eventName}`, // OLD Event Type for backward compatibility
+  `firebase.extensions.firestore-bigquery-export.v1.${eventName}`, // NEW Event Type following the updated convention
 ];
```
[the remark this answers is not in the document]
CorieW: Checked this against the SDK and it does not throw. `Channel.publish` delegates to the internal client, which filters rather than validates:

```js
// firebase-admin/lib/eventarc/eventarc-client-internal.js:81-89
async publish(events) {
    if (!Array.isArray(events)) { events = [events]; }
    return this.publishToEventarcApi(await this.resolvedChannelName, events
        .filter(e => typeof this.channel.allowedEventTypes === 'undefined' ||
        this.channel.allowedEventTypes.includes(e.type))
        .map(toCloudEventProtoFormat));
}
```

The public docstring on `Channel.publish` says the same: "If channel was created with `allowedEventTypes` and event type is not on that list, the event is ignored." An empty filtered array short-circuits in `publishToEventarcApi` before any HTTP call, so nothing rejects and the trigger is unaffected.

That makes the suggested filter a duplicate of what the SDK already does one layer down, so I am leaving it out.

The behaviour it describes is real and worth stating, though: with `EXT_SELECTED_EVENTS` set to only the `firestore-bigquery-export` types, the legacy copy is silently dropped. That is the intended shape — the extension declared all eight types (both families) in `extension.yaml:652-682` and let the user select which ones to receive. This PR restores the publish so `EXT_SELECTED_EVENTS` can name a `firestore-counter` type at all; before it, selecting one had no effect.

kits/firestore-bigquery-export/src/events.ts
```diff
@@ -18,13 +18,17 @@ import * as eventArc from "firebase-admin/eventarc";
...
+ * Old Event Type: firebase.extensions.firestore-counter.v1.{eventName}
+ * New Event Type: firebase.extensions.firestore-bigquery-export.v1.{eventName}
  *
  * @param eventName The name of the event (e.g., "onStart", "onError", etc.)
- * @returns The event type string.
+ * @returns An array containing both the old and new event types
  */
 const getEventTypes = (eventName: string) => [
-  `firebase.extensions.firestore-bigquery-export.v1.${eventName}`,
+  `firebase.extensions.firestore-counter.v1.${eventName}`, // OLD Event Type for backward compatibility
```
IzaakGough, line 30: Non-blocking, and I don't think the code should change here since this is exactly the parity the PR is restoring.

Worth noting though: this type is byte-identical to the type the firestore-counter kit publishes today (`kits/firestore-counter/src/events.ts` builds `firebase.extensions.firestore-counter.v1.<name>`). In a project running both kits against the same `EVENTARC_CHANNEL`, the counter kit's `onStart` trigger will now fire for every document write this kit exports, with a payload of `{documentId, changeType, before, after, context}` rather than the counter's `{data, params}`, so a handler reading `event.data.params` gets `undefined`. Eventarc triggers filter on `type`, so it can't be filtered downstream; the only escape is setting `EXT_SELECTED_EVENTS` on this kit to the new types only.

In the extension this was accepted history, but in the kits world both are first-class packages someone may deploy side by side, so it might be worth a line in the differences entry with that mitigation named.

kits/firestore-bigquery-export/README.md
```diff
@@ -155,10 +155,12 @@ the instances cannot collide.
...
 
 When `EVENTARC_CHANNEL` is configured, the function publishes `onStart` and
-`onError` lifecycle events under
-`firebase.extensions.firestore-bigquery-export.v1.*`. The extension's
-`onSuccess` event is not published; see the events entry under
-"Differences from the Stream Firestore to BigQuery extension" below.
+`onError` lifecycle events. Each one is published twice, under
+`firebase.extensions.firestore-bigquery-export.v1.*` and under the legacy
+`firebase.extensions.firestore-counter.v1.*` type the extension also used, so
+triggers written against either type keep firing. The extension's `onSuccess`
```
IzaakGough: Non-blocking. This holds only when `EXT_SELECTED_EVENTS` is unset, which this README doesn't mention anywhere (the other kit READMEs do: `firestore-counter`, `firestore-send-email`, `firestore-vector-search`, `speech-to-text`). `firebase-tools` allowlisted `EXT_SELECTED_EVENTS` for `.env` and `ext:export` writes it, so a migrating user can land with it set to the `firestore-bigquery-export` types only, their legacy triggers stay dark, and nothing in the doc points at the knob.

Going the other way, unset (the kit default, since no kit manifest declares events) means both copies publish, so two Eventarc RPCs per write on a path where a publish failure rethrows in `handleWrite` and aborts the BigQuery export. A sentence naming `EXT_SELECTED_EVENTS` would cover both directions.

IzaakGough (approved):
Approving. The change is correct: `getEventTypes` returns both types, all four publishers already `Promise.all` over the array, and the payloads are identical apart from `type`. I checked the restored type against `firestore-bigquery-export/extension.yaml` (both families are declared there), and the suite passes locally including the four event tests.

Two non-blocking documentation notes inline, both about the consequences of the restored type rather than the mechanics.

kits/firestore-bigquery-export/src/events.ts
CorieW: Hmm, this is something worth considering.

cabljac: it's a valid point but like you say, for parity this is a follow-up

Issue #3113, opened 2026-09-03
🐛 [@firebase-function-kits/firestore-incremental-capture] Task queue dispatch fails with "Queue does not exist" due to double kit instance prefixing in tasks.ts


shettyvarun268:
### [REQUIRED] Step 2: Describe your configuration

- Extension name: `@firebase-function-kits/firestore-incremental-capture`
- Extension version: `0.0.2-rc.2` (and `0.0.2-rc.1`)
- Configuration values (redact info where appropriate):
  - `LOCATION`: `us-central1`
  - `SYNC_COLLECTION_PATH`: `posts`
  - `SYNC_DATASET`: `firestore_export`
  - `SYNC_TABLE`: `posts_incremental`
  - `BACKUP_INSTANCE_ID`: `backup-inst-1`
  - `DATASET_LOCATION`: `us`
  - `DATAFLOW_REGION`: `us-central1`
  - `BUCKET_NAME`: `<project-id>.firebasestorage.app`
  - `INSTANCE_ID`: `firestore-incremental-capture`

#### Steps to reproduce:

1. Install the kit into a project:
   ```bash
   firebase functions:kits:install @firebase-function-kits/firestore-incremental-capture@next
   ```
2. Configure `.env` with a watched collection (e.g. `SYNC_COLLECTION_PATH=posts`, `SYNC_DATASET=firestore_export`, `SYNC_TABLE=posts_incremental`).
3. Deploy the kit functions:
   ```bash
   firebase deploy --only functions:firestore-incremental-capture
   ```
4. Mutate or create a document in the watched collection (e.g. `posts/test-doc-1`).
5. Observe the function logs in Cloud Logging for `kit-firestore-incremental-capture-syncdata`.

##### Expected result

`syncData` should successfully enqueue a task to `syncChangelogTask` via Cloud Tasks. The queue `projects/<project>/locations/us-central1/queues/kit-firestore-incremental-capture-syncChangelogTask` receives the payload, dispatches it to the function, and writes the changelog row to BigQuery.

##### Actual result

`syncData` fails with `Error: Queue does not exist`:

```text
Error: Queue does not exist. If you just created the queue, wait at least a minute for the queue to initialize.
    at FunctionsApiClient.toFirebaseError (/workspace/node_modules/@firebase-function-kits/firestore-incremental-capture/node_modules/firebase-admin/lib/functions/functions-api-client-internal.js:421:16)
[...]

PR #3117, opened 2026-09-04
fix(firestore-incremental-capture): stop double-prefixing task queue names

IzaakGough:
Fixes #3113.

## What was broken

`queueName()` prepended `kit-<instance id>-` to the queue name, and the Admin SDK prepends the same prefix from `FIREBASE_KIT_INSTANCE_ID` in `resolveResourceId()`. Every enqueue targeted `kit-<id>-kit-<id>-<function>` and failed with "Queue does not exist", so changelog rows never reached BigQuery and restorations never started.

## What changed

`queueName()` returns the bare export name and lets the SDK add the prefix. `INSTANCE_ID` still namespaces the Dataflow jobs, the flex template object and the run-status documents, so it stays required; the doc comments and README that described it as naming the task queues are updated.

## How it was verified

`tests/tasks.enqueue.test.ts` points `CLOUD_TASKS_EMULATOR_HOST` at a local server and drives the real Admin SDK, asserting the queue URL it produces. It fails on the old code with the doubled prefix and passes on the new one. It also pins the case the fix cannot distinguish: with `FIREBASE_KIT_INSTANCE_ID` unset there is no prefix at all, so the kit has no fallback of its own.

The existing unit tests only checked the string `queueName()` returned, which is why this got through. Not exercised here: a live deploy. The reporter confirmed the same change works against one.


CorieW:
This fix requires `firebase-admin` **14.2.0+**, but `package.json` still allows **14.1.x**.

Version 14.1.x does not read `FIREBASE_KIT_INSTANCE_ID`. Once `queueName()` returns the bare function name, nothing adds the `kit-<instance-id>-` prefix, so enqueues target a non-existent unprefixed queue.

Please bump the minimum to `^14.2.0` and update the shrinkwrap.

CorieW (approved):
lgtm

Issue #3116, opened 2026-09-03
🐛 [@firebase-function-kits/firestore-bundle-builder] Missing requiresAPI declarations for Cloud Firestore and Cloud Storage APIs


shettyvarun268:
### [REQUIRED] Step 2: Describe your configuration

- Extension / Kit name: `firestore-bundle-builder` (`@firebase-function-kits/firestore-bundle-builder`)
- Extension / Kit version: `0.0.2-rc.2`
- Firebase Tools version: `15.29.0`
- Configuration values:
  - Clean GCP / Firebase project deployment where Cloud Firestore API was not previously enabled.

### [REQUIRED] Step 3: Describe the problem

`@firebase-function-kits/firestore-bundle-builder` fails with `HTTP 500` / `7 PERMISSION_DENIED: Cloud Firestore API has not been used in project ... before or it is disabled` upon initial request on fresh projects because `requiresAPI("firestore.googleapis.com", ...)` is not declared in `src/index.ts`.

#### Steps to reproduce:

1. Create a brand-new GCP/Firebase project (or test on a clean project where `firestore.googleapis.com` is not pre-enabled).
2. Scaffold and deploy `@firebase-function-kits/firestore-bundle-builder@0.0.2-rc.2` using Firebase CLI `15.29.0` with standard `.env`:
   ```env
   FUNCTION_DEFAULT_REGION=us-central1
   BUNDLESPEC_COLLECTION=bundles
   BUNDLE_STORAGE_BUCKET=<bucket-name>
   STORAGE_PREFIX=bundles
   ```
   ```bash
   firebase deploy --only functions --force
   ```
3. Invoke the deployed function URL:
   ```bash
   curl -i https://<function-url>/test-bundle
   ```

##### Expected result

`firebase deploy` identifies the required Google Cloud APIs via `requiresAPI()` and enables `firestore.googleapis.com` and `storage.googleapis.com` automatically during deployment. The function successfully queries Firestore and returns `HTTP 404: Could not find bundle with ID test-bundle`.

##### Actual result

`firebase deploy` completes without enabling `firestore.googleapis.com` because `requiresAPI` is missing from `src/index.ts`. When invoked, the function throws a runtime error and returns `HTTP 500`:

```text
[...]

PR #3119, opened 2026-09-04
fix(kits): declare the Cloud Firestore API in every kit that uses it

IzaakGough:
Reported in #3116 against `firestore-bundle-builder`, which returns an HTTP 500 `PERMISSION_DENIED` on a project where `firestore.googleapis.com` is disabled.

**What was broken.** No kit declared `firestore.googleapis.com`, so a functions deploy never enables it. `wantBackend.requiredAPIs` is populated only from `requiresAPI()` declarations, and the API is in neither the gen2 always-enabled set nor `STANDARD_APIS`. Most kits have a Firestore trigger, so `ensureTriggerRegions` looks the database up at deploy time and fails early with a usable message. Four kits use Firestore without a Firestore trigger (`firestore-bundle-builder`, `bigquery-firestore-export`, `delete-user-data`, `speech-to-text`), so nothing checked them and the failure landed at runtime.

**What changed.** Adds `firestore.googleapis.com` to `REQUIRED_APIS` in the 11 kits that use Firestore. `rtdb-limit-child-nodes` and `storage-resize-images` do not touch it and are unchanged. `firestore-genai-chatbot` is included: it never constructs a client, but it writes through the trigger event `DocumentReference`.

`firestore-incremental-capture` also moves off the narrow `firebase-functions/v2/options` import, since `requiresAPI` is only exported from the v2 barrel. The comment there said the barrel fails to load without `@firebase/app`; that does not reproduce on firebase-functions 7.3.2, verified by loading the compiled `lib/index.js` with `@firebase/app` unresolvable.

**Verification.** All 11 kits build and their suites pass. Added an assertion in `firestore-translate-text`, the one kit with an existing `requiresAPI` test harness, and confirmed it fails when the declaration is removed. The firebase-tools behaviour above comes from reading the installed 15.29.0 source, not from a live deploy.

**For the reviewer.** Enabling the API does not provision a Firestore database, so this does not by itself make a deploy succeed on a project that has none. It makes the dependency visible at install and turns a silent runtime 500 into an explicit declaration. Whether the CLI should also check for a database on non-triggered functions is a firebase-tools question, not addressed here.

### DIFF
--- a/kits/bigquery-firestore-export/src/index.ts
+++ b/kits/bigquery-firestore-export/src/index.ts
@@ -56,6 +56,10 @@ const REQUIRED_ROLES: ReadonlyArray<Role> = [
 ];
 const REQUIRED_APIS = [
   {
+    api: "firestore.googleapis.com",
+    reason: "Writes BigQuery export run state to Cloud Firestore.",
+  },
+  {
     api: "bigquery.googleapis.com",
     reason: "Runs scheduled queries and reads their destination tables.",
   },
--- a/kits/delete-user-data/src/index.ts
+++ b/kits/delete-user-data/src/index.ts
@@ -18,7 +18,7 @@ import { PubSub } from "@google-cloud/pubsub";
 import * as admin from "firebase-admin";
 import { getFirestore } from "firebase-admin/firestore";
 import type { Role } from "firebase-functions/v2";
-import { requiresRole } from "firebase-functions/v2";
+import { requiresAPI, requiresRole } from "firebase-functions/v2";
 import { onUserDeleted } from "firebase-functions/v2/identity";
 import { onMessagePublished } from "firebase-functions/v2/pubsub";
 import { CONFIG_EXPRESSIONS, configFromEnv } from "./config";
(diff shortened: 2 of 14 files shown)


CorieW (approved):
The Firestore API declarations look correct and address the reported no-trigger runtime failure. The changed-kit CI jobs are green.

Non-blocking follow-up: while touching `kits/firestore-incremental-capture/src/index.ts`, consider also declaring `bigquery.googleapis.com` and `dataflow.googleapis.com`. The [predecessor extension manifest](https://github.com/GoogleCloudPlatform/firebase-extensions/blob/main/firestore-incremental-capture/extension.yaml#L34-L43) declares both, and the [linked issue discussion](https://github.com/firebase/extensions/issues/3116#issuecomment-5541540870) identified them in the same audit. The required setup script currently enables them out of band, so I do not consider this relevant enough to block this focused Firestore change.

It would also be useful, but non-blocking, to add a direct regression assertion for `firestore-bundle-builder`, since that is the no-Firestore-trigger path reported in #3116.

IzaakGough:
Regression test for firestore-bundle-builder added in 0bf2766d8b26eb37cc476de3a7a7eb775babe0cf.

IzaakGough:
bigquery.googleapis.com and dataflow.googleapis.com are now declared for firestore-incremental-capture in 29208f6b547069ebf527fcfb3f365c82609833b8

PR #3121, opened 2026-09-07
fix(bigquery): preserve project on view updates

### LINKED ISSUE #3120: 🐛 [@firebase-function-kits/firestore-bigquery-export] afterFirstDeploy taskqueue initialization failed from missing PROJECT_ID

- Extension name: @firebase-function-kits/firestore-bigquery-export
- Extension version: [0.0.2-rc.6](https://www.npmjs.com/package/@firebase-function-kits/firestore-bigquery-export/v/0.0.2-rc.6)
  - Firestore in nam5
  - Functions in us-central1
  - Most other configuration the default... maybe using new syntax

#### Steps to reproduce:

1. Install extension
2. Install kit replacement
3. Export extension config to kit

(Note the above steps will only work with firebase-tools at head, but maybe are not necessary. You should be able to do this with a net new install of the kit configured as it deploys)

Kit deploys but afterFirstDeploy hook has issues.

##### Expected result

Kit should fully work after deployment

##### Actual result

Issues. During post-deployment lifecycle execution (afterFirstDeploy), the task queue initialization function kit-firestore-bigquery-export-jdf5-initBigQuerySync failed with an unhandled exception when updating an existing BigQuery snapshot view.

The generated SQL query attempted to query undefined:firestore_export.posts_raw_changelog, which BigQuery rejected due to the invalid project name undefined.

Attached is an analysis from AI

[firestore_bigquery_export_lifecycle_issue.md](https://github.com/user-attachments/files/31890505/firestore_bigquery_export_lifecycle_issue.md)

CorieW:
## Changes

- Pass the BigQuery client's project ID when regenerating an existing snapshot view.
- Add regression coverage proving the configured BigQuery project wins over the function runtime project.

## Testing

- `npm run compile`
- `jest --runInBand src/__tests__/bigquery/materializedViews/initializeLatestView.test.ts`

Closes #3120

### DIFF
--- a/firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/materializedViews/initializeLatestView.test.ts
+++ b/firestore-bigquery-export/firestore-bigquery-change-tracker/src/__tests__/bigquery/materializedViews/initializeLatestView.test.ts
@@ -80,5 +80,44 @@ describe("initializeLatestView", () => {
 
       expect(initializeLatestMaterializedView).not.toHaveBeenCalled();
     });
+
+    it("uses the BigQuery project when updating an existing view", async () => {
+      const originalProjectId = process.env.PROJECT_ID;
+      process.env.PROJECT_ID = "function-project";
+      const metadata = {
+        schema: {
+          fields: [{ name: "document_id" }, { name: "old_data" }],
+        },
+        view: { query: "SELECT FIRST_VALUE(data)" },
+      };
+      mockView.getMetadata.mockResolvedValueOnce([metadata]);
+
+      try {
+        await initializeLatestView({
+          bq: { projectId: "bigquery-project" } as any,
+          dataset: { id: "test_dataset" } as any,
+          view: mockView as any,
+          viewExists: true,
+          rawChangeLogTableName: "test_raw_table",
+          rawLatestViewName: "test_raw_view",
+          changeTrackerConfig: {
+            ...mockConfig,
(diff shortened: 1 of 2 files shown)


IzaakGough (approved):
Fix looks right. The update path was the last place still falling back to `process.env.PROJECT_ID`, and since `bq.projectId` is set to `config.bqProjectId || process.env.PROJECT_ID` it can't be worse than before. I checked out the head and confirmed the new test fails without the added line.

One thing to flag: `firestore-bigquery-change-tracker` is published separately and is still on 2.0.4, which is what `functions/package.json` resolves via `^2.0.4`, so deployed instances won't pick this up until a new version goes out. Happy for that to be a follow-up if releases are batched here, just wanted to make sure it isn't missed.

ajperel:
Is there a larger problem for kits here that all process.env.PROJECT_ID uses won't work? There were a bunch of them: https://github.com/search?q=repo%3Afirebase%2Fextensions+process.env.PROJECT_ID&type=code

I haven't tested a deployed function but from documentation I assume GCLOUD_PROJECT and GOOGLE_CLOUD_PROJECT will be populated alternatives.

cabljac:
Superseded by #3137, which cherry-picked this commit (da219826, author preserved) onto `next` and shipped it as tracker 2.2.1. This PR targeted `kits`, but the tracker publishes from `next`, so the fix had to land there to reach npm. 2.2.1 is now on npm and the kit stack (#3127) pins it. Thanks Corie.

PR #3122, opened 2026-09-07
refactor(kits): read the kit instance id from FIREBASE_KIT_INSTANCE_ID

### LINKED ISSUE #3063: refactor(kits): migrate all kits from the INSTANCE_ID param to the injected FIREBASE_KIT_INSTANCE_ID
firebase-tools >= 15.27.0 injects `FIREBASE_KIT_INSTANCE_ID` for kit instances at discovery, deploy, emulator, and serve time (`lib/functions/env.js` `loadFirebaseEnvs`; provided to discovery envs in `lib/deploy/functions/prepare.js`). The params machinery never consults it and the `FIREBASE_` prefix is reserved in `.env`, so the correct pattern is a direct `process.env.FIREBASE_KIT_INSTANCE_ID` read at discovery time - NOT a `defineString` param. PR #3060 implements this for `kits/delete-user-data`, including the clear missing-var error naming the CLI floor; use it as the template.

Remaining kits still declare `defineString("INSTANCE_ID")` as a user-supplied param (grep `INSTANCE_ID` under kits/*/src). Migrate each: read the injected var, derive instance-scoped defaults (topic names etc.) from it, drop the param, update tests, note the 15.27.0 minimum CLI version in each README. Also decide where the version floor is documented repo-wide (kits README?).

Resolves the #2974 "Look into" question about `FIREBASE_KIT_INSTANCE_ID`.

cabljac:
firebase-tools >= 15.27.0 injects `FIREBASE_KIT_INSTANCE_ID` for every kit instance, but bigquery-firestore-export, firestore-incremental-capture and firestore-vector-search still declared `INSTANCE_ID` as a user-supplied param. The params machinery never sees injected values and `FIREBASE_` is reserved in `.env`, so this follows the #3060 pattern: a plain `process.env` read, the param removed, and `configFromEnv` throwing an error that names the CLI floor when the variable is missing.

Where a kit derives a discovery-time value from the id (the bigquery Pub/Sub topic default, the vector-search query trigger path) the read happens when the deploy entry loads, so an unsupported CLI fails discovery instead of freezing `kit-undefined-*` names into the manifest. The vector-search `./lib` entry still loads without the variable, as its README promises; a test pins that.

The repo-wide CLI floor now lives in `kits/README.md`; each kit README repeats it, drops `INSTANCE_ID` from its params table, and gains a changelog entry.

Tests: bigquery-firestore-export 79, firestore-incremental-capture 75, firestore-vector-search 120, plus `tsc -b` and prettier. Not verified: a live deploy or emulator run against 15.27.0.

Fixes #3063


IzaakGough (approved):
lgtm

PR #3127, opened 2026-09-07
chore(firestore-bigquery-export): bump firebase-admin to ^14.2.0

cabljac:
First layer of the Cloud Tasks write buffer stack. Bumps firebase-admin from ^13.6.0 to ^14.2.0 and the change tracker from ^2.1.0 to ^2.2.1, regenerates the shrinkwrap, and adds one override; no source changes. The admin bump also moves `@google-cloud/firestore` from 7.11.6 to 8.7.1, a major on the client the tracker's backup writes go through.

The next layers enqueue onto a task queue by bare function name. firebase-admin 14.2.0 resolves the deployed `kit-<instance id>-` prefix itself from `FIREBASE_KIT_INSTANCE_ID`, which the Firebase CLI sets on every deployed kit function from 15.28.0, so the kit does not have to reconstruct the prefix.

Tracker 2.2.0 (#3133, #3134) is the first release that runs on firebase-admin 14: earlier versions pinned admin 13 and used the namespaced API, so npm nested a second admin copy under the tracker with no default app and every backup write failed. 2.2.1 (#3137) adds the view-update project fix from #3120. The shrinkwrap now holds a single `firebase-admin@14.3.0`, verified by resolving `firebase-admin/firestore` from the tracker's directory and calling `getFirestore` after the kit's `initializeApp`. The kit pins `firebase-functions ^7.3.3-rc.0`, which no stable range in the tracker can match, so a root `overrides` entry pins the tracker's firebase-functions to the kit's own; the shrinkwrap holds a single `firebase-functions@7.3.3-rc.0` as well, so tracker log lines keep the kit's trace context. That single-copy guarantee reaches consumers through the published shrinkwrap and holds for npm installs only: npm ignores `overrides` from a non-root package, and pnpm and yarn ignore a dependency's `npm-shrinkwrap.json`, so they resolve the tracker's `^6.3.2 || ^7.3.0` to a second firebase-functions copy.

62 unit tests and `tsc -b` pass on this layer; 89 on the top of the stack. The top of the stack was also deployed to a test project with Firebase CLI 15.29.0: `afterFirstDeploy` provisioned the dataset, changelog table, and view; create, update, and delete events landed in the changelog; a redeploy exercised the update-view path with the project preserved; and with the changelog table removed, the trigger enqueued onto `syncBigQuery`, the tracker wrote backup documents through admin 14 / firestore 8.7.1, and the queue drained without duplicate rows once the table was re-provisioned.

Part of #3031.

### DIFF
--- a/kits/firestore-bigquery-export/package.json
+++ b/kits/firestore-bigquery-export/package.json
@@ -31,12 +31,17 @@
     "serve": "firebase emulators:start --only functions"
   },
   "dependencies": {
-    "@firebaseextensions/firestore-bigquery-change-tracker": "^2.1.0",
+    "@firebaseextensions/firestore-bigquery-change-tracker": "^2.2.1",
     "@google-cloud/bigquery": "^7.6.0",
-    "firebase-admin": "^13.6.0",
+    "firebase-admin": "^14.2.0",
     "firebase-functions": "^7.3.3-rc.0",
     "generate-schema": "^2.6.0",
     "lodash": "^4.17.14"
+  },
+  "overrides": {
+    "@firebaseextensions/firestore-bigquery-change-tracker": {
+      "firebase-functions": "$firebase-functions"
+    }
   },
   "devDependencies": {
     "@types/lodash": "^4.17.0",
--- a/kits/firestore-bigquery-export/npm-shrinkwrap.json
+++ b/kits/firestore-bigquery-export/npm-shrinkwrap.json
@@ -9,9 +9,9 @@
       "version": "0.0.2-rc.6",
       "license": "Apache-2.0",
       "dependencies": {
-        "@firebaseextensions/firestore-bigquery-change-tracker": "^2.1.0",
+        "@firebaseextensions/firestore-bigquery-change-tracker": "^2.2.1",
         "@google-cloud/bigquery": "^7.6.0",
-        "firebase-admin": "^13.6.0",
+        "firebase-admin": "^14.2.0",
         "firebase-functions": "^7.3.3-rc.0",
         "generate-schema": "^2.6.0",
         "lodash": "^4.17.14"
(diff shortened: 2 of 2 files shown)


kits/firestore-bigquery-export/package.json
```diff
@@ -33,7 +33,7 @@
   "dependencies": {
     "@firebaseextensions/firestore-bigquery-change-tracker": "^2.1.0",
     "@google-cloud/bigquery": "^7.6.0",
-    "firebase-admin": "^13.6.0",
+    "firebase-admin": "^14.2.0",
```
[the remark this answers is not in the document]
cabljac: Confirmed and fixed. Tracker 2.2.0 (#3133, #3134) runs on admin 14; this layer now bumps to it and the shrinkwrap holds a single firebase-admin copy, checked by resolving firebase-admin/firestore from the tracker's directory and calling getFirestore after the kit's initializeApp.

kits/firestore-bigquery-export/package.json
```diff
@@ -31,13 +31,18 @@
     "serve": "firebase emulators:start --only functions"
   },
   "dependencies": {
-    "@firebaseextensions/firestore-bigquery-change-tracker": "^2.1.0",
+    "@firebaseextensions/firestore-bigquery-change-tracker": "^2.2.1",
     "@google-cloud/bigquery": "^7.6.0",
-    "firebase-admin": "^13.6.0",
+    "firebase-admin": "^14.2.0",
```
IzaakGough, line 36: Admin 14 also moves `@google-cloud/firestore` from 7.11.6 to 8.7.1 in the shrinkwrap, a major on the client that both the tracker's backup writes and the kit's own Firestore access go through. The description doesn't name it, and it's the largest unverified surface in the bump: the 62 tests all mock, and this kit has no `test:emulator` script, so CI's emulator step is skipped by `--if-present`. Worth listing next to the existing "not verified against a live deploy" line so the next reader knows where to look first.

kits/firestore-bigquery-export/package.json
```diff
@@ -31,13 +31,18 @@
     "serve": "firebase emulators:start --only functions"
   },
   "dependencies": {
-    "@firebaseextensions/firestore-bigquery-change-tracker": "^2.1.0",
+    "@firebaseextensions/firestore-bigquery-change-tracker": "^2.2.1",
```
IzaakGough, line 34: Small one: the description says the tracker moves to `^2.2.0`, this pins `^2.2.1`.

kits/firestore-bigquery-export/package.json
```diff
@@ -31,13 +31,18 @@
...
     "@google-cloud/bigquery": "^7.6.0",
-    "firebase-admin": "^13.6.0",
+    "firebase-admin": "^14.2.0",
     "firebase-functions": "^7.3.3-rc.0",
     "generate-schema": "^2.6.0",
     "lodash": "^4.17.14"
   },
+  "overrides": {
```
IzaakGough, line 41: Worth scoping what this block guarantees. npm ignores `overrides` coming from a non-root package, and the release workflow's prune step notes that kits are consumed as a published dependency whose shrinkwrap expands into the consumer's install. So the single `firebase-functions` copy at a consumer comes from the shrinkwrap, not from here. It does survive the prune (the entry isn't `dev`), so npm consumers are fine today.

pnpm and yarn ignore a dependency's `npm-shrinkwrap.json`, and would resolve the tracker's own `^6.3.2 || ^7.3.0` to a second copy, which loses exactly the trace context this is protecting. A line in the description scoping the guarantee to npm installs would make that clear.

IzaakGough (approved):
Approving. Checked the shrinkwrap holds single copies of `firebase-admin@14.3.0` and `firebase-functions@7.3.3-rc.0` with no nested tracker copies, and confirmed the override is genuinely required: tracker 2.2.1 declares `firebase-functions: "^6.3.2 || ^7.3.0"`, which a prerelease `7.3.3-rc.0` cannot satisfy under semver. Also confirmed npm does not record `overrides` in the lockfile, so its absence from the shrinkwrap root entry is expected rather than a stale lock.

The motivating behaviour is real: `firebase-admin/lib/functions/functions.js` reads `FIREBASE_KIT_INSTANCE_ID` in the task queue scope resolution. And the kit source only imports `firebase-admin/app` and `firebase-admin/eventarc`, both stable across the v14 major, so "no source changes" holds.

Three notes inline, none blocking.

kits/firestore-bigquery-export/package.json
cabljac: Landed in the description: named the `@google-cloud/firestore` 7.11.6 to 8.7.1 major, and replaced the unverified-live-deploy line. The top of the stack was deployed to a test project with CLI 15.29.0 and the backup-write path ran through admin 14 / firestore 8.7.1 (changelog table removed, trigger enqueued, backup docs written keyed by event id, queue drained after re-provisioning).

kits/firestore-bigquery-export/package.json
cabljac: Fixed, the description now says `^2.2.1`. The commit was amended for 2.2.1 after #3137 shipped and the body lagged.

kits/firestore-bigquery-export/package.json
cabljac: Agreed and added to the description: the single-copy guarantee comes from the published shrinkwrap and holds for npm installs only, with pnpm and yarn resolving the tracker's own range to a second functions copy. Also checked the mechanism: a registry install of the published rc.6 nests the kit's tree per its shrinkwrap (arborist keys this on the packument's `_hasShrinkwrap`), and the prune step keeps the entry since it is not `dev`.

PR #3128, opened 2026-09-07
feat(firestore-bigquery-export): add the sync task enqueue module

cabljac:
Second layer of the Cloud Tasks write buffer stack. Adds `src/tasks.ts` with `enqueueSyncTask` and `syncQueuePath`, and exports `SerializedDocumentChange` from the handlers as the task payload type. Nothing calls the module yet; the write path is wired in a later layer.

The queue path is deliberately unprefixed so firebase-admin resolves the kit instance prefix. The region comes from the CLI-set `FUNCTION_REGION` first, with `DATABASE_REGION` mapped through the dual-region table as the fallback for local runs; with neither set the bare function name is returned and the Admin SDK applies its default location, `us-central1`, which is also where the CLI places region-less functions. Enqueue retries in-process with exponential backoff and jitter, capped at 5s, and any attempt budget that is not a positive integer means a single attempt. The task id is the sha256 hex of the event id: hashed rather than sanitized so two ids differing only in punctuation cannot collapse into one task, and so ids are uniformly distributed as Cloud Tasks asks for named tasks. A retried enqueue of an event already in Cloud Tasks returns `functions/task-already-exists`, which is treated as success rather than buffering the event twice. Named tasks carry a lookup cost, but enqueue only runs when the inline insert has failed, so that cost sits on the failure path.

`tests/tasks.test.ts` pins the above with a mocked `firebase-admin/functions`. `tests/tasks.emulator.test.ts` runs the real firebase-admin 14.3.0 against a local `CLOUD_TASKS_EMULATOR_HOST` and asserts the kit-prefixed queue path in `FUNCTION_REGION`, the `us-central1` path with no region variables, and the hashed task name in the request body. 79 tests pass on this layer. No live queue has been exercised on this layer; the top of the stack was deployed to a test project and its enqueue path drained end to end (see #3127).

Part of #3031.

### DIFF
--- a/kits/firestore-bigquery-export/src/tasks.ts
+++ b/kits/firestore-bigquery-export/src/tasks.ts
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createHash } from "node:crypto";
+import { getFunctions } from "firebase-admin/functions";
+import type { SerializedDocumentChange } from "./handlers";
+import { firestoreLocationToFunctionRegion } from "./region";
+
+/** Export name of the write-buffer task function. */
+export const SYNC_BIGQUERY_FUNCTION = "syncBigQuery";
+
+const MAX_BACKOFF_MS = 5000;
+const BACKOFF_BASE_MS = 100;
+const JITTER_MS = 100;
+
+// Hashed rather than sanitized: a lossy mapping could collapse two event ids
+// into one task id, and Cloud Tasks wants ids uniformly distributed.
+function taskIdFor(change: SerializedDocumentChange): string {
+  return createHash("sha256").update(change.eventId).digest("hex");
+}
+
+/**
+ * Resolves the queue resource path for a task function of this kit instance.
+ *
+ * The name is deliberately unprefixed: firebase-admin >= 14.2.0 resolves the
+ * deployed `kit-<instance id>-` prefix itself from the
+ * `FIREBASE_KIT_INSTANCE_ID` env var, which the CLI sets on every deployed kit
+ * function. All functions of a kit instance deploy to one region, so the
+ * enqueuing function's own region is the queue's region.
+ *
+ * The CLI-set `FUNCTION_REGION` wins because it is the region the function
+ * was actually deployed to; a `DATABASE_REGION`-derived region can disagree
+ * with it on a first deploy or when the variable is unset, and is only the
+ * fallback for local runs where the CLI has not populated the environment.
+ * With neither set, the bare function name is returned and the Admin SDK
+ * applies its default location, `us-central1`, which is also where the CLI
+ * places functions that declare no region.
+ *
+ * @param functionName - The export name of the task function.
+ * @returns The queue resource path, `locations/<region>/functions/<name>`,
+ *   or the bare `<name>` when no region is known.
+ */
+export function syncQueuePath(
+  functionName: string = SYNC_BIGQUERY_FUNCTION
+): string {
+  const region =
+    process.env.FUNCTION_REGION ||
+    firestoreLocationToFunctionRegion(process.env.DATABASE_REGION);
+
+  return region
+    ? `locations/${region}/functions/${functionName}`
+    : functionName;
+}
+
+function backoffMs(attempt: number, jitter: number): number {
+  return (
+    Math.min(Math.pow(2, attempt) * BACKOFF_BASE_MS, MAX_BACKOFF_MS) + jitter
+  );
+}
+
+/**
+ * Enqueues a payload onto the `syncBigQuery` queue, retrying transient enqueue
+ * failures in-process with exponential backoff and jitter.
+ *
+ * The task id is derived from the event id, so a retried enqueue of an event
+ * that already reached Cloud Tasks is rejected rather than buffered twice.
+ *
+ * @param payload - The serialized change to enqueue.
+ * @param maxAttempts - How many enqueue attempts to make before giving up.
+ *   Anything but a positive integer means a single attempt: resolving without
+ *   an enqueue would report success for an event that was never buffered.
+ * @throws The last enqueue error, once every attempt has failed.
+ */
+export async function enqueueSyncTask(
+  payload: SerializedDocumentChange,
+  maxAttempts: number
+): Promise<void> {
+  const queue = getFunctions().taskQueue(syncQueuePath());
+  const id = taskIdFor(payload);
+
+  // Math.max(1, NaN) is NaN and would skip the loop entirely.
+  const attemptBudget =
+    Number.isInteger(maxAttempts) && maxAttempts >= 1 ? maxAttempts : 1;
+  const jitter = Math.random() * JITTER_MS;
+  let attempts = 0;
+
+  while (attempts < attemptBudget) {
+    if (attempts > 0) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, backoffMs(attempts, jitter))
+      );
+    }
+
+    attempts++;
+    try {
+      await queue.enqueue(payload, { id });
+      return;
+    } catch (enqueueErr) {
+      // The event is already buffered; a second task would double-write the row.
+      // firebase-admin prefixes its codes: `functions/task-already-exists`.
+      if (
+        (enqueueErr as { code?: string })?.code ===
+        "functions/task-already-exists"
+      ) {
+        return;
+      }
+
+      if (attempts >= attemptBudget) {
+        throw enqueueErr;
+      }
+    }
+  }
+}
--- a/kits/firestore-bigquery-export/tests/tasks.test.ts
+++ b/kits/firestore-bigquery-export/tests/tasks.test.ts
@@ -1,0 +1,229 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createHash } from "node:crypto";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+vi.mock("firebase-admin/functions", () => ({
+  getFunctions: vi.fn(),
+}));
+
+import { getFunctions } from "firebase-admin/functions";
+import type { SerializedDocumentChange } from "../src/handlers";
+import { enqueueSyncTask, syncQueuePath } from "../src/tasks";
+
+function sha256(value: string): string {
+  return createHash("sha256").update(value).digest("hex");
+}
+
+function makeChange(
+  overrides: Partial<SerializedDocumentChange> = {}
+): SerializedDocumentChange {
+  return {
+    timestamp: "2026-01-01T00:00:00.000Z",
+    eventId: "evt-1",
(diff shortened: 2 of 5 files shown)


kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  return `locations/${region}/functions/${functionName}`;
+}
+
+function backoffMs(attempt: number, jitter: number): number {
+  return (
+    Math.min(Math.pow(2, attempt) * BACKOFF_BASE_MS, MAX_BACKOFF_MS) + jitter
+  );
+}
```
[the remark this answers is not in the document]
cabljac: Kept as is. The extension computes the jitter once per enqueue and reuses it across attempts (functions/src/index.ts:298), and this module ports that on purpose. Per-attempt jitter would be a behaviour change for a three-attempt in-process loop with a 5s cap, where herd effects are not a concern.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+    Number.isInteger(maxAttempts) && maxAttempts >= 1 ? maxAttempts : 1;
+  const jitter = Math.random() * JITTER_MS;
+  let attempts = 0;
+
+  while (attempts < attemptBudget) {
+    if (attempts > 0) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, backoffMs(attempts, jitter))
+      );
+    }
```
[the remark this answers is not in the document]
cabljac: Same as above: intentional parity with the extension.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+export async function enqueueSyncTask(
+  payload: SerializedDocumentChange,
+  maxAttempts: number
+): Promise<void> {
+  const queue = getFunctions().taskQueue(syncQueuePath());
+  const id = taskIdFor(payload);
+
+  // Math.max(1, NaN) is NaN and would skip the loop entirely.
+  const attemptBudget =
+    Number.isInteger(maxAttempts) && maxAttempts >= 1 ? maxAttempts : 1;
```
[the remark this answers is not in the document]
cabljac: Not reachable. maxAttempts comes from MAX_ENQUEUE_ATTEMPTS, validated to an integer 1 to 10 at deploy and defaulted to 3 in resolveExportConfig; nothing passes Infinity. The clamp exists so a non-integer cannot skip the loop and report success without an enqueue.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+export function syncQueuePath(
+  functionName: string = SYNC_BIGQUERY_FUNCTION
+): string {
+  const region =
+    process.env.FUNCTION_REGION ||
+    firestoreLocationToFunctionRegion(process.env.DATABASE_REGION);
+
+  if (!region) {
+    throw new Error(
```
IzaakGough: With neither variable set this throws, but `index.ts` declares no region in exactly that case, so the functions deploy to the CLI default. The README says the same ("or `us-central1` if `DATABASE_REGION` is unset and you did not override the deploy region"), and `tests/index.test.ts` pins it. The Admin SDK also resolves a location-less queue name to `us-central1` on its own: with a local `CLOUD_TASKS_EMULATOR_HOST`, `getFunctions().taskQueue("syncBigQuery")` produces `/projects/test-project/locations/us-central1/queues/kit-default-syncBigQuery/tasks`.

Nothing calls this module yet so it is harmless at this commit, but once the write path is wired, an environment without `FUNCTION_REGION` (library consumers, emulator runs) will fail every document write instead of using the region the queue is actually in. That is the shape of the `checkImageContent` regression in `storage-resize-images`. Defaulting to `us-central1`, or dropping the location segment and letting the SDK default, would match where the functions land.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const MAX_BACKOFF_MS = 5000;
+const BACKOFF_BASE_MS = 100;
+const JITTER_MS = 100;
+
+/** Cloud Tasks accepts `[A-Za-z0-9_-]{1,500}` as a task id. */
+const TASK_ID_DISALLOWED = /[^A-Za-z0-9_-]/g;
+
+function taskIdFor(change: SerializedDocumentChange): string {
+  return change.eventId.replace(TASK_ID_DISALLOWED, "-").slice(0, 500);
```
IzaakGough: Every character outside `[A-Za-z0-9_-]` maps to `-`, so two event ids that differ only in punctuation produce the same task id. Against the real SDK, `a.b` and `a/b` both give `.../tasks/a-b`. The second enqueue then returns `functions/task-already-exists`, which `enqueueSyncTask` treats as success, so that change never reaches BigQuery and nothing logs it. `slice(0, 500)` has the same effect for ids that differ only past character 500.

Whether this is reachable depends on the format of the Eventarc Firestore CloudEvent `id`, which I have not seen a live one of. Hashing the event id (sha256 hex) sidesteps both the collapse and the truncation, and it also helps with the distribution point below.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+  while (attempts < attemptBudget) {
+    if (attempts > 0) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, backoffMs(attempts, jitter))
+      );
+    }
+
+    attempts++;
+    try {
+      await queue.enqueue(payload, { id });
```
IzaakGough, line 109: Worth a decision rather than a change. The Cloud Tasks `tasks.create` reference notes a cost for named tasks: "Because there is an extra lookup cost to identify duplicate task names, these calls have significantly increased latency", and "Choosing task ids that are sequential or have sequential prefixes, for example using a timestamp, causes an increase in latency and error rates in all task commands", because "the infrastructure relies on an approximately uniform distribution of task ids". It also puts the id reuse window at up to 24 hours.

For a streaming export this runs on every document write, so it would be good to know whether Firestore event ids are uniformly distributed before the write path lands. Hashing the id gives that for free.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+export async function enqueueSyncTask(
+  payload: SerializedDocumentChange,
+  maxAttempts: number
+): Promise<void> {
+  const queue = getFunctions().taskQueue(syncQueuePath());
+  const id = taskIdFor(payload);
+
+  // Math.max(1, NaN) is NaN and would skip the loop entirely.
+  const attemptBudget =
```
IzaakGough, line 95: This replaces rather than clamps: `10.5` and `Infinity` both become a single attempt, while the doc comment says "Clamped to at least 1". `Math.max(1, Math.floor(maxAttempts) || 1)` matches what the comment describes. Only matters if the later layer feeds it a non-integer.

kits/firestore-bigquery-export/src/tasks.ts
```diff
@@ -0,0 +1,128 @@
...
+      // The event is already buffered; a second task would double-write the row.
+      // firebase-admin prefixes its codes: `functions/task-already-exists`.
+      if (
+        (enqueueErr as { code?: string })?.code ===
+        "functions/task-already-exists"
+      ) {
+        return;
+      }
+
+      if (attempts >= attemptBudget) {
```
IzaakGough, line 121: Only `task-already-exists` short-circuits, so a permanent failure such as `functions/invalid-argument` (an id the SDK rejects, an oversized payload) retries with backoff toward a certain failure and adds a few seconds to an invocation that fails regardless.

kits/firestore-bigquery-export/tests/tasks.test.ts
```diff
@@ -0,0 +1,212 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+vi.mock("firebase-admin/functions", () => ({
```
IzaakGough, line 20: The prefix behaviour lives in the SDK's `resolveResourceId`, which this mock replaces, so the assertion that `taskQueue` was called with `locations/us-central1/functions/syncBigQuery` would pass just as well if that string were wrong for the SDK.

`kits/firestore-vector-search/tests/task-queues.test.ts` is the harness for this: real `firebase-admin`, `CLOUD_TASKS_EMULATOR_HOST` pointed at a local server, assert the resulting URL. Running that shape against this code produces `/projects/test-project/locations/us-central1/queues/kit-default-syncBigQuery/tasks` with `task.name` ending `/evt-1`, so the path is right today. There is just nothing here that would catch it regressing.

IzaakGough (commented):
Read through the module and ran the suite (75 pass). I also pointed a local `CLOUD_TASKS_EMULATOR_HOST` at `enqueueSyncTask` with real `firebase-admin` 14.3.0 to check the central premise, and it holds: the unprefixed path resolves to `queues/kit-default-syncBigQuery`, and `functions/task-already-exists` is a real code (`FirebaseFunctionsError` prefixes `functions/`, thrown on HTTP 409).

Nothing here is blocking for this PR, since nothing calls `tasks.ts` yet and it is not exported from `lib.ts` or `index.ts`. Two comments are worth settling before the layer that wires the write path: the task id sanitization, which can silently drop a row, and the region throw, which would fail every write where `FUNCTION_REGION` is unset. The rest are a design question, a test coverage note, and two nits.

kits/firestore-bigquery-export/src/tasks.ts
cabljac: Landed. With neither variable set `syncQueuePath` now returns the bare function name and the Admin SDK applies its default location, `us-central1`, which is also where the CLI places region-less functions. The two throw tests became bare-name tests, and a new real-SDK test (`tests/tasks.emulator.test.ts`, `CLOUD_TASKS_EMULATOR_HOST`) asserts the resulting queue path for both the `FUNCTION_REGION` and the no-region case.

kits/firestore-bigquery-export/src/tasks.ts
cabljac: Landed: the task id is now the sha256 hex of the event id. For the record, the live Eventarc Firestore ids from a test deploy yesterday were UUIDs (`57815e20-b745-45af-9fd8-0093ed2066ac`), so the collapse was not reachable with today's format, but hashing removes both the assumption and the 500-char cap. Unit test covers `a.b` vs `a/b` giving distinct ids; the emulator-host test checks the task name the SDK actually sends.

kits/firestore-bigquery-export/src/tasks.ts
cabljac: Decision: keep named tasks. One correction to the premise: enqueue does not run on every document write, only when the inline BigQuery insert fails (`handlers.ts`, the catch around `recordEventToBigQuery`), so the named-task lookup latency sits on the failure path. What the name buys is deterministic dedupe across the in-process enqueue retries: an enqueue that timed out client-side but landed server-side would otherwise buffer the row twice, and BigQuery's `insertId` dedupe is best-effort. The sha256 id gives the uniform distribution the reference asks for, and the 24h reuse window is fine since the same event id is the same change.

kits/firestore-bigquery-export/src/tasks.ts
cabljac: Reworded the doc comment to describe what the code does: anything but a positive integer means a single attempt. Not switching to `Math.max(1, Math.floor(x) || 1)`: `Infinity` survives that and makes the loop unbounded. The later layer feeds it `maxEnqueueAttempts`, a validated 1-10 int param defaulting to 3, so non-integers do not reach it in practice.

kits/firestore-bigquery-export/src/tasks.ts
cabljac: Kept, for parity: the extension's `attemptToEnqueue` retries every error up to `MAX_ENQUEUE_ATTEMPTS` with the same backoff. With the default of 3 the cost is 200ms + 400ms plus jitter, under a second, on an execution that is already on the failure path. Enumerating permanent SDK codes to short-circuit would be a new behaviour with its own maintenance, and `task-already-exists` is the one code where the outcome differs (success, not failure).

kits/firestore-bigquery-export/tests/tasks.test.ts
cabljac: Landed `tests/tasks.emulator.test.ts`, the vector-search harness shape: real firebase-admin 14.3.0, `CLOUD_TASKS_EMULATOR_HOST` at a local server, `FIREBASE_KIT_INSTANCE_ID=test-instance`. It asserts `/projects/test-project/locations/europe-west2/queues/kit-test-instance-syncBigQuery/tasks` with `FUNCTION_REGION` set, the `us-central1` path with no region variables, and the hashed task name in the request body.

IzaakGough (approved):

PR #3130, opened 2026-09-07
feat(firestore-bigquery-export): reinstate the Cloud Tasks write buffer

cabljac:
Fourth layer of the Cloud Tasks write buffer stack and the behaviour change, per the option D decision on #3031: migrating users keep the failure behaviour and cost profile they have today. Supersedes #3103.

A failed inline write now enqueues onto a new `syncBigQuery` task queue (5 attempts, 60s minimum backoff, throttled by `MAX_DISPATCHES_PER_SECOND`) after up to `MAX_ENQUEUE_ATTEMPTS` in-process attempts, instead of self-healing once and rethrowing to the trigger retry policy. The task handler re-attempts the write and rethrows so Cloud Tasks retries; the tracker parks rows whose insert is rejected in `BACKUP_COLLECTION`, on the inline attempt and on each queue attempt. A change that cannot be enqueued at all is logged first, then published as `onError`, and dropped, as the extension does; the log-before-publish order is deliberate so the only trace of a dropped row survives a failing events channel. The trigger drops its retry policy. The function needs `roles/cloudtasks.enqueuer`. The queue function sets `maxInstances` to 500 to match the gen1 handler, since the gen2 default of 100 would 429 dispatches and burn attempts. The queue's `rateLimits.maxDispatchesPerSecond` is the CEL ternary `params.MAX_DISPATCHES_PER_SECOND < 1 ? 100 : params.MAX_DISPATCHES_PER_SECOND`: a blank `.env` value reaches deploy-time CEL as 0, and the ternary restores the same default the runtime falls back to. `REQUIRED_APIS` keeps `firestore.googleapis.com` from #3119, now pinned by `tests/required-apis.test.ts`.

One deliberate change from the extension: the success event is published after the insert and swallowed on failure, closing a duplicate-row route. The README rewrite for the new export and failure path is the next layer, #3131.

96 tests pass on this layer. Verified live on a test project with Firebase CLI 15.29.0 from the top of the stack: the deployed env carries `FIREBASE_KIT_INSTANCE_ID` and `FUNCTION_REGION`, the gen2 task function accepted the SDK's ID token, the queue was created with the resolved throttle, and with the changelog table removed the trigger enqueued, the tracker wrote backup documents keyed by event id, and the queue drained without duplicate rows once the table was re-provisioned. Not exercised live: the Cloud Tasks task size limit. A change near Firestore's 1 MB document cap, doubled by `oldData` on an update, exceeds the task body limit, so the enqueue fails and the row is logged and dropped. This is parity: the extension enqueues the same payload through the same path, and `EXCLUDE_OLD_DATA` is its mitigation. Listed under known limits in the README in #3131.

Fixes #3031.


kits/firestore-bigquery-export/src/handlers.ts
```diff
@@ -55,11 +60,15 @@ export interface HandlerContext {
   tracker: FirestoreBigQueryEventHistoryTracker;
   config: ResolvedExportConfig;
```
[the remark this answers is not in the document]
cabljac: Kept. The backup write is keyed by insertId, which is the event id, and uses set, so repeated attempts overwrite one document rather than adding entries. Writing on every terminal attempt is the extension's behaviour and the README documents the rows as possibly failed for that reason. A second tracker instance to gate the backup on retryCount would be a divergence.

kits/firestore-bigquery-export/src/handlers.ts
```diff
@@ -204,7 +212,69 @@ export async function handleDocumentWrite(
...
+
+  logs.logEventAction(
+    "Firestore event received by onDispatch trigger",
+    change.fullResourceName,
+    change.eventId,
+    change.changeType
+  );
+
+  try {
+    await recordEventToBigQuery(change, ctx.tracker);
```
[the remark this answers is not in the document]
cabljac: See above: single document per event, no duplicates, parity with the extension.

kits/firestore-bigquery-export/tests/handlers.test.ts
```diff
@@ -70,6 +73,33 @@ function makeCtx(
     tracker: tracker as unknown as HandlerContext["tracker"],
     config: config as HandlerContext["config"],
```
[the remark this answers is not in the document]
cabljac: Not needed, no terminalTracker is being added.

kits/firestore-bigquery-export/src/index.ts
```diff
@@ -55,19 +61,22 @@ const LIFECYCLE_RETRY_CONFIG = {
...
   "roles/bigquery.user",
   // Gen2 Firestore triggers need Eventarc receive and run.invoker on the function SA.
   "roles/eventarc.eventReceiver",
   "roles/run.invoker",
+  // The trigger enqueues failed writes onto its own syncBigQuery task queue.
+  "roles/cloudtasks.enqueuer",
 ];
 const REQUIRED_APIS = [
-  {
-    api: "firestore.googleapis.com",
```
IzaakGough, line 68: This drops the `firestore.googleapis.com` entry that both the base branch and `next` carry (added in `eceaae0e`, merged as #3119). Nothing in the description or CHANGELOG mentions removing it, so it looks like a conflict resolution that went the wrong way.

On a project where the Firestore API is not already enabled, `firebase deploy` no longer enables it and the Firestore trigger fails to deploy. `tests/index.test.ts` mocks `requiresAPI` without asserting on it, so nothing catches this. `kits/firestore-incremental-capture/tests/required-apis.test.ts` is a pattern worth copying here.

kits/firestore-bigquery-export/src/config.ts
```diff
@@ -289,7 +289,7 @@ const params = {
   backupCollection: defineString("BACKUP_COLLECTION", {
     label: "Backup Collection Name",
     description:
-      "This (optional) parameter will allow you to specify a collection for which failed BigQuery updates will be written to.",
+      "Strongly recommended. The Firestore collection where rows that still fail after the sync queue's five attempts are written; without it, those rows are dropped. See the README for how to reconcile backed-up rows into BigQuery.",
```
IzaakGough: The description promises that rows "that still fail after the sync queue's five attempts" are written here, but the backup only happens inside the tracker's `insertData`, so only for rows that reached BigQuery and were rejected. A row whose *enqueue* fails never reaches the queue or `insertData`, and is dropped with no backup.

The concrete case is the task size limit: `enqueueSyncTask` base64-encodes `{data: change}`, and an update carries both `data` and `oldData`, so a document approaching Firestore's 1 MB limit produces a task body over the Cloud Tasks cap. That is a non-retryable INVALID_ARGUMENT, so all three enqueue attempts burn (with the backoff sleeps) and the row is lost even with `BACKUP_COLLECTION` set. Worth scoping the description to insert failures either way, given the size limit is listed as unverified.

IzaakGough (commented):
Read every hunk and ran the suite (90 pass). The parity claims check out against the extension: the same `maxAttempts: 5` / `minBackoffSeconds: 60` retry config and `maxConcurrentDispatches: 500`, the same `skipInit: true` with no provisioning on the queue path, no inline `onSuccess`, and `roles/cloudtasks.enqueuer` is enough on its own (admin sets an `Authorization` ID-token header for kit-scoped enqueues rather than an `oidcToken`, so no `actAs`). Backup writes use `.doc(row.insertId)`, so repeated queue attempts do not duplicate backup docs.

Inline comments below, plus one that has no diff line to hang off:

**README.md is not updated for the new export.** "Export the three functions from your functions codebase entry" and the block at lines 45-49 still list only `fsexportbigquery`, `initBigQuerySync`, `setupBigQuerySync`, and lines 61-63, 90-91 and 324 repeat that list. The CLI deploys only what the entry file exports, so someone following the README deploys no `syncBigQuery` function and no queue. Every failed inline write then hits `enqueueSyncTask`, gets NOT_FOUND from Cloud Tasks on all `MAX_ENQUEUE_ATTEMPTS` tries, and the row is logged and dropped, which is the data loss this PR exists to prevent, while looking like the buffer is working. The migration sections are inverted too: lines 239-247 say "This kit has no task queue on the write path... `MAX_DISPATCHES_PER_SECOND` and `MAX_ENQUEUE_ATTEMPTS` no longer exist", and lines 251-253 say "`onSuccess` is no longer published... the task queue handler is gone".

kits/firestore-bigquery-export/src/handlers.ts
```diff
@@ -87,38 +96,37 @@ async function recordEventToBigQuery(
...
   ctx: HandlerContext
 ): Promise<void> {
   try {
-    await ctx.ensureInitialized();
-    await recordEventToBigQuery(change, ctx.tracker);
-  } catch (retryErr) {
-    await events.recordErrorEvent(retryErr as Error);
+    await ctx.enqueue(change);
+  } catch (enqueueErr) {
+    await events.recordErrorEvent(enqueueErr as Error);
```
IzaakGough: Minor, and the same ordering the extension has, so not a regression: awaiting `recordErrorEvent` before `logFailedEventAction` means that if publishing the error event rejects (Eventarc channel misconfigured or unavailable), the rejection propagates out of `handleDocumentWrite`, the "Failed to enqueue event to Cloud Tasks" line is never written, and with no retry policy the execution just fails. The only diagnostic for a dropped row disappears exactly when the events channel is unhealthy. Swapping the order, or wrapping the publish, keeps the log.

kits/firestore-bigquery-export/src/handlers.ts
```diff
@@ -134,8 +142,8 @@ export async function handleDocumentWrite(
 
   // No provisioning on the hot path: BigQuery resources are provisioned
   // out-of-band (afterFirstDeploy / afterRedeploy tasks). If they are missing,
-  // the inline write fails, self-heals once, then falls back to the trigger
-  // retry policy.
+  // the inline write fails and the change buffers through the syncBigQuery
```
IzaakGough: This reads as though a missing dataset or table eventually recovers through the queue, but it cannot: `toTrackerConfig` sets `skipInit: true`, so `tracker.record()` never provisions, and `handleSyncBigQueryTask` deliberately skips `ensureInitialized`. If `afterFirstDeploy` failed permanently or the dataset was deleted, all five queue attempts fail the same way and every row is lost unless `BACKUP_COLLECTION` is set.

That is the intended design and matches the extension, so just a rewording so a later reader does not count on a repair path that is not there.

kits/firestore-bigquery-export/src/index.ts
cabljac: Right, and no reason behind it: #3103 predates #3119, and the split kept #3103's file content, so the rebase took the older `REQUIRED_APIS`. Restored the `firestore.googleapis.com` entry with the base branch's reason text, and added `tests/required-apis.test.ts` on the incremental-capture pattern: it asserts both APIs by name and reason and that exactly two are declared.

kits/firestore-bigquery-export/src/config.ts
cabljac: Agreed. The description now says: rows whose BigQuery insert is rejected, on the inline attempt and on each queue attempt, and that a change which cannot be enqueued at all is not backed up. The README's failure-handling section in #3131 already lists a failed enqueue as logged and dropped; the task size limit stays listed as unverified.

kits/firestore-bigquery-export/src/handlers.ts
cabljac: Swapped: the log line is written first, then `recordErrorEvent`. A rejected publish still propagates as before, but the only trace of the dropped row is on disk by then.

kits/firestore-bigquery-export/src/handlers.ts
cabljac: Reworded. The comment now says there is no provisioning on the queue path either, that a missing dataset or table fails the inline write and every queue attempt the same way until a redeploy runs the lifecycle task, and that `BACKUP_COLLECTION` is the only net.

cabljac:
On the README: the rewrite lands in the next layer, #3131 (`docs/kits-fbe-failure-handling`). It already says "Export the four functions" with `syncBigQuery` in the block, replaces the migration sections that claimed there is no queue, and documents the failure path, recovery of parked rows, and the known limits. Reviewing this layer's README on its own does show the stale text; the stack merges bottom-up and #3131 sits directly above.

IzaakGough (approved):

PR #3131, opened 2026-09-07
docs(firestore-bigquery-export): document failure handling and recovery

cabljac:
Last layer of the Cloud Tasks write buffer stack. README only.

Adds a "Failure handling" section describing the sync queue, how to recover rows parked in `BACKUP_COLLECTION` back into BigQuery, and the known limits that have not been verified live. Adds the two new params to the settings table, `syncBigQuery` to the list of functions to re-export, `roles/cloudtasks.enqueuer` to the roles list, and updates the differences section so failed writes are described as the same buffer the extension has rather than a trigger retry.

Part of #3031.

### DIFF
--- a/kits/firestore-bigquery-export/README.md
+++ b/kits/firestore-bigquery-export/README.md
@@ -5,9 +5,10 @@ BigQuery Firebase Extension as an npm package you add to your own Firebase
 Functions codebase and deploy.
 
 It listens for document writes on a collection, serializes each change, and
-writes it to a BigQuery changelog table. Failed writes are retried through a
-Firebase Functions runtime retry policy. The functions run in your own Firebase
-project; there is no hosted version, so you deploy them yourself.
+writes it to a BigQuery changelog table. Failed writes buffer through a Cloud
+Tasks queue (`syncBigQuery`), which retries them on its own throttled schedule.
+The functions run in your own Firebase project; there is no hosted version, so
+you deploy them yourself.
 
 ## Install
 
@@ -30,6 +31,7 @@ conflicts with that automatic setup.
 | `roles/datastore.user`         | write failed-row records back to Firestore (only if you configure a backup collection) |
 | `roles/eventarc.eventReceiver` | receive Gen2 Firestore trigger events                                                  |
 | `roles/run.invoker`            | allow Eventarc to invoke the Gen2 Cloud Run service                                    |
+| `roles/cloudtasks.enqueuer`    | enqueue failed writes onto the kit's own `syncBigQuery` task queue                     |
 | `bigquery.googleapis.com`      | mirror Firestore collection changes in BigQuery                                        |
 
 If the dataset lives in a different project (`BIGQUERY_PROJECT_ID`), grant the
@@ -38,12 +40,13 @@ CMEK dataset, also grant the BigQuery service account access to your KMS key.
 
 ## Usage
 
-Export the three functions from your functions codebase entry:
+Export the four functions from your functions codebase entry:
 
 ```ts
 // functions/src/index.ts
 export {
   fsexportbigquery,
+  syncBigQuery,
   initBigQuerySync,
   setupBigQuerySync,
 } from "@firebase-function-kits/firestore-bigquery-export";
@@ -59,6 +62,7 @@ DATABASE_REGION=europe-west2
 ```
 
 - `fsexportbigquery` is the Firestore trigger.
+- `syncBigQuery` is the write-buffer task queue that retries failed writes.
 - `initBigQuerySync` is the first-deploy provisioning lifecycle task.
 - `setupBigQuerySync` is the reconfigure provisioning lifecycle task.
 
@@ -87,8 +91,15 @@ later, behind the `kits` experiment):
 `instances` maps each instance id to the directory (relative to
 `firebase.json`) holding that instance's `.env`. The CLI prefixes every
 function and task queue name with `kit-<instance id>-`, so the functions above
-deploy as `kit-default-fsexportbigquery`, `kit-default-initBigQuerySync`, and
-`kit-default-setupBigQuerySync`.
+deploy as `kit-default-fsexportbigquery`, `kit-default-syncBigQuery`,
+`kit-default-initBigQuerySync`, and `kit-default-setupBigQuerySync`.
+
+Deploy with Firebase CLI 15.28.0 or later: it sets the
+`FIREBASE_KIT_INSTANCE_ID` env var on the deployed functions, which the trigger
+needs to address its own `syncBigQuery` queue. On functions deployed with an
+older CLI, enqueues fail (logged at error level and published as an `onError`
+event; the event is dropped, as in the extension) until you redeploy with a
+newer CLI.
 
 ```sh
 firebase experiments:enable kits
@@ -112,7 +123,9 @@ loads them at deploy time and prompts for any required values that are missing.
 | `datasetLocation`                | `DATASET_LOCATION`                  | no       | `us`               | BigQuery dataset location                                          |
 | `database`                       | `DATABASE`                          | no       | `(default)`        | Firestore database id                                              |
 | `bigqueryProjectId`              | `BIGQUERY_PROJECT_ID`               | no       | project id         | Dataset project, if different                                      |
-| `backupCollection`               | `BACKUP_COLLECTION`                 | no       | (empty)            | Firestore collection for failed rows                               |
+| `backupCollection`               | `BACKUP_COLLECTION`                 | no       | (empty)            | Strongly recommended: collection for rows whose BigQuery insert failed |
+| `maxDispatchesPerSecond`         | `MAX_DISPATCHES_PER_SECOND`         | no       | `100`              | `syncBigQuery` queue dispatch rate (1-500)                         |
+| `maxEnqueueAttempts`             | `MAX_ENQUEUE_ATTEMPTS`              | no       | `3`                | In-process enqueue attempts before giving up (1-10)                |
 | `transformFunction`              | `TRANSFORM_FUNCTION`                | no       | (empty)            | Optional transform Cloud Function                                  |
 | `tablePartitioning`              | `TABLE_PARTITIONING`                | no       | `NONE`             | Table partitioning strategy                                        |
 | `timePartitioningField`          | `TIME_PARTITIONING_FIELD`           | no       | (empty)            | Time-partitioning column name                                      |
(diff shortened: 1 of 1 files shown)


kits/firestore-bigquery-export/README.md
```diff
@@ -216,10 +229,61 @@ curl -fsS -X POST -H "Content-Type: application/json" -d '{"data":{}}' \
...
+   extension: a failure _before_ the write is attempted (serializing the
+   change, publishing the `onStart` event) fails the execution once and the
+   event is not redelivered.
+3. `syncBigQuery` re-attempts the write on the queue's schedule: 5 attempts,
+   60 seconds minimum backoff, throttled to `MAX_DISPATCHES_PER_SECOND`
+   dispatches per second (500 concurrent max, and the function allows 500
+   instances so that ceiling is reachable; gen2 would otherwise cap at 100).
+4. On every terminal insert failure the tracker writes the row to
+   `BACKUP_COLLECTION` (when configured), keyed by the event id, before the
+   task fails. After the fifth attempt the task is dropped. **Without a backup collection, the row is dropped with it** -
```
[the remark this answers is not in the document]
cabljac: Kept: the task is the subject of the same sentence, so the referent is clear.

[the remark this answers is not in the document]
cabljac: Correction: landed, reads better.

kits/firestore-bigquery-export/README.md
```diff
@@ -154,11 +167,11 @@ the instances cannot collide.
...
 ## Events
 
-When `EVENTARC_CHANNEL` is configured, the function publishes `onStart` and
-`onError` lifecycle events under
-`firebase.extensions.firestore-bigquery-export.v1.*`. The extension's
-`onSuccess` event is not published; see the events entry under
-"Differences from the Stream Firestore to BigQuery extension" below.
+When `EVENTARC_CHANNEL` is configured, the functions publish lifecycle events
+under `firebase.extensions.firestore-bigquery-export.v1.*`: `onStart` and
+`onError` from the write path, and `onSuccess` from the `syncBigQuery` task
```
IzaakGough, line 174: One thing this section leaves out: publication is filtered by `EXT_SELECTED_EVENTS`. `events.ts` passes it as the channel's `allowedEventTypes`, firebase-admin splits it on commas (`eventarc.js`), then filters every publish with an exact `Array.includes` (`eventarc-client-internal.js`). So a type missing from the list is dropped silently, and an empty value becomes `[""]` after the split, which suppresses every event.

That matters for a migrated config. `ext:export` writes `EXT_SELECTED_EVENTS` alongside `EVENTARC_CHANNEL`, and the extension offered both namespaces as separate choices, so a value listing only the `firestore-counter.v1.*` types means nothing publishes at all, and one that omits `...v1.onSuccess` means the event described here never arrives. The differences section covers repointing Eventarc triggers, which is the subscribe side, so the publish-side allowlist is the piece with nothing written about it.

Worth a sentence here, plus rows for `EVENTARC_CHANNEL` and `EXT_SELECTED_EVENTS` in the settings table, since neither appears there today. How likely a given migrated value is to be missing the new types I cannot say, that depends on when the user installed and what they ticked. The filtering mechanism itself is from reading firebase-admin 14.2.0 and firebase-tools.

kits/firestore-bigquery-export/README.md
```diff
@@ -216,10 +229,66 @@ curl -fsS -X POST -H "Content-Type: application/json" -d '{"data":{}}' \
...
+   (up to `MAX_ENQUEUE_ATTEMPTS` in-process attempts with backoff, keyed by
+   event id so a retried enqueue cannot buffer the same event twice) and the
+   execution succeeds. The trigger declares no retry policy, as in the
+   extension: a failure _before_ the write is attempted (serializing the
+   change, publishing the `onStart` event) fails the execution once and the
+   event is not redelivered.
+3. `syncBigQuery` re-attempts the write on the queue's schedule: 5 attempts,
+   60 seconds minimum backoff, throttled to `MAX_DISPATCHES_PER_SECOND`
+   dispatches per second (500 concurrent max, and the function allows 500
+   instances so that ceiling is reachable; gen2 would otherwise cap at 100).
```
IzaakGough: The parenthetical rationale does not hold for gen2. `onTaskDispatched` sets no `concurrency` (the string does not appear in `firebase-functions/lib/v2/providers/tasks.js`), so the deploy defaults apply: `memoryToGen2Cpu(256)` is 1, and `concurrency` defaults to 80 when cpu >= 1 (`resolveCpuAndConcurrency` in firebase-tools, applied to every non-gcfv1 endpoint, task queues included). At 80 requests per instance, even the default 100 instance cap covers around 8,000 concurrent requests, well above `maxConcurrentDispatches: 500`.

One request per instance is gen1 behaviour, which is where 500 makes sense for the extension's handler. Raising `maxInstances` does no harm, but as written this (and the matching comment in `index.ts`) leaves a reader thinking gen2 caps concurrent dispatch at 100. Dropping the parenthetical would be the simplest fix. Read from the published packages, not deployed.

kits/firestore-bigquery-export/README.md
```diff
@@ -112,7 +123,9 @@ loads them at deploy time and prompts for any required values that are missing.
 | `datasetLocation`                | `DATASET_LOCATION`                  | no       | `us`               | BigQuery dataset location                                          |
 | `database`                       | `DATABASE`                          | no       | `(default)`        | Firestore database id                                              |
 | `bigqueryProjectId`              | `BIGQUERY_PROJECT_ID`               | no       | project id         | Dataset project, if different                                      |
-| `backupCollection`               | `BACKUP_COLLECTION`                 | no       | (empty)            | Firestore collection for failed rows                               |
+| `backupCollection`               | `BACKUP_COLLECTION`                 | no       | (empty)            | Strongly recommended: collection for rows the queue gave up on     |
```
IzaakGough: "rows the queue gave up on" reads as final loss, but the tracker parks a row on every terminal insert failure within a single `record()` call, so each queue attempt writes one (same doc id, so it overwrites rather than accumulating). Step 4 and the recovery section both describe it that way. Known limits bullet 2 ("Rows that exhaust the queue...") reads the same as this row does.

As worded, someone opening the collection after an outage will read every doc as a lost row, when many are from attempts that later succeeded. Something like "collection for rows whose BigQuery insert failed" would match the behaviour. From reading `insertData` in the published change tracker 2.2.1.

IzaakGough (commented):
Docs only, and I checked the claims against the code on the base branch plus the published `@firebaseextensions/firestore-bigquery-change-tracker@2.2.1`, `firebase-admin@14.2.0` and firebase-tools. The role, both new params, the queue shape (5 attempts, 60s backoff, dispatch throttling), the trigger having no retry policy, enqueue dedupe by task id, `onSuccess` coming only from the queue handler in both kit and extension, the queue handler not provisioning, and the older-CLI enqueue failure all check out.

Five comments inline. All of it is from reading source, nothing deployed, and I have said where a claim rests on documented behaviour rather than something I exercised.

kits/firestore-bigquery-export/README.md
```diff
@@ -216,10 +229,66 @@ curl -fsS -X POST -H "Content-Type: application/json" -d '{"data":{}}' \
...
+   execution succeeds: the event is dropped, exactly as the extension did in
+   this window.
+
+### Recovering parked rows
+
+Rows in `BACKUP_COLLECTION` are changelog-shaped documents, not plain document
+snapshots, so `fs-bq-import-collection` cannot consume them. Treat them as
+"possibly failed": a transient failure that later succeeded on retry also
+leaves one behind, and nothing cleans them up. To recover after an outage,
+load the backup docs' fields into a temp table and `MERGE` them into the
```
IzaakGough: The two facts needed to write this query are not here. In 2.2.1 each backup doc is `{ insertId, json: { event_id, timestamp, document_name, document_id, operation, data, old_data, path_params? }, error_details }`, so the changelog columns sit under `json` rather than at the top level, and the doc id is the event id. Anyone following this will get the shape wrong on the first attempt, and #3031 is a tracking issue, so it is not somewhere a recovery procedure stays findable once it closes.

Even two lines on the doc shape and a MERGE skeleton in the README would make this usable during an outage. Read from `handleFailedTransactions` and `record` in the published 2.2.1.

kits/firestore-bigquery-export/README.md
```diff
@@ -216,10 +229,66 @@ curl -fsS -X POST -H "Content-Type: application/json" -d '{"data":{}}' \
...
+changelog table with a `WHEN NOT MATCHED` condition on `event_id` (the
+anti-join is mandatory because of those stale rows). See
+[firebase/extensions#3031](https://github.com/firebase/extensions/issues/3031)
+for the full recipe.
+
+### Known limits
+
+- A task queue is a project-level resource created for each task function.
+  Deleting the functions (or moving them to another region) can leave the old
+  queue behind; delete it in the Cloud Tasks console if it lingers.
```
IzaakGough: This limit is accurate, and it can be more precise: on delete the CLI disables the queue rather than removing it (`fabricator.deleteTrigger` routes task-queue endpoints to `disableTaskQueue`), so what lingers appears as DISABLED in the console. Saying that gives the reader something specific to look for.

Separately, and as a suggestion rather than a defect, a bullet on duplicate `event_id` rows in the changelog might be worth adding. `insertId` dedupe on streaming inserts is best effort for about a minute and the queue's minimum backoff is 60s, so an insert that landed but reported an error can be written a second time by the retry. The comment in `handleSyncBigQueryTask` relies on the same reasoning, and the MERGE advice already assumes stale event ids. I have not reproduced this one, and the dedupe window is from Google's documentation rather than a live test.

IzaakGough (approved):

kits/firestore-bigquery-export/README.md
cabljac: Landed. The Events section now explains the `EXT_SELECTED_EVENTS` allowlist: split on commas, exact match, silent drop, empty value suppresses everything, another product's types publish nothing, and a migrated config brings its value along so check it lists the `firestore-bigquery-export.v1.*` types including `onSuccess`. Both `EVENTARC_CHANNEL` and `EXT_SELECTED_EVENTS` have rows in the settings table, marked env-only.

kits/firestore-bigquery-export/README.md
cabljac: Confirmed against the CLI: `resolveCpuAndConcurrency` gives 256 MB one CPU and `DEFAULT_CONCURRENCY` 80, and `tasks.js` sets no concurrency. So the instance cap bought nothing. Rather than keep it with a corrected rationale, #3130 now drops `maxInstances` entirely (parity with the extension's uncapped handler) and the test asserts it is undefined; this README's step 3 loses the parenthetical.

kits/firestore-bigquery-export/README.md
cabljac: Landed: the table row now reads "collection for rows whose BigQuery insert failed", and the known-limits bullet is "a row whose insert still fails on the last queue attempt with no `BACKUP_COLLECTION` configured is gone".

kits/firestore-bigquery-export/README.md
cabljac: Landed. The recovery section now shows the document shape from `handleFailedTransactions` in 2.2.1 (`insertId`, `json` holding the changelog columns, `error_details`, keyed by event id), says the columns sit under `json`, and carries a `MERGE ... WHEN NOT MATCHED` skeleton on `event_id` with a note to add `path_params` and partition columns. The link to #3031 is gone since the recipe lives here now.

kits/firestore-bigquery-export/README.md
cabljac: Both landed. The queue bullet now says the CLI disables rather than removes the queue and that it shows as `DISABLED` in the console (from `fabricator.deleteTrigger` routing task queues to `disableTaskQueue`). Added a duplicate `event_id` bullet: best-effort `insertId` dedupe of about a minute versus the 60 second minimum backoff, the latest view is unaffected because it keys on `document_name` and takes the newest change, and the MERGE assumes duplicates.

PR #3145, opened 2026-09-08
fix(firestore-bigquery-export): take the extension's yes/no values for snapshot syntax and old data

cabljac:
`USE_NEW_SNAPSHOT_QUERY_SYNTAX` and `EXCLUDE_OLD_DATA` were `defineBoolean` params, so only the literal `true` enabled them and the README told users to rewrite `yes` to `true` after a migration. The extension defines both as `yes` / `no` selects (`extension.yaml:575`, `:589`), and the migration exports those values verbatim, so a migrated `.env` silently left both off. Agreed direction in extension chat: stay with the closest match to the extension so users enter nothing during migration, and do better for new kits or new params.

Both are now `defineString` with a `select({ Yes: "yes", No: "no" })` input and default `no`, mirroring the extension's labels, values, and default, and `configFromEnv` maps them with a `yesNo` helper. Only `yes` enables, compared after trimming and lowercasing so the select label `Yes` and stray whitespace still work; `no` and anything else, `true` included, read as off. firebase-tools copies `.env` values verbatim without checking them against the select, which is why the label and whitespace are forgiven; `true` is not, since the kit has no users who could have written it and the extension's select never emits it. `WILDCARD_IDS` stays `defineBoolean`: the extension's select there is `true` / `false`, which `BooleanParam` already parses. The README's "Boolean settings use `true` / `false`" difference section is removed since the difference is gone.

Audit of the rest of the kit's params: no other boolean-shaped extension param exists. `delete-user-data` has the same shape on `ENABLE_AUTO_DISCOVERY` and is not part of this stack.

Tests: four new cases in `config.test.ts` cover the unset value, `yes`, the `Yes` and whitespace spellings, and `no` / `true`. The `defineString` mock now matches the real `StringParam` (`process.env[name] || ""`, never the declared default, which only the CLI prompt consults), and the one test that relied on a default now stubs the env var the way the CLI populates it. 100 tests and `tsc` pass at the top of the stack. Not verified against a live migration export.

Part of #3031.


kits/firestore-bigquery-export/src/config.ts
```diff
@@ -633,6 +635,11 @@ function normalizePositiveInt(value: string): number | undefined {
   return normalized > 0 ? normalized : undefined;
 }
 
+/** The extension's yes/no select: only the literal `yes` enables a setting. */
+function yesNo(value: string): boolean {
```
IzaakGough, line 640: `yesNo` treats `true` as off, and nothing surfaces that. firebase-tools copies `.env` values verbatim into param values without validating them against a `select` input (`resolveParams` in `lib/deploy/functions/params.js`: params present in `userEnvs` are assigned directly, and only prompted params go through `canSatisfyParam`). So someone who followed the README section this PR removes ("Change any `yes` to `true` in your `.env`") deploys clean and silently loses the setting: `EXCLUDE_OLD_DATA=true` starts writing `old_data` payloads again, bigger rows and more cost, with no error and no log line. A hand-written `Yes` copied from the select label does the same.

Accepting `true` / `false` here as well, or logging a warning for a non-empty value that is neither `yes` nor `no`, would make that case loud instead of silent. The extension's select can never emit `true` / `false`, so parity is unaffected either way.

kits/firestore-bigquery-export/tests/config.test.ts
```diff
@@ -34,14 +34,16 @@ vi.mock("firebase-functions/params", () => ({
       return this.toString();
     }
   },
+  // Mirrors the real StringParam: an env var wins over the declared default.
```
IzaakGough: The real `StringParam.runtimeValue()` is `process.env[this.name] || ""` and never consults the declared default (firebase-functions `lib/params/types.js`), which is the same divergence the `defineInt` mock just below calls out. So "yes/no selects default to off" passes because the mock returns `"no"`, not because it exercises the runtime unset value `""`.

Harmless for these two params, since both resolve to `false` either way. The catch is a future `defineString` whose declared default is load-bearing: this mock would green-light it, and the param would resolve to `""` in production. `process.env[_name] || ""` matches the runtime, with any default-bearing test stubbing the env var explicitly.

IzaakGough (approved):
Change itself looks right: the extension declares both params as `yes` / `no` selects and compares `=== "yes"` at runtime, and `WILDCARD_IDS` correctly stays `true` / `false`. Two low-severity notes inline, neither blocking.

kits/firestore-bigquery-export/src/config.ts
cabljac: Landed the lenient option: `yesNo` now trims and lowercases, and accepts `yes` and `true` (so the select label `Yes` and the old README advice both work); `no`, `false`, and anything else read as off. No warning: `config.ts` has no logger and the remaining unknown-value case is a typo rather than a documented spelling. Changelog reworded to say `true` still works. Tests cover `true`, `Yes`, `no`, `false`, and `enabled`.

kits/firestore-bigquery-export/tests/config.test.ts
cabljac: Confirmed in `lib/params/types.js` and landed: the mock is now `process.env[_name] || ""` with no default handling, and the one test that leaned on a default (`BIGQUERY_PROJECT_ID` from `projectID`) stubs the env var, which is what the CLI writes at deploy after resolving the default. The yes/no default test now exercises the real unset value.

kits/firestore-bigquery-export/src/config.ts
```diff
@@ -634,6 +636,11 @@ function normalizePositiveInt(value: string): number | undefined {
...
+  return value === "yes";
+}
```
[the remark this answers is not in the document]
cabljac: Landed as part of the lenient parser: `yesNo` now trims and lowercases before comparing.

kits/firestore-bigquery-export/src/config.ts
cabljac: Correction after discussion: `true` is not accepted after all. The kit has no users yet, so nobody wrote `true` from the old README, and keeping it would be dead compatibility code. What stays is trim plus lowercase, so the select label `Yes` and stray whitespace still enable the setting; `true` reads as off like any other value. Tests updated.

Issue #3068, opened 2026-09-01
fix(bigquery-firestore-export): helper crashes on real BigQuery TIME values


cabljac:
The kit's value conversion converts BigQuery temporal types via `Timestamp.fromDate(new Date(value))`, but a real `TIME` value (`BigQuery.time("10:30:00").value === "10:30:00"`) makes `new Date("10:30:00")` invalid and `Timestamp.fromDate` throws `Value for argument "seconds" is not a valid integer.` - so a scheduled query whose results include a TIME column crashes the run. Probe-verified during review of #3056 (IzaakGough); that PR pins the throwing behavior with a test so this fix flips it deliberately. Decide the correct representation (probably keep the raw string, matching what the extension stored). Ledger: #2974.

IzaakGough:
I went looking for what the extension stored for a TIME column, and it turns out it throws on the same line the kit does.

Both sides route `BigQueryTime` through the same branch:

- extension v0.2.2, `functions/src/helper.ts:365-377`: `shouldConvertToTimestamp` matches `BigQueryTime`, then `Timestamp.fromDate(new Date(value.value))`
- kit, `src/helper.ts:180-188`: same four types, same `Timestamp.fromDate(new Date(row.value))`

`BigQuery.time("10:30:00").value` is `"10:30:00"`, so `new Date(...)` is Invalid Date and both throw `Value for argument "seconds" is not a valid integer.` Probed from the kit's tree:

```
ctor: BigQueryTime value: "10:30:00"
THROW: Value for argument "seconds" is not a valid integer.
```

Nothing in the extension returns `.value` unchanged for a temporal type. `Geography` is the only type it does that for. The extension's test at `functions/__tests__/helper.test.ts:644` uses `new BigQueryTime('1970-01-01T10:30:00Z')`, which is not a value BigQuery produces, so it converts cleanly and the suite never reaches the crash.

That makes keeping the raw string a new behaviour in the kit rather than a restored one. Since parity is the goal, is that the change we want? Options as I see them:

1. Leave the kit matching the extension and keep the test from #3056 that pins the throw.
2. Fix it in both, so a migrated deploy does not shift shape.
3. Fix the kit only and document it, alongside the array vs `"0"`/`"1"` map divergence this function already has.

Happy to take whichever. Worth saying I verified this by reading both implementations and probing the value conversion locally, not against a live scheduled query, though given the two identical code paths I do not think a real run could ever have stored a TIME value.

PR #3146, opened 2026-09-08
fix(bigquery-firestore-export): write BigQuery TIME values as strings

IzaakGough:
Fixes #3068.

**What was broken.** A BigQuery `TIME` column arrives as `{ value: "10:30:00" }`, so `new Date(row.value)` is invalid and `Timestamp.fromDate` throws `Value for argument "seconds" is not a valid integer.` The throw happens inside the write loop in `writeRunResultsToFirestore`, before any `add()` promise exists, so the whole run is lost: no output rows, no run document, no `latest`.

**What changed.** `convertUnsupportedDataTypes` returns the string BigQuery gave us for a `TIME`, the same way it already does for `Geography`. Firestore has no time-of-day type, so any `Timestamp` would have to invent a date to hold the time, and the string keeps the microsecond precision a `Timestamp` cannot. `TIMESTAMP`, `DATE` and `DATETIME` are untouched.

**How it was verified.** Against a live BigQuery table with `TIME`, fractional-second `TIME`, `NULL`, arrays and structs, running `writeRunResultsToFirestore` into the Firestore emulator: before, it threw and wrote nothing; after, all three rows land with the times as strings. Unit tests cover the plain, microsecond, midnight, nested and null cases.

This diverges from the extension, which throws on the same line, so no installed instance can have stored a `TIME` value. Documented in the README and CHANGELOG.

### DIFF
--- a/kits/bigquery-firestore-export/CHANGELOG.md
+++ b/kits/bigquery-firestore-export/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Initial release of kit, see README for differences between the legacy extension and this kit
+- A BigQuery `TIME` column no longer crashes a run: its value is written to Firestore as the string BigQuery returned (`"10:30:00"`) rather than passed to `Timestamp.fromDate`, which threw and lost the whole run. This is a deliberate divergence from the legacy extension, which throws on the same line
--- a/kits/bigquery-firestore-export/README.md
+++ b/kits/bigquery-firestore-export/README.md
@@ -224,7 +224,21 @@ position, `{ "0": ..., "1": ... }`, because the conversion treated every
 non-scalar value as an object. The kit writes a real Firestore array instead.
 Anything reading those fields by numeric string key needs updating, and rows
 written before and after the change are not the same shape. Scalars, timestamps,
-dates, times, datetimes, bytes and geography values convert exactly as before.
+dates, datetimes, bytes and geography values convert exactly as before.
+
+### A TIME column is written as a string
+
+A `TIME` value arrives from BigQuery as `"10:30:00"`, and the extension passed it
+to `Timestamp.fromDate(new Date(...))`, which throws `Value for argument "seconds"
+is not a valid integer.` A run whose results included a `TIME` column therefore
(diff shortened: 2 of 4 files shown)


CorieW (approved):
Approving. The parity goal is met and there is no new defect.

**Baseline.** Extension v0.2.2 routes `BigQueryTime` through `Timestamp.fromDate(new Date(value.value))` ([`functions/src/helper.ts:375-377`](https://github.com/GoogleCloudPlatform/firebase-extensions/blob/main/bigquery-firestore-export/functions/src/helper.ts#L375-L377)) and throws on any real `TIME` value. Confirmed locally: `BigQuery.time("10:30:00.123456").value` is `"10:30:00.123456"` and `new Date(...)` is Invalid Date. The extension's test at [`__tests__/helper.test.ts:663`](https://github.com/GoogleCloudPlatform/firebase-extensions/blob/main/bigquery-firestore-export/functions/__tests__/helper.test.ts#L663) hides this by using `new BigQueryTime('1970-01-01T10:30:00Z')`, which BigQuery never produces.

**Why this is an accepted divergence, not a parity gap.** The only predecessor behavior is a crash that loses the whole run (the throw is synchronous inside `writes.push(collection.add(convert(...)))`, before `Promise.allSettled`). No migrated consumer can hold a stored `TIME` value, so nothing changes shape on migration. `TIMESTAMP`, `DATE` and `DATETIME` are untouched. All 15 tests in `helper-values.test.ts` pass on this head.

Non-blocking notes:

1. **Ledger.** #3068 asked for a decision and #2974 has no entry for `TIME`. Please add a line under the `bigquery-firestore-export` decide/Note items: kit stores the raw string (option 3), extension unchanged, this PR.
2. **Inherited bug in the extension.** The same crash is live in v0.2.2. If option 2 is ever wanted, that belongs in a separate GoogleCloudPlatform/firebase-extensions issue, and the fake-value test above should be corrected there.
3. **README nit.** "This is the one type whose value differs from the extension" sits directly under the array-vs-`{ "0": ..., "1": ... }` section, which is also a value-shape difference. "Only scalar type" would be exact.

Issue #3123, opened 2026-09-07
fix(storage-resize-images): IS_ANIMATED select label is True, the extension says Yes


cabljac:
`kits/storage-resize-images/src/config.ts:189` declares the `IS_ANIMATED` select as `select({ True: true, "No (1st frame only)": false })`. The extension's `extension.yaml:289-292` labels the same option `Yes`. The kit's param audit tests pin option values only, not labels, which is how this slipped through (`tests/config.test.ts`).

Fix: change the label to `Yes` and add an `IS_ANIMATED select` test that reads the param from `declaredParams` and asserts both labels and values (`[{ label: "Yes", value: true }, { label: "No (1st frame only)", value: false }]`).

Split out of #3064 (review thread by @IzaakGough) so that PR stays scoped to the `OFF` content-filter mapping. A grep across `kits/*/src/config.ts` finds no other `True:`/`False:` select label.

IzaakGough:
Closed by #3148

PR #3147, opened 2026-09-08
fix(storage-resize-images): label the IS_ANIMATED true option Yes

IzaakGough:
Closes #3123.

`kits/storage-resize-images/src/config.ts` declared the `IS_ANIMATED` true option as `True`, while `storage-resize-images/extension.yaml` labels it `Yes`. Only the deploy-time prompt is affected: the stored values are `true`/`false` either way, so no `.env` from an earlier deploy needs editing.

The param audit tests pinned option values but not labels, which is how it slipped through, so this also adds an `IS_ANIMATED select` test that reads the param out of `declaredParams` and asserts labels and values together.

Verified by generating the params spec the CLI reads at deploy time (`loadStack` over the built `lib/`) and diffing its options against the ones parsed out of `extension.yaml`: they now match exactly. Full kit suite passes (183 passed, 3 skipped, the skips being the pre-existing live Vertex AI content-filter tests). I also mutated the declaration seven ways (label typo, stringified value, extra option, reordered options, labels dropped, param renamed, select input removed) and confirmed the new test fails on each, rather than passing whichever way the labels went.

`CONTENT_FILTER_LEVEL`'s `"Off (No filtering)": "False"` mapping is untouched here, it belongs to #3064.


CorieW:
Superseded by https://github.com/firebase/extensions/pull/3148

PR #3148, opened 2026-09-08
fix(kits): restore extension select params for yes/no config parity

CorieW:
## Summary

The kits migration turned every extension yes/no select into `defineBoolean`. `BooleanParam` only parses the literal `true` ([`src/params/types.ts:717`](https://github.com/firebase/firebase-functions/blob/master/src/params/types.ts)), so an `.env` copied from an extension that stored `yes` / `no` silently reads as off. The `true` / `false` selects parsed fine but lost their option labels.

Rule applied per param:

- Extension stored `yes` / `no`: `defineString` + `select({ Yes: "yes", No: "no" })`, coerced in `configFromEnv()` with the extension's `=== "yes"`. Stored values unchanged; kit config types stay boolean.
- Extension stored `true` / `false` and the default is the first option: keep `defineBoolean`, restore labels with `select<boolean>({ Yes: true, No: false })`. Label-only, no runtime change.
- `MAKE_PUBLIC` (stored `true` / `false`, default `false` = second option): `defineString` with `default: "false"`, coerced with `=== "true"`. The CLI's `promptSelect` passes `default` raw but stringifies option values ([`params.ts`](https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/params.ts) `default: resolvedDefault as string` vs `value: option.value.toString()`), so a boolean default matches nothing and "Yes" was preselected, making every resized image public on Enter.

Same line as #3145, which kept `WILDCARD_IDS` boolean and converted the yes/no params. `firestore-bigquery-export` is no longer touched here.

## Affected params

| Kit | Param | Options (order) | Default | Kit param | Coercion |
| --- | --- | --- | --- | --- | --- |
| delete-user-data | `ENABLE_AUTO_DISCOVERY` | Yes=`yes`, No=`no` | `no` | `defineString` | `=== "yes"` |
| firestore-genai-chatbot | `ENABLE_DISCUSSION_OPTION_OVERRIDES` | Yes=`yes`, No=`no` | `no` | `defineString` | `=== "yes"` |
| firestore-genai-chatbot | `ENABLE_GENKIT_MONITORING` | Yes=`yes`, No=`no` | `no` | `defineString` | `=== "yes"` |
| firestore-send-email | `OAUTH_SECURE` | Yes=`true`, No=`false` | `true` | `defineBoolean` | `BooleanParam` |
| firestore-vector-search | `DO_BACKFILL` | Yes=`true`, No=`false` | none | `defineBoolean` | `BooleanParam` |
| firestore-vector-search | `UPDATE_ON_CONFIGURE` | Yes=`true`, No=`false` | none | `defineBoolean` | `BooleanParam` |
| speech-to-text | `ENABLE_AUTOMATIC_PUNCTUATION` | Enabled=`true`, Disabled=`false` | `true` | `defineBoolean` | `BooleanParam` |
| storage-resize-images | `MAKE_PUBLIC` | Yes=`true`, No=`false` | `false` | `defineString` | `=== "true"` |
| storage-resize-images | `IS_ANIMATED` | Yes=`true`, No (1st frame only)=`false` | `true` | `defineBoolean` | `BooleanParam` |
| storage-resize-images | `REGENERATE_TOKEN` | Yes=`true`, No=`false` | `true` | `defineBoolean` | `BooleanParam` |

## Changes

- Restored `IS_ANIMATED` "Yes" label (kit had "True") and speech-to-text "Enabled" / "Disabled".
- Removed README notes telling users to rewrite `yes` to `true` (delete-user-data, firestore-genai-chatbot, firestore-send-email).
- Tests in each kit's existing config test file assert param type, default, option labels, values, order, and the parsed boolean for each value and for unset. firestore-send-email uses `tests/config-runtime.test.ts` because its `config.test.ts` fakes `firebase-functions/params`.

## Verification

`npx tsc -b` clean and `npx vitest run` green in all six kits (send-email 158, vector-search 124, resize-images 172 + 6 skipped, delete-user-data 102, chatbot 58 + 3 skipped, speech-to-text 37). Prettier clean.

## Follow-up

#3145 coerces yes/no with a trimming, lowercasing `yesNo()` helper; the 3 yes/no params here use the extension's bare `=== "yes"`. Aligning them is a behavior change, so it is left for a follow-up.

## Accepted differences and inherited behavior

- Kit params with a default are not prompted as required, so the extension's `required: true` on the yes/no params is not reproduced. Platform difference, no value change.
- `OAUTH_SECURE` and `ENABLE_AUTOMATIC_PUNCTUATION` advertise default `true`, but an unset variable reads as `false` in both the extension and the kit. Reproduced for parity and pinned by tests.
- `UPDATE_ON_CONFIGURE` was declared but unused by the extension runtime; the kit uses it. Pre-existing, out of scope.


kits/firestore-vector-search/src/config.ts
```diff
@@ -162,11 +162,13 @@ const params = {
     label: "Embed existing documents?",
     description:
       "Should existing documents in the Firestore collection be embedded as well?",
+    input: select({ Yes: true, No: false }),
   }),
   updateOnConfigure: defineBoolean("UPDATE_ON_CONFIGURE", {
```
cabljac, line 185: is this meant to be boolean still?

kits/delete-user-data/src/config.ts
```diff
@@ -197,7 +197,7 @@ export function configFromEnv(): DeleteUserDataConfig {
     storageBucket:
       optional(params.storageBucket.value()) ?? process.env.STORAGE_BUCKET,
     storagePaths: optional(params.storagePaths.value()),
-    enableAutoDiscovery: params.enableAutoDiscovery.value(),
+    enableAutoDiscovery: params.enableAutoDiscovery.value() === "yes",
```
cabljac, line 200: nit: bare `=== "yes"` is exact parity with the extension, so fine to keep. Just flagging that #3145 landed a trimming/lowercasing `yesNo()` helper in bigquery-export (`src/config.ts:640`) for the same shape, so we now have two conventions. Not for this PR, but I think we should pick one and apply it everywhere in a follow-up, otherwise a hand-edited `YES` behaves differently per kit. Same applies to the two in chatbot `config.ts:377-378`.

kits/delete-user-data/tests/config.test.ts
```diff
@@ -33,6 +33,9 @@ class FakeStringParam extends FakeExpression<string> {
   }
 
   value(): string {
+    if (process.env[this.name] !== undefined) {
```
cabljac, line 36: nit: reading `process.env` first in `FakeStringParam.value()` changes what every string param in this suite resolves to, not just the new one. CI is green so nothing regressed, but worth a look that no other test here was silently relying on the default path.

cabljac (approved):
Nice one, this is the right split. I checked the values against the legacy yaml for delete-user-data, send-email and storage-resize-images and they line up: the three literal yes/no params go to defineString, the true/false ones stay defineBoolean with labels, and defineBoolean's runtime is `env === "true"` so those are label-only. I could not check chatbot, vector-search or speech-to-text the same way, their legacy sources are not in this repo, so worth a quick eyeball from you that the yes/no values there match what the extension shipped.

One blocker: it conflicts with kits now. #3122 added `kits/firestore-vector-search/tests/config.test.ts` for the instance-id test, and yours creates a different file at the same path, plus the CHANGELOG top line. Fold your declaration tests into the existing file rather than replacing it.

Parity comparison was AI-led and I only spot-checked it, so push back if anything above does not match what you saw.

kits/firestore-vector-search/src/config.ts
CorieW: Yes, intentionally. The extension stored `true` / `false` for this select, which `BooleanParam` already parses (`=== "true"`), so the runtime value is unchanged. Only the `yes` / `no` selects (3 params in other kits) moved to `defineString`. The `select<boolean>` here is label-only, restoring the extension's `Yes` / `No` prompt labels. Reasoning is now in the PR description under Summary and Decision.

Issue #3020, opened 2026-09-01
fix(firestore-translate-text): always emit start and completion events


cabljac:
The handler early-returns on `!event.data` without emitting; the extension always emitted both start and completion events.

Parity ledger: #2974, firestore-translate-text.

IzaakGough:
Closed by #3149

PR #3149, opened 2026-09-08
fix(firestore-translate-text): emit start and completion events on every write

cabljac:
The kit's `handleDocumentWrite` returned before recording anything when a gen2 write event arrived without change data, so that invocation published neither `onStart` nor `onCompletion`. The extension's `fstranslate` is a gen1 `onWrite` with no such path: every call publishes `onStart` first and `onCompletion` last, and each early return records completion on its way out.

The guard now runs after `logs.start` and `recordStartEvent`, and records `onCompletion` before returning, so the no-data path matches the other no-op branches (delete, unchanged input, missing input). The `onStart` payload carries `{ data: undefined, params }` there. The payload shape itself is untouched; restoring the extension's `{ change, context }` shape is PR #3098.

I compared every branch of the two handlers. Field-name checks, create, delete, the three update branches, the per-document translation-path return, and the error paths already published the same events in the same order (`onSuccess` only from `updateTranslations`, one `onError` per layer before `onCompletion`). The one remaining difference is that the kit awaits the handler-level `onError` where the extension does not; that only changes what happens when the publish itself rejects, so I left it alone.

A new `lifecycle events` block pins the exact event sequence for each branch, and the existing no-data test now asserts the pair is published. 112 tests and `tsc --noEmit` pass. No live deploy or Eventarc subscription was exercised.

Fixes #3020


IzaakGough (approved):
lgtm!

PR #3153, opened 2026-09-08
fix(firestore-bigquery-export): normalize DATABASE_REGION before the region lookup

cabljac:
The multi-region lookup in `region.ts` was case-sensitive, so a hand-edited `.env` carrying `NAM5` passed through untouched and reproduced the deploy failure the mapping exists to prevent. The value is now trimmed and lowercased before the lookup, regional values come out lowercased too, and a whitespace-only value counts as unset.

This is the same change #3102 made to firestore-send-email, whose `region.ts` was copied from this kit. Both kits now behave identically.

Tests cover `NAM5`, `Eur3`, ` nam7 `, a mixed-case regional value, and a whitespace-only value. Unit suite passes locally (105 tests). No live deploy run.


kits/firestore-bigquery-export/src/region.ts
```diff
@@ -24,14 +24,17 @@ const MULTI_REGION_TO_FUNCTION_REGION: Record<string, string> = {
...
+ * should deploy to. The lookup is case-insensitive and ignores surrounding
+ * whitespace, as the CLI's own region handling is. Regional locations pass
+ * through lowercased; an unset or blank location returns `undefined`, meaning
+ * the functions declare no region.
  */
 export function firestoreLocationToFunctionRegion(
   location: string | undefined
 ): string | undefined {
-  if (!location) {
+  const normalized = location?.trim().toLowerCase();
```
IzaakGough, line 35: Could the README pick up the same note? `README.md` still tells users to set `FUNCTION_REGION` to their `DATABASE_REGION` value for a manual enqueue, and the `DATABASE_REGION places the functions` section says regional locations are used as-is, so neither mentions the lowercasing. Someone with `DATABASE_REGION=Europe-West2` who follows it literally sets `FUNCTION_REGION=Europe-West2`, but the function deployed to `europe-west2`, so `gcloud functions describe --region Europe-West2` and `taskQueue("locations/Europe-West2/functions/kit-default-initBigQuerySync")` both point at a region that does not exist.

`firestore-send-email/README.md` already carries "The value is matched case-insensitively"; the same sentence here would finish the parity. Not blocking.

IzaakGough (approved):
Change looks good. Verified against both callers of `firestoreLocationToFunctionRegion` (`src/index.ts` and `src/tasks.ts`), and `region.ts` plus `tests/region.test.ts` are now identical to the `firestore-send-email` versions. Tests pass locally.

One docs follow-up left inline.

IzaakGough:
Closing as looks like this was already done in https://github.com/firebase/extensions/commit/4c5b2ca7c6bc0da437370a06b056f46bf35fd9bb.

Issue #3154, opened 2026-09-08
chore(kits): align firestore-incremental-capture, firestore-send-email and firestore-vector-search on firebase-functions ^7.3.3-rc.0


cabljac:
On the kits branch, ten of the thirteen kits under kits/ depend on firebase-functions ^7.3.3-rc.0. Three do not: firestore-incremental-capture pins 7.3.2, and firestore-send-email and firestore-vector-search ask for ^7.3.2. A caret range never matches a prerelease, so those three resolve to stable 7.3.2 while the rest run the 7.3.3-rc.0 release candidate.

Kits must all run on the same firebase-functions build so behaviour is uniform across them: requiresAPI declarations, select on boolean params and kit instance id handling all differ between 7.3.2 and the rc.

Fix: bump the dependency in the three package.json files to ^7.3.3-rc.0, refresh each kit's npm-shrinkwrap.json so it locks the same rc tarball the other kits lock, and note the change in each CHANGELOG.

Parent tracker: #2974.

PR #3155, opened 2026-09-08
chore(kits): align the last three kits on firebase-functions ^7.3.3-rc.0

cabljac:
Ten of the thirteen kits depend on firebase-functions ^7.3.3-rc.0, but firestore-incremental-capture pinned 7.3.2 and firestore-send-email and firestore-vector-search asked for ^7.3.2. A caret never matches a prerelease, so those three resolved to stable 7.3.2 and ran a different firebase-functions build from the rest, which matters for requiresAPI, select on boolean params and kit instance id handling.

This bumps the dependency in the three package.json files to ^7.3.3-rc.0, regenerates each kit's npm-shrinkwrap.json (the only change in each is the firebase-functions entry, locking the same tarball and integrity hash the other ten kits already lock), and adds a CHANGELOG line to each kit. No source changes.

Verified locally per kit with npm install, tsc -b and vitest run: firestore-incremental-capture 75 tests, firestore-send-email 155, firestore-vector-search 120, all passing with a clean typecheck. Not verified: a deploy of any of the three kits against the rc, and the emulator suites (none of the three defines test:emulator).

Fixes #3154


IzaakGough (approved):
lgtm

Issue #3140, opened 2026-09-07
decision(speech-to-text): transcoded .wav lands in a different Storage location


cabljac:
The extension derived the transcoded `.wav` Storage destination from the local `/tmp` path. The kit derives it from the object name and normalises the prefix separator. Migrated users find their transcoded audio in a different location.

Extension:

```ts
storagePath: config.outputStoragePath
  ? `${config.outputStoragePath}${transcodeResult.outputPath}` // "transcriptions" + "/tmp/a.mp3.wav"
  : transcodeResult.outputPath.slice(1);                       // "tmp/a.mp3.wav"
```

Kit:

```ts
const transcodedObjectName = `${filePath}.wav`;
storagePath: config.outputStoragePath
  ? `${config.outputStoragePath.replace(/\/$/, "")}/${transcodedObjectName}`
  : transcodedObjectName;
```

For an input object named `a.mp3`:

| `OUTPUT_STORAGE_PATH` | Extension writes | Kit writes |
| --- | --- | --- |
| unset | `tmp/a.mp3.wav` | `a.mp3.wav` |
| `transcriptions` | `transcriptions/tmp/a.mp3.wav` | `transcriptions/a.mp3.wav` |
| `transcriptions/` | `transcriptions//tmp/a.mp3.wav` | `transcriptions/a.mp3.wav` |

The `.txt` transcript path had a related remnant (`replace("tmp/", "")`), fixed in https://github.com/firebase/extensions/pull/3073 (#3026). The `.wav` location itself is still open.

Decide one of:

- Parity: reproduce the extension's `tmp/`-prefixed paths, double slash included, so existing consumers keep finding files where they were.
- Keep the kit's path: the extension's `tmp/` segment is an artefact of the local download path, not a design. Document the change and the old-vs-new matrix in the kit README and the Notes entries (#3035).

Parity ledger: #2974, speech-to-text §1.

DECISION: go with Parity

IzaakGough:
One update on the Parity option, from testing #3157 on a live deploy: the third row of the table cannot be honoured usefully.

Cloud Storage accepts `transcriptions//tmp/a.mp3.wav` as an object name, so the transcode uploads fine, but Speech-to-Text rejects the same path as an audio URI:

```
3 INVALID_ARGUMENT: Audio URI `gs://<bucket>/transcriptions//tmp/clip.wav.wav` is an invalid GCS path.
```

[...]

PR #3157, opened 2026-09-08
fix(speech-to-text): write the transcoded .wav under tmp/ like the extension

cabljac:
The extension named the transcoded WAV after the local temp file it had just written, so the object landed at `tmp/<original path>.wav` under `OUTPUT_STORAGE_PATH`. The kit named it after the input object instead, so migrated users could not find their transcoded audio.

This restores the extension's paths for both outputs. Because the Speech API output was named after the uploaded object, putting `tmp/` back also moved the `.txt`, so `transcribeAndUpload` now takes an explicit transcript object name and the handler strips the segment back off, exactly as the extension did. `path.posix.join` supplies the normalisation `path.join(os.tmpdir(), filePath)` gave for free.

One deviation, and it needs a decision: a trailing slash on `OUTPUT_STORAGE_PATH` is stripped rather than doubled. The extension's `transcriptions//tmp/a.mp3.wav` is a valid object name but not a valid Speech API audio URI, so that config uploaded the audio and never wrote a transcript, leaving the Firestore document on "Transcoding audio file.". #3140 chose parity "double slash included"; that would reproduce a total failure, so this reverses that half.

Verified on a live deploy (three instances, one bucket): all six paths are as documented above, with real transcripts. The double-slash failure and its fix were both observed there, not inferred. `vitest run` (44 tests), `tsc --noEmit` and prettier pass.

Fixes #3140


kits/speech-to-text/src/handlers.ts
```diff
@@ -148,18 +148,11 @@ export async function handleObjectFinalized(
       message: "Transcoding audio file.",
     });
 
-    /**
-     * Bucket-relative object name for the transcoded file, derived from the
-     * input object's path/name (not the local `/tmp` path).
-     */
-    const transcodedObjectName = `${filePath}.wav`;
+    const transcodedObjectName = `tmp/${filePath}.wav`;
```
IzaakGough: Putting `tmp/` back on the transcoded object also moves the transcript.

`transcribeAndUpload` receives the uploaded transcoded file and builds the Speech API output from its name (`src/transcribe.ts`, `gs://${bucket.name}/${name}_transcription.txt`). The legacy extension stripped the prefix at that point (`name.replace("tmp/", "")` in `legacy/functions/src/transcribe-audio.ts`), so for `a.mp3` with `OUTPUT_STORAGE_PATH` unset:

| | WAV | transcript |
|---|---|---|
| extension | `tmp/a.mp3.wav` | `a.mp3.wav_transcription.txt` |
| kit before this PR | `a.mp3.wav` | `a.mp3.wav_transcription.txt` |
| kit after this PR | `tmp/a.mp3.wav` | `tmp/a.mp3.wav_transcription.txt` |

The transcript matched the extension before this change and does not after. That falls out of the #3026 decision to derive the transcript from the transcoded object name, so it is not a plumbing slip, but the net effect is that the PR gains parity on the intermediate audio and loses it on the primary output. That seems worth making an explicit call rather than letting it ride as a side effect.

If the transcript should stay where it is, deriving the Speech output from `filePath`, or stripping only the leading `tmp/` segment this line adds rather than any `tmp/` inside the user's own object name, would give both.

IzaakGough: Smaller point on the same line. The extension built this path from `path.join(os.tmpdir(), filePath)` and then sliced off the leading `/`, so it got normalisation for free. Template interpolation does not, and GCS object names can contain `//` or `..` segments. An object named `audio//clip.mp3` landed at `tmp/audio/clip.mp3.wav` under the extension and lands at `tmp/audio//clip.mp3.wav` here. Rare, but the README says the path is produced exactly as the extension did, and on this axis it is not.

Also worth a one-line comment noting the `tmp/` segment is deliberate parity with the extension. It has been added and removed before (#3026), and without a note naming the constraint it reads like a leaked local temp path.

kits/speech-to-text/README.md
```diff
@@ -134,27 +134,27 @@ ffmpeg transcode to LINEAR16, the same long-running recognition request, the sam
...
+The transcoded WAV is written to `tmp/<original path>.wav`, or
+`<OUTPUT_STORAGE_PATH>/tmp/<original path>.wav` when `OUTPUT_STORAGE_PATH` is
+set. The `tmp/` segment is an artefact of the extension naming the copy after
+its local temporary file; the kit keeps it so lifecycle rules, cleanup jobs and
+client code written against the extension keep finding the file. The prefix is
+joined with a single `/` and is not normalised, so a trailing slash on
+`OUTPUT_STORAGE_PATH` produces a double slash (`transcriptions//tmp/a.mp3.wav`),
+exactly as the extension did.
+
+The transcript is written next to the WAV as
```
IzaakGough: This paragraph no longer says where the transcript lands relative to the extension. The version being replaced called out that the transcript stayed at `<original path>.wav_transcription.txt`, but with the WAV back under `tmp/` the transcript moves there too, so someone migrating and hunting for their `.txt` will not find it at the path the extension used. Moot if the code side changes.

kits/speech-to-text/CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- fix: the transcoded `.wav` is written to `tmp/<original path>.wav` again (under `OUTPUT_STORAGE_PATH` when set, joined with a single `/` and no trailing-slash normalisation), matching the legacy extension so existing consumers keep finding it ([#3140](https://github.com/firebase/extensions/issues/3140))
```
IzaakGough: Same point as on the README: the entry covers the `.wav`, but the `.txt` picks up the `tmp/` prefix as well. This is the line a migrating user reads before upgrading, so it is worth naming both outputs.

IzaakGough:
Pushed fixes for all four points in 5222924, and merged `kits` in since the branch was a way behind. Resolving the threads.

The transcript was the substantive one: `transcribeAndUpload` now takes an explicit transcript object name and the handler strips the `tmp/` segment it added, so the `.wav` and the `.txt` both land where the extension put them. The object path is built with `path.posix.join` for the normalisation, and there is a comment naming the `tmp/` segment as deliberate parity.

The README section is now a table covering both outputs across the unset, `transcriptions` and `transcriptions/` cases. The CHANGELOG entry names both outputs, and I folded the old #3026 line into it, since that line said the `.txt` keeps `tmp/` and this change reverses it. Happy to split them back out if you would rather keep the history separate.

Two legacy quirks are left alone and called out in the description: an `OUTPUT_STORAGE_PATH` containing its own `tmp/`, and an object name climbing above `tmp/` more than once.

IzaakGough:
Tested this on a live deploy, and it turned up one thing that changes the decision in #3140. Deployed three instances against one bucket so a single upload exercised all three prefix cases, on `us-central1` with a real 6 second speech clip.

The `tmp/` parity works exactly as intended for an unset prefix and for `transcriptions`, transcripts included. The trailing-slash case does not:

```
3 INVALID_ARGUMENT: Audio URI `gs://<bucket>/transcriptions//tmp/clip.wav.wav` is an invalid GCS path.
```

Cloud Storage accepts the double-slash object name, so the transcode uploads fine and the failure only lands at the recognition call. The result is an orphaned `.wav`, no transcript, and a Firestore document stuck on "Transcoding audio file.". The extension passed the same URI to the same API, so it failed the same way, which means no trailing-slash install has ever had a working transcript to stay compatible with. The kit's trailing-slash strip, which this PR had removed, was the only reason that config worked.

So 103cf4ef puts the strip back and keeps everything else. After that, all three configs succeed. Worth noting the `path.posix.join` change is load-bearing for the same reason: an object named `audio//clip.mp3` would hit the identical error, and now normalises instead.

Happy to be talked out of it if the double slash matters to someone, but as it stands parity there means reproducing a hard failure.

PR #3158, opened 2026-09-08
refactor(firestore-send-email): construct MailService from the named export

IzaakGough:
## What was broken

`src/nodemailer-sendgrid/index.ts` imported `@sendgrid/mail` as a default import. That package assigns a `MailService` instance straight to `module.exports` with no `__esModule` marker, so the import only resolved through tsc's `esModuleInterop` helper. Turn that option off and the emitted `require` has no `.default`, so the transport throws on construction. With `allowSyntheticDefaultImports` on the build still looks clean, so the break is silent until mail sends.

## What changed

The package also exports the class (`export {MailService}`). A named import compiles to a plain property access, so the emitted output contains no `__importDefault` at all.

Each transport now constructs its own `MailService` instead of calling `setApiKey` on the shared singleton. Not a bug today, the kit only builds transports with one key, but two keys previously clobbered each other.

`tests/build-interop.test.ts` stays, now guarding a revert to a default import. It reads `lib/`, so a `pretest` hook builds first: the release workflow's test job runs `npm test` without building.

## How it was verified

Clean `npm ci`, build, full suite green. Against the real `@sendgrid/mail` and the built `lib/`: the key reaches the client, two transports keep separate keys, the singleton is never mutated. Same rebuilt with `esModuleInterop: false` and `allowSyntheticDefaultImports: true`, the config that silently broke the old code.


CorieW (approved):
Approving. Verified against `@sendgrid/mail@8.1.6`: the default export really is a shared singleton, separate `MailService` instances get separate clients, and the named import emits a plain `require` with no `__importDefault` wrapper, so the interop bug is structurally gone. No regression from dropping the ambient key either, since there's no `SENDGRID_API_KEY` fallback and `helpers.ts` always passes one.

Two non-blocking follow-ups:

1. Consider keeping `tests/build-interop.test.ts`. With it gone, nothing exercises the built output or the real module shape, and it passes unchanged here.
2. The mock returns the same object for every construction, so a revert to `sgMail.setApiKey(...)` would still pass. Worth asserting `sgMail.MailService` was constructed.

IzaakGough:
@CorieW I agree that having the build-interop test would be good. The release pipeline at the moment doesn't build before running the tests so if we add this back it will still fail. Do you think we should add a build step in the yaml before running the tests or maybe a script in the package.json only for this kit?

CorieW:
Oops, sorry Izaak. Didn't mean to send that comment yet. Had to review it first, but it jumped the gun.

CorieW:
Script for now, to avoid touching the release workflow. A `"pretest": "npm run build"` in the kit's package.json covers it, since the release job's bare `npm test` will then build first. Worth a comment on the line, as no other kit has one. The PR workflow already builds before testing, so this is only about the release path.

PR #3160, opened 2026-09-09
chore(kits): align firebase-functions on 7.3.3-rc.1

CorieW:
## Changes

- 11 kits: `firebase-functions` `^7.3.3-rc.0` -> `^7.3.3-rc.1` in `package.json` and `npm-shrinkwrap.json`.
- `kits/firestore-bigquery-export` and `kits/storage-resize-images` unchanged; already on `^7.3.3-rc.1`.

## Why rc.1 and not 7.3.2

`7.3.3-rc.1` is the newest published `firebase-functions` (npm `next`). The npm `latest` tag is `7.3.2`, older than the rc line every kit already uses, so moving there would be a downgrade and would revert the rc pins `release-kit.yaml` applies via `use_firebase_functions_rc`.

## Risk

`rc.0` -> `rc.1` changes no transitive requirements. Comparing the two shrinkwrap entries already on this branch, `dependencies`, `peerDependencies`, `peerDependenciesMeta`, `engines`, `bin` and `license` are identical. Every lockfile diff is therefore limited to `version`, `resolved` and `integrity`, and the integrity matches the value already committed for the two kits on rc.1.

## Verification

`npm ci && npm run build && npm test` pass locally for all 11 changed kits. Emulator suites left to CI.

The root `pnpm-lock.yaml` is not touched here; it is stale against `kits` independently of this change and is refreshed separately.

### DIFF
--- a/kits/firestore-incremental-capture/package.json
+++ b/kits/firestore-incremental-capture/package.json
@@ -35,7 +35,7 @@
     "@google-cloud/bigquery": "^7.6.0",
     "@google-cloud/dataflow": "^3.2.0",
     "firebase-admin": "^14.2.0",
-    "firebase-functions": "^7.3.3-rc.0"
+    "firebase-functions": "^7.3.3-rc.1"
   },
   "devDependencies": {
     "@types/node": "^26.2.0",
--- a/kits/firestore-vector-search/package.json
+++ b/kits/firestore-vector-search/package.json
@@ -35,7 +35,7 @@
     "@google-cloud/firestore": "^7.6.0",
     "@google-cloud/vertexai": "^0.1.3",
     "firebase-admin": "^14.1.0",
-    "firebase-functions": "^7.3.3-rc.0",
+    "firebase-functions": "^7.3.3-rc.1",
     "genkit": "^1.27.0",
     "openai": "^4.20.1",
     "zod": "^3.22.4"
--- a/kits/bigquery-firestore-export/npm-shrinkwrap.json
+++ b/kits/bigquery-firestore-export/npm-shrinkwrap.json
@@ -14,7 +14,7 @@
         "@google-cloud/bigquery-data-transfer": "^5.0.1",
         "@google-cloud/pubsub": "^4.11.0",
         "firebase-admin": "^14.1.0",
-        "firebase-functions": "^7.3.3-rc.0"
+        "firebase-functions": "^7.3.3-rc.1"
       },
       "devDependencies": {
         "typescript": "^5.9.3",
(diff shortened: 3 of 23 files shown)


kits/firestore-incremental-capture/package.json
```diff
@@ -35,7 +35,7 @@
     "@google-cloud/bigquery": "^7.6.0",
     "@google-cloud/dataflow": "^3.2.0",
     "firebase-admin": "^14.2.0",
-    "firebase-functions": "^7.3.3-rc.0"
+    "firebase-functions": "^7.3.3-rc.1"
```
IzaakGough, line 38: The pending changelog entry for this kit still says rc.0:

```
kits/firestore-incremental-capture/CHANGELOG.md:1
- chore: run on firebase-functions ^7.3.3-rc.0, the same release candidate as the other kits
```

`.github/workflows/release-kit.yaml` reads the changelog whole (`NOTES=$(cat "$CHANGELOG_FILE")`) and publishes it as the release notes, so the next release would name a version this kit no longer uses. Worth bumping the line to rc.1 here.

kits/firestore-send-email/package.json
```diff
@@ -33,7 +33,7 @@
   "dependencies": {
     "@sendgrid/mail": "^8.1.4",
     "firebase-admin": "^14.1.0",
-    "firebase-functions": "^7.3.3-rc.0",
+    "firebase-functions": "^7.3.3-rc.1",
```
IzaakGough, line 38: Same as the other two kits: `kits/firestore-send-email/CHANGELOG.md:2` still reads `^7.3.3-rc.0`, and that text is published verbatim as the release notes.

kits/firestore-vector-search/package.json
```diff
@@ -35,7 +35,7 @@
     "@google-cloud/firestore": "^7.6.0",
     "@google-cloud/vertexai": "^0.1.3",
     "firebase-admin": "^14.1.0",
-    "firebase-functions": "^7.3.3-rc.0",
+    "firebase-functions": "^7.3.3-rc.1",
```
IzaakGough, line 38: Same here: `kits/firestore-vector-search/CHANGELOG.md:1` still reads `^7.3.3-rc.0`, and that text is published verbatim as the release notes.

IzaakGough (approved):
Checked the rc.1 integrity hashes in each shrinkwrap against the npm registry, confirmed every lockfile still matches its package.json with a single firebase-functions entry, and ran a real `npm ci` plus build and tests for `kits/rtdb-limit-child-nodes` from this branch. Diffing the rc.0 and rc.1 tarballs, the only change is an additional global param registry that merges into the manifest, so no kit behaviour changes.

One small thing across three kits: their pending changelog entries still name rc.0. `release-kit.yaml` cats the changelog and publishes it verbatim, so the next release of those kits would ship notes naming a version they no longer use. Comments inline. Approving since it is just the note text.

CorieW:
Good catch, fixed in 7dbeaf9. All three pending changelog entries now name `^7.3.3-rc.1`, and `grep -rn '7.3.3-rc.0' kits/` is clean.

Left the other 10 kits alone: they have no pending entry naming a firebase-functions version, and given the rc.0 to rc.1 delta is just the extra global param registry you found, there is no user-visible change to write up for them.

Also rebased #3161 on top so the stack stays linear.

Issue #3105, opened 2026-09-03
decision(firestore-vector-search): OpenAI vector index is 512 dimensions while the vectors are 1536


CorieW:
Follow-up from #3096 ([review](https://github.com/firebase/extensions/pull/3096#pullrequestreview-5094021859)). #3029 decided the embedding *model*; it did not decide the index dimension, and once it closes nothing tracks this. Today it lives only in the kit's README.

## The problem

With `EMBEDDING_PROVIDER: openai`, the vector index the kit creates does not cover the vectors it writes, so `findNearest` cannot use it.

| | Value | Source |
|---|---|---|
| Index dimension | **512** | `dimensionFor()` in `src/export-config.ts:90`, passed to `createIndex` |
| Vector written | **1536** | `text-embedding-ada-002` native size, `src/embeddings/client/text/open_ai.ts` |

This is inherited from the extension, deliberately. The extension has two dimension tables that disagree: `getDimension()` returns 512 for openai while `OpenAIEmbedClient` is constructed with `dimension: 1536` and ada-002 returns 1536. #3096 restored the extension's model, so the kit now reproduces the mismatch exactly (parity ledger #2974, `firestore-vector-search` §1 and §2).

Verified live on a deployed instance. The vectors store fine — Firestore's limit is 2048 — but querying them fails:

```
9 FAILED_PRECONDITION: Missing vector index configuration. Please create the required
index with the following gcloud command: gcloud firestore indexes composite create
--project=<project> --collection-group=<collection> --query-scope=COLLECTION
--field-config=vector-config='{"dimension":"1536","flat": "{}"}',field-path=embedding
```

So `EMBEDDING_PROVIDER: openai` embeds correctly and is unqueryable through `queryCallable`, `queryOnWrite`, or any direct `findNearest`, unless the user creates a 1536-dimension index by hand.

## Options

1. **`dimensionFor("openai")` returns 1536.** The index then matches what ada-002 actually produces and the provider works end to end. Diverges from the extension's 512, but that 512 never matched its own client, so this repairs a self-contradiction rather than changing a working behaviour.
2. **Pin `dimensions: 512` on the OpenAI request.** ada-002 does not support the `dimensions` parameter, so this means moving to `text-embedding-3-small` — which is what the kit did before #3096, and what #3029 decided against.
3. **Leave it and keep documenting it.** Full parity, including the bug. The default provider is unaffected; anyone on openai gets a README note and a manual index.

[...]

Issue #3029, opened 2026-09-01
decision(firestore-vector-search): embedding model and dimension defaults


cabljac:
OpenAI moved ada-002 to 3-small@512, incompatible with existing indexes (§2); gemini/vertex dropped the hardcoded `outputDimensionality: 768` for `config.dimension` with truncation (§5). Decide: configurable model with documented default change, or restore parity.

Ledger: #2974.

PR #3163, opened 2026-09-09
fix(firestore-vector-search): declare the OpenAI vector index at 1536 dimensions

CorieW:
Fixes #3105, option 1 plus the migration note it asks for.

With `EMBEDDING_PROVIDER: openai` the index the kit creates is 512 dimensions but the vectors it writes are 1536 (`text-embedding-ada-002`), so every query fails with `9 FAILED_PRECONDITION: Missing vector index configuration`. Inherited from the extension, whose `getDimension()` returns 512 while its `OpenAiEmbedClient` is built with `dimension: 1536`. The 512 never matched its own client, so there is no working behaviour to preserve.

## Changes

- `dimensionFor("openai")` returns 1536. Only the index declaration changes: the OpenAI client never reads `config.dimension`, and gemini, vertex, multimodal and custom are untouched.
- README and CHANGELOG: migration note with the `gcloud` commands to delete the old 512 index, since `createIndex` matches on collection name and field path only and would keep it.
- Tests: `dimension` per provider in `tests/export-config.test.ts`, and a `tests/setup.test.ts` case pinning the dimension-blind existence check as inherited behaviour.

## Notes

- Deliberate parity divergence, so it needs the same decision treatment #3029 got.
- Left inherited: making the existence check compare dimensions would also silently rebuild drifted indexes on every install. Worth a separate issue.
- `npm run build`, `vitest run` (132 tests) and `prettier --list-different` all clean. Cold-run `import()` timeouts reproduce on the baseline.
- The live `FAILED_PRECONDITION` failure was confirmed by the reporter, not re-verified on a deployed instance here.


CorieW:
Closing this: it is out of parity with the extension, and the kit is already in strict parity here.

Both sides are identical on the OpenAI path. The index is declared at 512 and the vectors are ada-002 at 1536, in the extension and in the kit:

| | extension | kit (`origin/kits`) |
|---|---|---|
| index dimension | 512 (`functions/src/config.ts:50`) | 512 (`dimensionFor()`, `src/export-config.ts`) |
| model | `text-embedding-ada-002` | `text-embedding-ada-002` |
| batch size | 16 | 16 |
| vector written | 1536 | 1536 |

So the 512 is not a kit defect, it is the extension's behaviour reproduced exactly, which is what #3096 set out to do. Changing `dimensionFor("openai")` to 1536 would make the kit create a different Firestore index resource than the extension does, which is a divergence this PR should not introduce on its own.

The underlying bug is real but inherited, and it belongs in both codebases rather than in the parity ledger. Tracking it separately.

Issue #3156, opened 2026-09-08
storage-resize-images: FUNCTION_MEMORY prompt preselects 512 MB instead of the extension's 1 GB default


CorieW:
The `storage-resize-images` kit's `FUNCTION_MEMORY` prompt preselects **512 MB**; the extension preselected **1 GB**.

- Kit: `defineInt("FUNCTION_MEMORY", { default: 1024, input: select({ "512 MB": 512, "1 GB": 1024, ... }) })` ([`kits/storage-resize-images/src/config.ts:201-214`](https://github.com/firebase/extensions/blob/kits/kits/storage-resize-images/src/config.ts#L201-L214))
- Extension: `type: select`, first option `512`, `default: 1024` ([`storage-resize-images/extension.yaml`](https://github.com/firebase/extensions/blob/kits/storage-resize-images/extension.yaml))
- Cause: firebase-tools `promptSelect` passes a non-string `default` to inquirer unchanged but stringifies every option value, so `1024` matches no choice and the first option is highlighted (firebase/firebase-tools#11053).

A user who accepts the prompt gets a 512 MB function where the extension gave them 1 GB; the README recommends at least 2 GB for animated GIFs.

Same class as `MAKE_PUBLIC`, which #3148 fixed by declaring the param as a string. `FUNCTION_MEMORY` is the only other non-string select across `kits/*/src/config.ts` whose default is not the first option.

**Options:** declare it as `defineString` with `"512"` / `"1024"` / ... values and `parseInt` at read (stored `.env` values unchanged), or wait for the firebase-tools fix. Either way, add a test that pins the default and option order.

Related: #3148, #2974, firebase/firebase-tools#11053

CorieW:
Fixed by https://github.com/firebase/extensions/pull/3164, merged to `kits`.

PR #3164, opened 2026-09-09
fix(storage-resize-images): preselect the extension's 1 GB memory default

CorieW:
Fixes #3156.

The `FUNCTION_MEMORY` prompt highlighted `512 MB` where the extension preselected `1 GB`, so accepting the prompt deployed the function with half the memory. The CLI's `promptSelect` compares its `default` against `option.value.toString()`, so the int `1024` matches nothing and the first option wins (firebase/firebase-tools#11053).

`MAKE_PUBLIC` was fixed in #3148 by declaring the param as a string. That does not work here: `FUNCTION_MEMORY` also feeds `availableMemoryMb` as `{{ params.FUNCTION_MEMORY }}`, and the CLI resolves that as a number only for an int param, so a string param fails every deploy with `Illegal type coercion of param FUNCTION_MEMORY to type number` (checked against firebase-tools 15.26.0). The param therefore stays an int and the default is listed first.

## Changes

- `src/config.ts`: list `1 GB` first in the `FUNCTION_MEMORY` select.
- `tests/config.test.ts`: pin the type, default, labels, values and first option; fails on the previous ordering.
- `CHANGELOG.md`: note the fix.

Accepted difference: option order only, the extension listed `512 MB` first. Labels, stored `.env` values and the default are unchanged, so no migration. Also correct if the CLI later stringifies the default, since `1024` then matches `"1024"`.

Verified: `npm run build` and `npx vitest run` in `kits/storage-resize-images`, 189 passed / 6 skipped (live Vertex tests).


kits/storage-resize-images/tests/config.test.ts
```diff
@@ -168,6 +168,39 @@ describe("configFromEnv", () => {
...
+      { label: "8 GB", value: 8192 },
+    ]);
+
+    // The comparison the CLI actually makes: `default` against
+    // `option.value.toString()`. An int default matches nothing, so the first
+    // option is highlighted and has to be the extension's default.
+    const preselected = options.filter(
+      (option) => String(option.value) === declared.default
+    );
+    expect(preselected).toEqual([]);
```
IzaakGough: This block always passes: `String(option.value)` is a string and `declared.default` is the number `1024`, so the strict `===` is false for every option and `preselected` is `[]` no matter what the option order or default is. The bug being fixed here would not make it fail either.

The comment above it describes it as the comparison the CLI makes, so a later reader could reasonably trust it as the guard and weaken or drop `expect(options[0]).toEqual({ label: "1 GB", value: 1024 })`, which is the assertion actually pinning the fix.

Suggest either dropping the `preselected` block, or making it non-vacuous, for example asserting `typeof declared.default !== "string"` next to the `options[0]` check so the reason the first option matters is stated without a comparison that can never match.

IzaakGough (approved):
Fix looks right. I checked the mechanism against firebase-tools 13.35.1: `promptSelect` passes the declared `default` straight through, and inquirer 8's list prompt only treats a numeric default as an index when it is less than the number of choices, so `1024` with 5 options falls back to index 0. `select()` in firebase-functions maps via `Object.entries` with non-integer keys, so insertion order is preserved and listing `1 GB` first is what actually fixes it. Keeping the param as `defineInt` is needed too, since `src/index.ts` passes `{{ params.FUNCTION_MEMORY }}` as the function's `memory` option.

One note on the test, inline.

kits/storage-resize-images/tests/config.test.ts
CorieW: Great feedback, you're right. Fixed in 4662b2508: dropped the filter, now asserts `typeof declared.default` is not a string next to the `options[0]` check. Re-verified it fails when the options go back to `512 MB` first.

ajperel:
FYI - @wandamora is fixing some stuff in the CLI for kits with advanced parameters. Not sure if https://github.com/firebase/firebase-tools/pull/11056 will impact this PR.

wandamora:
> FYI - @wandamora is fixing some stuff in the CLI for kits with advanced parameters. Not sure if [firebase/firebase-tools#11056](https://github.com/firebase/firebase-tools/pull/11056) will impact this PR.

My PR interacts with `EXT_MIGRATED_SYSTEM_MEMORY`, so I believe both PRs have no impact on each other.

PR #3173, opened 2026-09-10
feat(kits): place event-triggered functions by resource location

IzaakGough:
2nd gen Eventarc triggers are tied to the region of the resource they listen to, and the CLI cannot always place the function next to it. Two things go wrong.

For `storage-resize-images`, `speech-to-text` and `rtdb-limit-child-nodes`, the trigger's bucket or instance is a param expression, so the CLI's region lookup sees the literal `{{ params.X }}` and falls back to `us-central1` (firebase/firebase-tools#11020). Any bucket or database outside that region fails the deploy.

For `firestore-counter` and `firestore-vector-search`, the CLI places the Firestore-triggered functions correctly but gets no signal for the scheduled, task and callable ones, so an instance splits across regions: a live deploy with no region set put `embedOnWrite` and `queryOnWrite` in `us-east1` next to the database and the other six functions in `us-central1`.

## Changes

Each kit gains a location param and deploys its functions to the Cloud Run region derived from it, following #3101 and #3102. Locations that are not themselves Cloud Run regions map to one inside them, using the same targets firebase-tools picks.

| Kits | Param | Mapping |
|---|---|---|
| `firestore-counter`, `firestore-vector-search` | `DATABASE_REGION` | `nam5`/`nam7` to `us-central1`, `eur3` to `europe-west1` |
| `storage-resize-images`, `speech-to-text` | `BUCKET_REGION` | `us` to `us-east1`, `eu` to `europe-west1`, `asia` to `asia-east1` |
| `rtdb-limit-child-nodes` | `DATABASE_REGION` | none needed |
| `firestore-incremental-capture` | `LOCATION`, already present | docs only |

Every function in an instance is pinned, not just the triggered one, because `firestore-vector-search` and `firestore-counter` resolve task queues and schedules from the enqueuing function's own region.

`firestore-translate-text` and `firestore-genai-chatbot` get no param. Each has a single Firestore trigger on the default database with no `database` option, so the CLI already places it correctly and the param only added a required prompt. Their READMEs are corrected instead, since they claimed a no-region function lands in `us-central1`.

`firestore-bigquery-export` gains no param here, but its region helper now trims and lowercases the location like the other kits do, so a hand-edited `NAM5` resolves instead of passing through and failing the deploy.

## Testing

Unit tests per kit, plus reading the manifest each built output emits: every pinned function takes the region, and unset emits none. Region claims were checked against live deploys of `firestore-vector-search` and `storage-resize-images`; the mapping itself is not deploy-verified on a multi-region database.

## Decisions for the reviewer

- Dual-region buckets (`nam4`, `eur4`, `asia1`) are not offered as values. The READMEs point those users at a member region instead, since a trigger in either half of the pair fires for the bucket, and all six are already in the list. Not deploy-verified.
- The #3101 agreement with the firebase-tools team rests on the trigger's `database` being a param expression and on `ext:migrate` exporting `DATABASE_REGION`. Only the `firestore-send-email` and `firestore-bigquery-export` extensions declare that param, so this rollout needs confirming: on the #11020 basis for the storage and RTDB kits, and on the split-placement basis for `firestore-counter` and `firestore-vector-search`.
- Issue #2974 records that `LOCATION` params were removed in kits. The entry needs updating to separate the extension's function-location param from these resource-location params.
- `firestore-incremental-capture` already pinned all five functions to `LOCATION`, so it only gained a label and description. Confirm that reusing it beats adding `DATABASE_REGION` beside it.


IzaakGough:
Thanks, that is right and it matches what `src/functions/kits/install.ts` does: `promptAndWriteKitParams` runs at install and writes `.env` before any deploy, so the second-deploy behaviour only applies when the prompt happens during `firebase deploy` with the value still missing.

I took the substance but not the wording, for two reasons. The command is `firebase functions:kits:install` (`src/commands/functions-kits-install.ts:16`) rather than `kit:install`, and `firebase ext:migrate` writes `.env` the same way, so both are worth naming. The corrected paragraph now reads:

> `firebase functions:kits:install` and `firebase ext:migrate` prompt for this value and write it to `.env` before anything is deployed, so a single deploy places the function correctly. If you instead run `firebase deploy` with the value still missing from `.env`, the prompt comes after discovery has already chosen a region, so your answer only takes effect on the following deploy.

Applied to all eight kits in this PR, plus `firestore-bigquery-export` and `firestore-send-email`, which had already shipped the original claim.

IzaakGough:
## Sequencing against the Vertex AI location work

This PR interacts with #3162 and #2943, which take opposite routes on the same question, and #3028 is still the open decision. Worth settling that order before this merges.

**The interaction.** `DATABASE_REGION` decides the function's region, and `firestore-translate-text` (`src/translate/common.ts:75`) and `firestore-vector-search` (`src/embeddings/client/genkit.ts:41`) both call Vertex AI in whatever region the function runs in. So placing a function next to its database also moves its Vertex AI call there. That is already true on `kits` and is not something this PR introduces, but this PR is what makes the region a derived value rather than `us-central1`. I have noted it in both READMEs.

`firestore-genai-chatbot` is not coupled today: its Vertex location comes only from `VERTEX_AI_MODEL_LOCATION` (`src/export-config.ts:208`). #3162 is what adds the `FUNCTION_REGION` fallback, so that kit becomes coupled only if #3162 lands.

If #2943 wins instead and those kits move to the `global` endpoint, the coupling goes away and the README notes I added to translate and vector search should come out.

**Suggested order:** settle #3028, land whichever of #3162 / #2943 wins, then I rebase this on top and correct the notes to match.

**File overlap with #3162**, whichever goes second needs a merge:

```
kits/firestore-genai-chatbot/CHANGELOG.md
kits/firestore-genai-chatbot/README.md
kits/firestore-genai-chatbot/src/export-config.ts
kits/firestore-translate-text/README.md
kits/firestore-vector-search/README.md
```

`src/export-config.ts` is the only code file: this PR adds `region?: string` to `DeployTimeOptions`, #3162 reworks the Vertex resolution in the same file. #3162 is `CLEAN` right now, so this PR is the one that will go conflicting if #3162 merges first. Happy to take that side of it.

IzaakGough:
Self-review. I checked the region claims against live deploys of the vector-search and storage-resize-images kits in a test project whose default Firestore database is in `us-east1`. Four things need changing.

**1. The mismatch guidance contradicts itself, and incremental-capture has it the wrong way round.**
Deployed vector-search with `DATABASE_REGION=us-central1` against the `us-east1` database. The deploy succeeded, Eventarc created the trigger in `us-east1` pointing at the `us-central1` function, and a document write fired it. So `firestore-incremental-capture/src/config.ts:98` ("a 2nd gen Firestore trigger only fires for a function in the same region as its database") and `firestore-incremental-capture/README.md:356` ("fails the deploy rather than firing nothing") are both wrong, and the line the other five kits get is right. Verified on vector-search, which has the same trigger shape (no `database` option, default database); incremental-capture itself was not deployed.

**2. "Lands in `us-central1` on a first deploy" is wrong for the Firestore kits.**
A clean first deploy with `DATABASE_REGION` empty put `embedOnWrite` and `queryOnWrite` in `us-east1`, next to the database, and the other six functions in `us-central1`. The same sentence is accurate for the storage kits: storage-resize-images with no `BUCKET_REGION` against a `europe-west1` bucket resolved to `us-central1` and was rejected with `A function in region us-central1 cannot listen to a bucket in region europe-west1`. So this wants correcting per kit, not globally, and it is worth deciding whether the Firestore kits need a required region prompt at all when the CLI already derives it.

**3. `EMBEDDING_PROVIDER`'s us-central1 note is out of date.**
`gemini-embedding-001` on Vertex answered in `us-central1`, `us-east1`, `europe-west1` and `me-central2`, and a deployed run with `EMBEDDING_PROVIDER=vertex` and `DATABASE_REGION=us-east1` wrote a real vector with `state: COMPLETED`. The regions that do fail are `europe-north2`, `africa-south1` and `europe-west12` (publisher model not found), `northamerica-south1` (404) and `europe-west10` (502), all of which `DATABASE_REGION` now offers. Naming those beats repeating the us-central1 claim.

**4. `LOCATION` in incremental-capture cannot honour its own description.**
The description says it must match the Firestore database location, but `LOCATION_OPTIONS` omits 21 of the 46 locations the Firestore API lists, 18 of them regional (`africa-south1`, `asia-south2`, `asia-southeast3`, `australia-southeast2`, `europe-north2`, `europe-southwest1`, `europe-west8/9/10/12`, `me-central1/2`, `me-west1`, `northamerica-northeast2`, `northamerica-south1`, `southamerica-west1`, `us-east5`, `us-south1`). The constraint is Dataflow's, so the description should say an unlisted database location is unsupported rather than the list growing.

Not a problem: the empty-`DATABASE_REGION` split is real but harmless. Every enqueue site sits in a task function, and those stay together in `us-central1` with the queues, so the failure the comment in `firestore-vector-search/src/index.ts:93` describes cannot happen here. I ran the backfill chain under the split config and it completed. One aside worth knowing: region inference only applies to functions that do not exist yet, so a redeploy keeps whatever region a function already had.

kits/firestore-send-email/README.md
```diff
@@ -178,25 +178,29 @@ is honored: the function deploys near your database.
...
+
+`firebase functions:kits:install` and `firebase ext:migrate` prompt for this
+value and write it to `.env` before anything is deployed, so a single deploy
+places the function correctly. If you instead run `firebase deploy` with the
+value still missing from `.env`, the prompt comes after discovery has already
+chosen a region, so your answer only takes effect on the following deploy.
+
+With an explicit empty `DATABASE_REGION=` line in `.env`, the function declares
+no region and the Firebase CLI resolves one at deploy time: it keeps the region
+it is already deployed in, and on a first deploy the CLI places it next to the
```
cabljac: I don't think this is right for this kit. The trigger passes `database: deploy.database` (src/index.ts:102), and firebase-tools runs resolveDefaultRegionsForBuild before params resolve (prepare.ts:332 vs resolveBackend at :343), so the Firestore lookup sees the literal `{{ params.DATABASE }}`, throws, and falls back to us-central1. That is #11020, which the PR already cites. The four kits that pass no database option are fine, it is only send-email and bigquery-export where the param-ized database breaks the lookup.

Suggest: say the function lands in us-central1 with an empty value and link #11020. Read from firebase-tools main, not deploy-verified, so shout if you have seen it place correctly.

kits/firestore-bigquery-export/README.md
```diff
@@ -397,23 +397,30 @@ it is honored: the functions deploy near your database.
...
+`firebase functions:kits:install` and `firebase ext:migrate` prompt for this
+value and write it to `.env` before anything is deployed, so a single deploy
+places the functions correctly. If you instead run `firebase deploy` with the
+value still missing from `.env`, the prompt comes after discovery has already
+chosen a region, so your answer only takes effect on the following deploy.
+
+With an explicit empty `DATABASE_REGION=` line in `.env`, the functions declare
+no region and the Firebase CLI resolves one at deploy time: a function keeps
+the region it is already deployed in, and on a first deploy the CLI resolves
+each function separately, so `fsexportbigquery` lands next to the database and
```
cabljac: Same issue as send-email: this kit passes `database: CONFIG_EXPRESSIONS.database` (src/index.ts:171), so the CLI's default-region lookup sees the literal param expression, throws, and `fsexportbigquery` falls back to us-central1 rather than landing next to the database (#11020). Suggest saying all three functions land in us-central1 with an empty value. Source-read, not deploy-verified.

kits/rtdb-limit-child-nodes/src/config.ts
```diff
@@ -117,7 +129,13 @@ export function envDeployOptions(): DeployTimeOptions {
       ? params.nodePath.toCEL()
       : params.nodePath.value();
 
+  // The region option does not accept a param expression, so the value is read
```
cabljac: "The region option does not accept a param expression" isn't true of firebase-functions v7: options.d.ts declares `region?: SupportedRegion | string | Expression<string> | ResetValue`, and firebase-tools build.ts:509-521 resolves a string-param region. I checked `onValueCreated({ region: defineString("DATABASE_REGION") }).__endpoint.region` and it holds the Expression.

For the Firestore and storage kits the hardcoding still stands, the multi-region mapping needs a nested ternary the CLI's CEL can't express, so please reword the comment there to that constraint. For RTDB there is no mapping, so `region: params.databaseRegion` should work directly and would remove both README caveats (first-deploy prompt value being ignored, and placement depending on .env being present at discovery).

kits/firestore-vector-search/src/index.ts
```diff
@@ -89,23 +90,37 @@ const FUNCTION_SECRETS = [geminiApiKey, openAiApiKey];
 // Only the task functions reach getSingleEmbedding, but every function here
```
cabljac: This comment is about `secrets: FUNCTION_SECRETS` but the new REGION_OPTION block now sits between them, so it reads as describing the region. Please move it down to sit above the option objects.

cabljac (changes requested):
@IzaakGough this looks like the right general design, the mappings match firebase-tools' location.ts and the mutation checks on the new tests hold up. Requesting changes for a few things, mostly docs:

- The send-email and bigquery-export READMEs say an empty DATABASE_REGION lands the function next to the database on first deploy. For those two kits I don't think that holds, see inline.
- The "region does not accept a param expression" comment is repeated across the kits and isn't true of the SDK. For RTDB that changes the design, see inline.
- This is breaking for existing installs: a required select with no default makes `firebase deploy --non-interactive` throw until .env gains the new key, and the interactive prompt has no "keep no region" option, so any answer deletes and recreates the function. The CHANGELOGs describe it but nothing flags it as breaking. Please add a BREAKING marker or an upgrade note to each.
- PR body: "any database or bucket outside us-central1 failed the deploy" only holds for storage and RTDB. The Firestore kits deployed fine cross-region (#11020 says it succeeds silently, and the READMEs say so too), the win there is latency. Worth rewording so the motivation is accurate.

Tests: ran vitest per kit on the branch, all green except storage-resize-images content-filter.test.ts, which fails on a missing @genkit-ai/core/schema import and is pre-existing, not yours. No live deploy on my side, so please treat the region claims below as source-read rather than deploy-verified and tell me if you have seen otherwise.

kits/rtdb-limit-child-nodes/src/config.ts
cabljac: this might be an RC thing

kits/firestore-send-email/README.md
IzaakGough: You're right, and I confirmed it: `stackToWire` in firebase-functions turns `database: params.databaseId` into `{{ params.DATABASE }}` in `eventFilters`, `prepare.ts` runs `resolveDefaultRegionsForBuild` before `resolveBackend`, and `firestore.getDefaultRegion` hands that literal to `getDatabase`, whose throw is swallowed into `FALLBACK_DEPLOYMENT_REGION`. So an empty value lands `processQueue` in `us-central1`.

Fixed in 2d7be58 - the section now says `us-central1`, explains that the default region resolves before params do, and links #11020. Still source-read on my side too, not deploy-verified.

kits/firestore-bigquery-export/README.md
IzaakGough: Same fix here in 2d7be58: all three functions land in `us-central1` with an empty value, with the reason and a link to #11020. That also drops the "splitting the instance across two regions" warning, which only existed because I had `fsexportbigquery` resolving next to the database.

kits/rtdb-limit-child-nodes/src/config.ts
IzaakGough: The comment was wrong, thanks. I checked `options.d.ts` and then the manifest: `onValueCreated({ region: defineString("DATABASE_REGION") }).__endpoint.region` holds the Expression, `stackToWire` emits `{{ params.DATABASE_REGION }}`, and `build.ts` resolves it through `resolveList`/`resolveString`.

2d7be58 rewords it everywhere it appeared (8 kits). For the Firestore and storage kits it now names the real constraint: the location to Cloud Run region mapping needs a nested ternary, and `cel.ts` only parses identity, one comparison, and one ternary.

I've left RTDB reading `process.env` for now, and the comment there says why rather than claiming the option won't take an expression. Two things the expression would cost: the explicit empty `DATABASE_REGION=` escape hatch the README documents would resolve to a region of `""` rather than to no region at all, and `normalizeRegion`'s trim/lowercase for a hand-edited `.env` goes away. Both look worth keeping, but I haven't deploy-tested the empty case either way, so happy to be talked round if you think the first-deploy win outweighs them.

kits/firestore-vector-search/src/index.ts
IzaakGough: Moved in 2d7be58, it now sits directly above the option objects.

kits/rtdb-limit-child-nodes/src/config.ts
IzaakGough: Tried it, and you were right: done in 045896f5. Ran real discovery on the built kit and resolved the manifest through firebase-tools' own `buildFromV1Alpha1` / `toBackend`: the wire manifest carries `{{ params.DATABASE_REGION }}` and resolves to the param value, with no `.env` needed at discovery.

On the rc question, that isn't it. `stackToWire` emits the same `{{ params.DATABASE_REGION }}` on 6.3.2, 6.6.0, 7.0.0, 7.3.2 and 7.3.3-rc.2, and the kits are CJS so there's no dual-package hazard around the `instanceof Expression` check. The version sensitivity is on the CLI side and it favours the expression: the CEL region block in `build.ts` is unchanged back to 13.0.0, whereas reading `process.env` needs 15.28.0 exactly, since 15.27.0 passes only `firebaseEnvs` to `discoverBuild` and 15.28.0 added `...userEnvs`.

One thing I dropped rather than kept: an empty `DATABASE_REGION=` now resolves to a region of `""` rather than to no region, and nothing in `validate.ts` catches it. So the value is required, and the README no longer offers the empty line as a way to let the CLI choose. Anyone who wanted `us-central1` can just pick it. Still not deploy-verified, only resolved offline.

CorieW (changes requested):
The code is correct: mappings match firebase-tools `location.ts`, `region` accepts a string param, tests cover unset, mapped and all-functions cases, and CI is green. Two items need resolving before merge.

1. **Scope.** firebase/firebase-tools#11020 only affects triggers whose `database` is a param expression. `firestore-counter`, `firestore-translate-text`, `firestore-genai-chatbot` and `firestore-vector-search` trigger on `(default)`, so the CLI already infers the region (your self-review shows `embedOnWrite` in `us-east1`). For `translate-text` and `genai-chatbot`, `DATABASE_REGION` is a new required prompt that changes nothing. Drop it from those two kits, or reframe the PR as "explicit single-region placement" and record that decision. Either way, fix the PR body: "the function lands in `us-central1` away from its database" is false for these four kits.

2. **Ledger.** #2974 "Valid differences" records `LOCATION params have been removed in kits, as kits builds this feature in.` This PR adds one to seven kits. Update the ledger and confirm the firebase-tools agreement from #3101 covers this rollout.

kits/firestore-translate-text/README.md
```diff
@@ -203,6 +207,39 @@ Its service account needs `roles/eventarc.eventReceiver` and `roles/run.invoker`
 on top of `roles/datastore.user`; the Firebase CLI grants these for you. The
 Cloud Translation API is still required whichever provider you choose.
 
+### DATABASE_REGION decides where the function runs
```
CorieW: Non-blocking, migration note. `firebase ext:migrate` already exports the extension's function location as `FUNCTION_DEFAULT_REGION` (firebase-tools `src/extensions/export.ts:208-209`). The kit does not read it, so a migrated user is prompted for `DATABASE_REGION` and may land the function somewhere other than where the extension ran it. One sentence here saying the migrated value is not used would cover that.

kits/firestore-genai-chatbot/src/config.ts
```diff
@@ -97,6 +98,59 @@ const POSITIVE_INT_VALIDATION = {
 };
 
 const params = {
+  databaseRegion: defineString("DATABASE_REGION", {
```
CorieW: Non-blocking, migration note. The GoogleCloudPlatform extension exports its `LOCATION` param on `ext:migrate`. This kit reads only `DATABASE_REGION`, so the migrated value is ignored and the user is prompted again. If the param stays, either read `LOCATION` as a fallback or say in the README that it is ignored.

kits/firestore-counter/src/config.ts
```diff
@@ -26,6 +27,59 @@ export interface ConfigExpressions {
 }
 
 const params = {
+  databaseRegion: defineString("DATABASE_REGION", {
```
CorieW, line 30: Non-blocking. The list grows from the extension's ~24 regions to 45. `controllerCore` is a scheduled function, and Cloud Scheduler availability in `europe-west10`, `europe-west12`, `europe-north2`, `northamerica-south1`, `me-central2` and `africa-south1` is not checked here. A README caveat or one deploy to a new region would close this.

kits/storage-resize-images/src/region.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Maps a Cloud Storage bucket location to the Cloud Run region the functions
+ * should deploy to. The lookup is case-insensitive and ignores surrounding
+ * whitespace, as the CLI's own region handling is. Regional locations pass
+ * through lowercased; an unset or blank location returns `undefined`, meaning
+ * the functions declare no region.
+ *
+ * Dual-region locations (`nam4`, `eur4`, `asia1`, ...) are not mapped and pass
```
CorieW, line 35: Non-blocking. The README points dual-region users (`nam4`, `eur4`, `asia1`) at a member region. That places the function in a single region while the Eventarc trigger follows the bucket's dual-region location; the PR says this is not deploy-verified. Add a line that dual-region placement is untested, or confirm with one deploy.

kits/firestore-bigquery-export/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
...
+- fix: restore explicit function placement from `DATABASE_REGION`, now with the Firestore-location-to-Cloud-Run-region mapping. The `DATABASE_REGION` parameter is back and all three functions deploy to the region derived from it: regional locations pass through unchanged, and the multi-region locations map to a Cloud Run region (`nam5`/`nam7` to `us-central1`, `eur3` to `europe-west1`) instead of failing the deploy. With an explicit empty `DATABASE_REGION=` line in `.env` the functions still declare no region and the CLI falls back as before (`us-central1` by default, `FIREBASE_FUNCTIONS_DEFAULT_REGION` to override); omitting the line fails a non-interactive deploy. Placement requires firebase-tools >= 15.28.0 (older CLIs do not load `.env` at discovery and keep the fallback). If your `.env` already carries `DATABASE_REGION` from an extension migration, upgrading to this version moves the functions to the mapped region on your next deploy, which deletes and recreates them.
```
CorieW: Suggestion. This line shipped in `0.0.2-rc.12`. Add a new unreleased entry for the empty-line behavior instead of editing the released one.

kits/firestore-send-email/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
...
 - chore: run on firebase-functions ^7.3.3-rc.1, the same release candidate as the other kits
-- fix: map `DATABASE_REGION` to a valid Cloud Run region before using it as the function region. Firestore multi-region locations (`nam5`, `nam7`) now deploy the function to `us-central1` and `eur3` to `europe-west1` instead of failing the deploy; regional locations pass through unchanged. The value is matched case-insensitively. With the parameter unset or empty the function declares no region and the Firebase CLI resolves one at deploy time.
+- fix: map `DATABASE_REGION` to a valid Cloud Run region before using it as the function region. Firestore multi-region locations (`nam5`, `nam7`) now deploy the function to `us-central1` and `eur3` to `europe-west1` instead of failing the deploy; regional locations pass through unchanged. The value is matched case-insensitively. With an explicit empty `DATABASE_REGION=` line in `.env` the function declares no region and the Firebase CLI resolves one at deploy time; omitting the line fails a non-interactive deploy.
```
CorieW: Suggestion. This line has already been released. Add a new unreleased entry for the empty-line behavior instead of editing it.

CorieW (commented):
Inline non-blocking notes, following up on my earlier review.

IzaakGough:
Thanks, the scope point holds for two of the four kits and the param is gone from those in 1385a9ea.

`firestore-translate-text` and `firestore-genai-chatbot` have a single Firestore trigger each on the default database with no `database` option, so the CLI places them and `DATABASE_REGION` was a required prompt for nothing. Reverted, with the README claim those sections had before corrected: they said a no-region function lands in `us-central1`, and it does not, it lands next to the database.

`firestore-counter` and `firestore-vector-search` are different. The inference only reaches Firestore-triggered functions, and these kits have others: `controllerCore` is `onSchedule`, and six of vector-search's eight are task or callable. My live deploy with no region set shows the split, `embedOnWrite` and `queryOnWrite` in `us-east1` and the other six in `us-central1`, which is what the param prevents. Keeping it there, and the PR body now gives that as the reason rather than #11020.

On the agreement: #3101 and #3102 rest it on the trigger's `database` being a param expression and on `ext:migrate` exporting `DATABASE_REGION` into `.env`. Only `firestore-send-email/extension.yaml:82` and `firestore-bigquery-export/extension.yaml:233` declare that param, so neither half carries over to the four Firestore kits. The storage and RTDB kits do hit #11020 (they pass `bucket` and `instance` as param expressions), so that half is the same reasoning. Happy to take the counter and vector-search case back to the firebase-tools team before this merges, unless you would rather do that.

On #2974: the bullet stays true if it is scoped to what it meant. The extension's `LOCATION` is a system param choosing the function's region, and kits do build that in. What these kits add is a resource-location param saying where the database or bucket lives, which the CLI cannot infer when the resource is a param expression or when the function has no resource trigger at all. I can edit the entry to draw that line and link #3101, #3102 and #11020, if that works for you.

IzaakGough:
Following up on the ledger point: I have updated the entry in #2974 rather than leaving it as an offer. The `LOCATION` line is now two bullets, one for the extension's function-region param (removed, kits build it in) and one for the resource-location params, with shipped and proposed marked separately so the ledger does not state this PR as fact. Details are in https://github.com/firebase/extensions/issues/2974#issuecomment-5679361415, and happy for you to reword it.

The firebase-tools half of that point is still open.

kits/firestore-translate-text/README.md
IzaakGough: Done in 1385a9ea and 2423e2ab. `DATABASE_REGION` is gone from this kit, since the single trigger has no `database` option and the CLI already places the function next to the database. The migration point stands either way, so the README now says the `FUNCTION_DEFAULT_REGION` that `ext:migrate` writes is not read and a migrated instance does not keep the region the extension ran in. The same sentence went into the seven kits that keep a region param, since a migrated .env carries the key there too.

kits/firestore-genai-chatbot/src/config.ts
IzaakGough: Done in 1385a9ea. The param is dropped from this kit for the scope reason, so nothing competes with the migrated value now. The README section says the extension's `LOCATION` and the `FUNCTION_DEFAULT_REGION` written by `ext:migrate` are both unread, and describes what the CLI does instead: it places the function next to the database when it does not exist yet and keeps the existing region on a redeploy.

kits/firestore-counter/src/config.ts
IzaakGough: Good catch, and it is worse than six. Checked it in 2423e2ab: `gcloud scheduler locations list` and the published region list agree on 30 Cloud Scheduler regions, and twelve of the 45 locations this select offers resolve to a region outside that set: `us-east5`, `northamerica-northeast2`, `northamerica-south1`, `southamerica-west1`, `europe-north1`, `europe-north2`, `europe-southwest1`, `europe-west10`, `europe-west12`, `asia-south2`, `australia-southeast2` and `africa-south1`. `me-central2` is fine. Since the CLI creates the job in the function's own region (firebase-tools `fabricator.ts:1117`), those twelve fail the deploy.

The README now names them and says to give `DATABASE_REGION` the nearest location that does have Scheduler, which costs a cross-region hop but works, and the parameter description points at that list. Read from the region lists rather than a failed deploy.

Left the select at 45 rather than trimming it, because the question it asks is where your database is, and a user in `europe-north1` should still be able to answer it. If you would rather the twelve mapped to a nearby Scheduler region automatically, that is a design change I am happy to make, but it moves the triggered functions away from the database too, so I did not want to pick it unilaterally.

kits/storage-resize-images/src/region.ts
IzaakGough: Went and tested it instead of caveating it, so please read the reply below rather than this one. Short version: I created a `nam4` bucket and deployed against it, the member-region advice holds, and the README and this doc comment now state the result. be86a336 replaces the caveat I first added in 2423e2ab.

kits/firestore-bigquery-export/CHANGELOG.md
IzaakGough: Agreed, and confirmed it shipped: `npm view` lists the version carrying that line. Restored the released bullet word for word in 2423e2ab and put the empty-line correction in a new entry at the top instead.

kits/firestore-send-email/CHANGELOG.md
IzaakGough: Agreed, and confirmed it shipped: `npm view` lists the version carrying that line. Restored the released bullet word for word in 2423e2ab and put the empty-line correction in a new entry at the top instead.

kits/rtdb-limit-child-nodes/src/config.ts
IzaakGough: Worth flagging here since it bears on the mapping: firebase/firebase-tools#11026 is open and fixes #11020. It moves param evaluation ahead of region resolution and substitutes the trigger filters just for the lookup, so `database: "{{ params.DATABASE }}"` in `eur3` resolves to `europe-west1` through the CLI's own `FIRESTORE_DUAL_REGION_TO_REGION_MAPPING` and `STORAGE_MULTI_REGION_TO_REGION_MAPPING`. Review-required, last touched 2026-09-03.

If it lands, the tables in each kit's `region.ts` duplicate the ones the CLI already has, and the #11020 reason for the param goes away for `storage-resize-images`, `speech-to-text` and `rtdb-limit-child-nodes`, plus the two already shipped. The scope argument that dropped `DATABASE_REGION` from translate-text and genai-chatbot would then apply to those too. It would also remove the second-deploy caveat and the firebase-tools 15.28.0 floor these READMEs document, since the prompt would then run before region resolution rather than after it.

Two things it does not cover: `firestore-counter` and `firestore-vector-search` still need the param, since their scheduled, task and callable functions have no event source to resolve a region from, and it does nothing about Cloud Scheduler availability.

Going ahead with the param for now. #11026 is expected to land at some point, so this is a workaround with a known exit rather than a permanent feature, but it places the functions correctly on every CLI version and those two kits need it either way. Once #11026 ships, dropping the param from the three kits that only needed it for #11020 is a small follow-up.

kits/storage-resize-images/src/region.ts
IzaakGough: Tested it properly, and the advice holds. Replaced the caveat with the result in be86a336.

Created a `nam4` bucket, deployed this kit with `BUCKET_REGION=us-central1` (one half of the pair), and uploaded an image. The deploy placed `generateResizedImage` in `us-central1`, Eventarc created the trigger at `projects/izaak-testing/locations/nam4/triggers/generateresizedimage-243603` with `bucket: izaak-nam4-sri-test-2026` and a destination of `locations/us-central1/functions/generateResizedImage`, and the resized output appeared about eight seconds after the upload. So the trigger lives in the bucket's dual-region location and dispatches to a function in one member region.

Also confirmed the other half: with `BUCKET_REGION=nam4` the deploy fails at `generateUploadUrl` with `403, Location nam4 is not found or access is unauthorized`, which is the mapping gap the section describes.

Run on `storage-resize-images`; `speech-to-text` carries the same note attributed to that. Test resources are deleted.

CorieW (approved):
lgtm

PR #3175, opened 2026-09-10
chore(delete-user-data): drop identity type shim, bump firebase-functions to 7.3.3-rc.2

CorieW:
`firebase-functions@7.3.3-rc.2` publishes real type declarations for the v2 Auth triggers (`onUserCreated` / `onUserDeleted` are no longer `@internal`). The hand-written augmentation in `src/identity-shim.d.ts` is now obsolete.

Runtime is unchanged: rc.1 to rc.2 differs only in comments in `lib/v2/providers/identity.js`. Same `gcfv2` endpoint, same `google.firebase.auth.user.v2.deleted` event, same payload. No behavior change for `clearData`.

## Changes

- Delete `kits/delete-user-data/src/identity-shim.d.ts`.
- Bump `firebase-functions` to `^7.3.3-rc.2` in `package.json` and `npm-shrinkwrap.json`.

## Verification

`npm ci && npm run build && npm test`: build clean without the shim, 102 tests pass.

## Notes

- Export is still `@beta` upstream.
- The wider rc.2 API (`AuthOptions.tenantId`, `V1Compat` overload) is not adopted here.
- Other 12 kits stay on rc.1 (#3160); aligning them is a separate change.


IzaakGough (approved):
lgtm, thanks for removing this shim file

PR #3178, opened 2026-09-10
 prototype(storage-resize-images): Update firebase-admin and     firebase-functions peer dependencies to higher versions.

ajperel:
Prototyping how to update kits and wrapper dependencies in sync to avoid
    issues with firebase-admin version skew across nested packages.


wandamora (approved):

ajperel:
This prototype will go out as [0.0.2-rc.6](https://www.npmjs.com/package/@firebase-function-kits/storage-resize-images/v/0.0.2-rc.6)

Issue #3139, opened 2026-09-07
decision(firestore-vector-search): status field shape, nested per-instance vs flat


cabljac:
The extension wrote a nested, per-process status via `@invertase/firebase-extension-utilities`:

```
status.<EXT_INSTANCE_ID> = { state, startTime, updateTime, completeTime, createTime }
status.textQuery = { ... }
```

with states `PROCESSING | COMPLETED | ERROR | BACKFILLED | FAILED_BACKFILL`, two writes per document.

The kit writes a flat field, one write per document:

```
status = { state: "COMPLETED" }
status = { state: "ERROR", message }
```

No `PROCESSING`, no timestamps, no `BACKFILLED` / `FAILED_BACKFILL`.

Any consumer reading `status.<instanceId>.state` breaks on migration, and documents embedded by the extension carry a status the kit's terminal-state guard does not recognise until the field is flattened. https://github.com/firebase/extensions/pull/3092 (#3011) restores the skip-on-terminal-state rule against the flat shape and explicitly leaves the shape itself out of scope.

Decide one of:

- Parity: restore the nested per-instance shape and the full state set, so extension consumers and already-embedded documents carry over unchanged.
- Keep flat: document the shape change and the migration step (flatten `status.<instanceId>` to `status`) in the kit README and the Notes entries (#3035).

Parity ledger: #2974, firestore-vector-search §6.

PR #3180, opened 2026-09-11
fix(firestore-vector-search): restore the extension's status writes

CorieW:
Rebased onto `kits` now that #3092 has merged. Decision pending in #3139.

The extension's `FirestoreOnWriteProcessor` keyed each document's status by the instance id, marked the document `PROCESSING` before embedding it, and recorded timestamps. The kit wrote a flat `status.state` with no start event. This broke security rules, indexes, and consumers keyed to `status.<instance id>.state`, and left a migrated collection with a status the kit did not recognise.

This PR restores the extension's status writes. It was split out of #3092, which is scoped to the terminal-state skip (§6a/6b) and leaves the status shape (§6) as an open decision in the parity ledger (#2974). It also touches the backfill and update task paths, which overlap #3097; rebase whichever merges second.

## Changes

- Write the status at `status.<instance id>` with `startTime`, `updateTime`, `completeTime`, and a `createTime` carried forward, as the extension did.
- Mark the document `PROCESSING` before embedding, and read that path for the terminal-state guard.
- Record a failed embed as `ERROR` without rethrowing; the extension left the error in the logs.
- Skip a write that leaves the input field untouched, matching the extension's `fieldDependencyArray: [inputField]`.
- Backfill and update tasks write `BACKFILLED` or `FAILED_BACKFILL`, skip a document whose status is anything but `BACKFILLED`, and do not throw on a failed embed, matching the extension's backfill handler.
- Write with `update()` rather than `set({ merge: true })`, so a document deleted mid-embed is not recreated.
- README: replace the status-shape divergence with the remaining query-document difference, and correct the `UPDATE_ON_CONFIGURE` and re-embed notes.

## Testing

`vitest run` in `kits/firestore-vector-search`: 154 tests pass. `tsc -b` clean.

Differential run of the kit and the extension's `FirestoreOnWriteProcessor` against one Firestore emulator: 41 of 41 scenarios produced identical documents and 5 of 5 trigger-recursion scenarios produced identical event and embed counts.

Related: #3011, #3092, #3097, #3139, #2974.


kits/firestore-vector-search/src/handlers.ts
```diff
@@ -108,37 +140,65 @@ export async function handleEmbedOnWrite(
...
+    // failed and, with retries enabled, re-embed from the same stale event.
+    await after.ref.update(
+      statusPath(ctx.config, "state"),
+      "ERROR",
+      statusPath(ctx.config, "updateTime"),
+      FieldValue.serverTimestamp()
     );
     logs.error("embedOnWrite", err);
-    throw err;
   }
```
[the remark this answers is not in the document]
CorieW: Declined: the suggested change would diverge from the extension, and this PR's goal is parity with it.

The extension has the same exposure. Its `PROCESSING` write runs outside any `try`, and its status writes are unguarded `update()` calls, so a document deleted mid-embed throws `NOT_FOUND` there too (`@invertase/firebase-extension-utilities@0.1.3`):

```js
// lib/firestore-onwrite-processor/index.js
37:  await change.after.ref.update(update);        // writeStartEvent
105: await this.writeStartEvent(change, processesToRun);   // called outside the try
132: await this.writeCompletionEvent(change, finalOutput, completedProcesses, failedProcesses);
```

Swallowing `NOT_FOUND` here would change observable behavior relative to the extension. The race is an inherited problem and is tracked as a separate follow-up rather than fixed in a parity change. Resolving.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -282,22 +342,42 @@ async function embedPath(
...
+    // The extension marked the document and reported the failure in its task
+    // result rather than throwing, so a failed document was not retried.
+    await ref.update(
+      statusPath(ctx.config, "state"),
+      "FAILED_BACKFILL",
+      statusPath(ctx.config, "completeTime"),
+      FieldValue.serverTimestamp()
+    );
+    logs.error("backfillTask", err);
+  }
```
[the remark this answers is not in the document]
CorieW: Declined for the same reason as the `handleEmbedOnWrite` thread: the suggested change would diverge from the extension.

The extension's backfill `catch` also issues an unguarded `update()`, so a document deleted mid-embed throws `NOT_FOUND` there as well (`@invertase/firebase-extension-utilities@0.1.3`):

```js
// lib/distributed-tasks/firestore_backfill/handler_from_process.js:101-110
catch (e) {
    functions.logger.error(e);
    await admin.firestore().collection(options.collectionName).doc(document.id).update({
        [`${options.statusField}.${process.id}.state`]: "FAILED_BACKFILL",
        [`${options.statusField}.${process.id}.completeTime`]: admin.firestore.FieldValue.serverTimestamp(),
    });
```

This is an inherited problem, tracked separately from the parity work. Resolving.

CorieW (commented):
Two non-blocking edge-case notes from the deep review. Neither affects a document the extension or the kit ever writes; both are recorded so the gap is visible if we later decide to close it.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -108,37 +140,65 @@ export async function handleEmbedOnWrite(
...
+  const startTime = FieldValue.serverTimestamp();
+  const existingCreateTime = (
+    data[ctx.config.statusFieldName] as
+      | Record<string, { createTime?: unknown } | undefined>
+      | undefined
+  )?.[ctx.config.instanceId]?.createTime;
+  await after.ref.update(statusPath(ctx.config), {
+    state: "PROCESSING",
+    startTime,
+    createTime: existingCreateTime ?? after.createTime,
```
CorieW: **Non-blocking, edge case:** `??` here differs from the extension's `||`.

```js
// @invertase/firebase-extension-utilities@0.1.3
// lib/firestore-onwrite-processor/index.js:32
[this.orderField]: startData || change.after.createTime,
```

The two agree whenever the existing `createTime` is a Timestamp or absent, which is every value either implementation writes. They differ only if a consumer has stored a falsy non-null value (`0`, `""`, `false`) under `status.<instance id>.createTime`: the extension replaces it with the document's create time, the kit keeps it.

Switching to `||` would make the two identical. Not blocking, since no real document reaches this path.

kits/firestore-vector-search/src/handlers.ts
```diff
@@ -282,22 +342,42 @@ async function embedPath(
...
-      [ctx.config.outputFieldName]: FieldValue.vector(embedding),
-      [ctx.config.statusFieldName]: { state: "COMPLETED" },
-    },
-    { merge: true }
-  );
+  // The extension's `getValidDocs` skipped any document that already carried a
+  // status other than `BACKFILLED`, so a backfill or an update pass never
+  // re-embedded a document `embedOnWrite` had completed or failed.
+  const state = statusState(data, ctx.config);
+  if (state !== undefined && state !== "BACKFILLED") return;
```
CorieW: **Non-blocking, edge case:** the skip condition is not quite the extension's.

```js
// @invertase/firebase-extension-utilities@0.1.3
// lib/distributed-tasks/firestore_backfill/handler_from_process.js:72-75
data[options.statusField] &&
    data[options.statusField][process.id] &&
    data[options.statusField][process.id]["state"] &&
    data[options.statusField][process.id].state !== "BACKFILLED"
```

The extension skips when `state` is **truthy** and not `BACKFILLED`. `statusState` (l.97) returns the value only when it is a string, so the kit differs on two malformed inputs:

- `state: ""` : extension embeds (falsy), kit skips.
- `state: 1`, `true`, or an object : extension skips (truthy), kit embeds.

Both implementations only ever write the five known string states, so a migrated collection never hits either case. Matching the truthy check exactly (`const raw = data[status]?.[instanceId]?.state; if (raw && raw !== "BACKFILLED") return;`) would close the gap if we want it closed. Not blocking.

kits/firestore-vector-search/src/handlers.ts
CorieW: Fixed in f7db9aa81: the start event now uses `existingCreateTime || after.createTime`, matching `startData || change.after.createTime`. Covered by a test that writes `createTime: 0` and expects the document's create time; it fails on `??`.

kits/firestore-vector-search/src/handlers.ts
CorieW: Fixed in f7db9aa81: the skip now reads the raw state and tests it for truthiness, so `1`, `true`, and an object skip the document and `""` embeds it, as `getValidDocs` did. Four tests cover those cases and all fail on the string-typed check.

cabljac (approved):
lgtm

Issue #3152, opened 2026-09-08
decision(firestore-bigquery-export): DATASET_LOCATION is immutable in the extension but silently ignored on change in the kit


cabljac:
## Problem

The extension declares `DATASET_LOCATION` with `immutable: true` (`firestore-bigquery-export/extension.yaml`), so a reconfigure cannot change it. `firebase-functions/params` has no immutability, so the kit (`kits/firestore-bigquery-export/src/config.ts:236`) accepts a new value on any redeploy.

What happens today when the value changes on an existing install:

- The tracker's `initializeDataset` finds the dataset by id and skips creation (`firestore-bigquery-change-tracker/src/bigquery/index.ts:308`), so the dataset stays in its original location. The new value is silently ignored. Nothing in the tracker or the kit runs a query job with the configured location, so nothing fails.
- If `DATASET_ID` changes at the same time, a new dataset is created in the new location and the old one is left behind with its tables and view, same as the extension when the dataset id changes.

So the divergence is: the extension refuses the edit, the kit accepts it and does nothing. Not data loss, but a config that lies.

## Options

1. Guard in the lifecycle task: read the existing dataset's location and log an error (or fail `setupBigQuerySync`) when it differs from `DATASET_LOCATION`.
2. Notes entry only (#3035): document that `DATASET_LOCATION` only applies when the dataset is first created, and that moving a dataset means a new `DATASET_ID` plus a backfill.

Recommend 1 plus 2: the guard is a few lines and turns a silent no-op into a visible one.

## Cosmetic, same param family

The kit's `DATABASE_REGION` select offers `asia-southeast3` (`config.ts:177`), which the extension's `extension.yaml` does not list. Harmless (it is a real Firestore location), but a divergence to drop or note when this is picked up.

Parity ledger: #2974.

PR #3071, opened 2026-09-01
docs(firestore-bigquery-export): stop claiming the kit publishes onSuccess and onCompletion

cabljac:
The kit README's Events section claimed the function publishes `onStart`, `onError`, `onSuccess`, and `onCompletion`. The code publishes `onStart` and `onError` only: `src/handlers.ts` calls `recordStartEvent` and `recordErrorEvent`, and nothing calls `recordSuccessEvent` or `recordCompletionEvent`. The extension emitted `onSuccess` from its `syncBigQuery` task queue handler, which the kit deliberately removed, and the README's differences section already documents that. So the README was the wrong side; this corrects the Events section to list `onStart` and `onError` and point at the differences entry. Re-adding `onSuccess` emission (parity with the extension) is a separate decision, not attempted here.

Doc-only change, so no tests. Note #3066 also edits this README; its hunks (config table, provisioning, region sections) do not overlap this one.

Fixes #3018

PR #3181, opened 2026-09-11
docs(firestore-bigquery-export): note DATASET_LOCATION is not immutable

cabljac:
The extension declares `DATASET_LOCATION` with `immutable: true`, so a reconfigure could not change it. `firebase-functions/params` has no immutability, so the kit accepts a new value on any redeploy and the config can lie about where the dataset lives. Per the decision on #2974, this documents the behaviour instead of adding a guard.

The README's differences section gains a "DATASET_LOCATION is not immutable" entry: the extension refused the edit, the kit cannot, and on a redeploy the value is ignored with no error or warning. To change location, set a new `DATASET_ID` alongside the new `DATASET_LOCATION`, redeploy, and backfill with `fs-bq-import-collection`; the old dataset is left behind. The config table row now says the value is used only when the dataset is created.

Verified from code, not a live deploy: the tracker (2.2.1) passes `datasetLocation` only to `bq.dataset(id, { location })`, and `@google-cloud/bigquery` sends that location only on `dataset.create()` and `dataset.createQueryJob()`. `initializeDataset` skips creation when the dataset exists, and the tracker never calls `dataset.query`; the materialized view uses `bq.query()` with no location, which BigQuery resolves from the referenced table. So a mismatch is a true no-op on every path.

README only; the CHANGELOG is not touched, matching #3071 and #3131.

Fixes #3152

### DIFF
--- a/kits/firestore-bigquery-export/README.md
+++ b/kits/firestore-bigquery-export/README.md
@@ -121,7 +121,7 @@ loads them at deploy time and prompts for any required values that are missing.
 | `datasetId`                      | `DATASET_ID`                        | no       | `firestore_export` | BigQuery dataset                                                   |
 | `tableId`                        | `TABLE_ID`                          | no       | `posts`            | BigQuery changelog table                                           |
 | `databaseRegion`                 | `DATABASE_REGION`                   | yes      | (prompted)         | Firestore database location; also places the functions             |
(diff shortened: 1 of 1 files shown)


CorieW (approved):
lgtm

Issue #3165, opened 2026-09-09
decision(bigquery-firestore-export): TRANSFER_CONFIG_NAME enables the extension's dead link-config branch


CorieW:
The kit exposes `TRANSFER_CONFIG_NAME` and runs a link-an-existing-scheduled-query branch that the extension ships but can never reach. #3030 was meant to rule on this and was closed without a recorded decision for it.

**Extension** (GoogleCloudPlatform/firebase-extensions)
- `transferConfigName` is hardcoded to `undefined`, no param exists: [functions/src/config.ts#L25](https://github.com/GoogleCloudPlatform/firebase-extensions/blob/46dab5e3a3d67d9826ef14524c2798bab131e5b4/bigquery-firestore-export/functions/src/config.ts#L25)
- so the `if (config.transferConfigName)` branch in `upsertTransferConfig` is dead code: [functions/src/index.ts#L42](https://github.com/GoogleCloudPlatform/firebase-extensions/blob/46dab5e3a3d67d9826ef14524c2798bab131e5b4/bigquery-firestore-export/functions/src/index.ts#L42)

**Kit** (`kits` branch)
- `TRANSFER_CONFIG_NAME` is a real param: [src/config.ts#L92](https://github.com/firebase/extensions/blob/025e39ae46a601ca9989c7310a6bcea1ad087542/kits/bigquery-firestore-export/src/config.ts#L92)
- `handleUpsertTransferConfig` takes the link branch when it is set: [src/handlers.ts#L111](https://github.com/firebase/extensions/blob/025e39ae46a601ca9989c7310a6bcea1ad087542/kits/bigquery-firestore-export/src/handlers.ts#L111)

Impact: kit users can adopt an existing scheduled query, which no extension user could do. Behaviour beyond the extension, not a regression, but it is an unreviewed feature surface (#2970 is one defect already found in it).

**Options:**
1. Forward: keep the param, fix #2970 via #2960, document it in the kit README.
2. Parity: remove `TRANSFER_CONFIG_NAME` and the link branch (breaking config change for anyone already using it on `kits`).

Related: #2974 (ledger, Behavior divergence section), #3030 (the prior decision issue, closed without ruling on this item), #2970, #2960.

cabljac:
[...]

Issue #3185, opened 2026-09-11
bigquery-firestore-export: re-add TRANSFER_CONFIG_NAME adoption of an existing scheduled query, removed for parity


cabljac:
The kit shipped a `TRANSFER_CONFIG_NAME` param that adopts an existing BigQuery scheduled query instead of creating one. The extension carries the same branch as dead code (`transferConfigName` hard-coded to `undefined`), so no extension user could ever reach it. #3165 ruled parity for kits-stable: the param and the link branch come out.

This tracks bringing the feature back as a deliberate addition once kits-stable ships.

What exists already:

- #2970 documents the one defect found in the branch: the adopted query kept notifying its old Pub/Sub topic, so `processMessages` never fired and no output reached Firestore, with nothing logged.
- #2960 is a live-verified fix: an update masked to `notification_pubsub_topic` alone, skipped when the topic already matches, leaving the adopted query's SQL, schedule and destination untouched. Closed unmerged with the parity ruling; the branch is the starting point.

Open questions before re-adding: whether to refuse a config whose topic does not match rather than repoint it, what happens when the adopted query belonged to an extension instance that is later uninstalled (its `ext-` topic is deleted), and README coverage for the adoption flow.

Parity ledger: #2974, bigquery-firestore-export §3.

PR #3186, opened 2026-09-11
fix(bigquery-firestore-export): remove the TRANSFER_CONFIG_NAME link path

cabljac:
The bigquery-firestore-export extension hard-codes `transferConfigName: undefined` in its `functions/src/config.ts` and exposes no param for it, so the `if (config.transferConfigName)` branch in `upsertTransferConfig` is dead code there. The kit made that branch live by adding a `TRANSFER_CONFIG_NAME` param and taking the link path in `handleUpsertTransferConfig`.

The ruling on #3165 is parity. Nobody is on `kits` yet, so nothing depends on it, and the only time the path was exercised it turned out to be broken (#2970). Shipping a kit-only feature that the extension never offered, in the state it is in, is not worth it.

Removed: the param, `transferConfigName` on both config interfaces and in `resolveConfig`, the link branch, and the `linkedTransferConfigMissing` and `linkedTopicMismatch` log helpers. Kept: `getTransferConfig` in `dts.ts`, which the update path still uses through `constructUpdateTransferConfigRequest`. README rows and the adoption sections are gone.

Tests now pin that `TRANSFER_CONFIG_NAME` is not a declared param, and the existing test that an instance with a Firestore config document carrying `name` goes down the update path. 81 tests pass, `tsc --noEmit` is clean, prettier reports no differences. Nothing was deployed or run live.

This is a breaking config change for anyone who set `TRANSFER_CONFIG_NAME` on `kits`; the CHANGELOG entry is marked breaking. Re-adding adoption as a deliberate feature is tracked in #3185, with PR #2960 as the starting point; none of it is implemented here.

This PR targets `kits`, not the default branch, so the issue will need closing by hand on merge.

Fixes #3165

### DIFF
--- a/kits/bigquery-firestore-export/CHANGELOG.md
+++ b/kits/bigquery-firestore-export/CHANGELOG.md
@@ -1,3 +1,4 @@
+- **Breaking:** `TRANSFER_CONFIG_NAME` and the link-an-existing-transfer-config path are removed for parity with the extension, which never exposed them. Re-adding adoption is tracked in #3185
 - Initial release of kit, see README for differences between the legacy extension and this kit
 - The instance id now comes from `FIREBASE_KIT_INSTANCE_ID`, which the Firebase CLI (15.27.0 or later) provides to each kit instance; `INSTANCE_ID` is no longer a configuration parameter
 - A BigQuery `TIME` column no longer crashes a run: its value is written to Firestore as the string BigQuery returned (`"10:30:00"`) rather than passed to `Timestamp.fromDate`, which threw and lost the whole run. This is a deliberate divergence from the legacy extension, which throws on the same line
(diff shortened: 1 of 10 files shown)


CorieW (approved):

CorieW (approved):
LGTM.

The extension's link branch is unreachable (`transferConfigName: undefined`, `functions/src/config.ts:25`), so removing it brings the kit's `handleUpsertTransferConfig` in line with the extension: look up by `extInstanceId`, create or update.

Checked on 65c975d: no stray references, `getTransferConfig` still serves the update path, `tsc` clean, 81/81 tests pass, CHANGELOG marks it breaking and links #3185.

Inherited, non-blocking: the first `extInstanceId` match wins silently, as in the extension.

Issue #3192, opened 2026-09-14
🐛 [All Kits] Kit Params are not specifying when they require a non-empty value.


ajperel:
### [REQUIRED] Step 2: Describe your configuration

- Kit name: All of them, though I first saw this in `delete-user-data`
- Extension version: latest
- Configuration values (redact info where appropriate): n/a

#### Steps to reproduce:

1. firebase functions:kits:install --package <kit>@next
2. Get to the point where you're configuring the kit params
3. Hit enter without entering text for them

##### Expected result

When you hit a param that _must_ have a value the CLI should force you to enter something (even if only a space, if you want more than that need regex validator)

##### Actual result

You can add a ton of empty params.

##### Analysis:

https://github.com/firebase/extensions/blob/kits/kits/delete-user-data/src/config.ts -- these are missing use of `input: { text: { nonEmpty: true } }` even when a param probably should require some value like FIRESTORE_PATHS.

We should broadly add `nonEmpty: true` across kits where required.

shettyvarun268:
Found a case with firestore-bigquery-export where this causes an issue. DATABASE, DATASET_ID, TABLE_ID, and COLLECTION_PATH should set nonEmpty: true, because without it firebase deploy succeeds with empty params while initBigQuerySync and fsexportbigquery fail silently at runtime, leading to data loss.

shettyvarun268:
- **`firestore-bigquery-export`**: `DATABASE`, `DATASET_ID`, `TABLE_ID`, `COLLECTION_PATH` (+ add missing `DATASET_LOCATION` and `BACKUP_COLLECTION` to `config.ts`)
- **`storage-resize-images`**: `IMG_BUCKET`, `IMG_SIZES`
- **`rtdb-limit-child-nodes`**: `NODE_PATH`, `DATABASE_INSTANCE`
- **`delete-user-data`**: `FIRESTORE_PATHS`, `RTDB_PATHS`, `STORAGE_PATHS` (at least one path or explicit validation so it doesn't silently no-op)
- **`firestore-counter`**: `SHARDS_COLLECTION` (currently relies on `assertRequiredParams` at runtime rather than `nonEmpty: true` at deploy time)
- **`firestore-send-email`**: `SMTP_CONNECTION_URI` / `MAIL_COLLECTION`
- **`firestore-translate-text`**: `COLLECTION_PATH`, `LANGUAGES`
- **`speech-to-text`**: `EXTENSION_BUCKET`
[...]

PR #3193, opened 2026-09-15
fix(kits): reject required params set to an empty value in .env

IzaakGough:
Closes #3192.

## What was broken

An empty entry in `.env` (`FIRESTORE_DATABASE_ID=`) overrides a param's declared default and skips the CLI's prompt-time validation, so it reaches the handlers silently. In delete-user-data that meant `getFirestore("")`, which resolves to the default database instead of the configured one.

The install prompt could produce that state too: `BACKUP_INSTANCE_ID` and `SELECTED_DATABASE_INSTANCE` declared no default and no validation rejecting `""`, and the CLI writes whatever a prompt resolves to into `.env`.

## What changed

Each kit declares its `required: true` params and checks them at module scope, so deploy-time discovery fails before the function ships rather than on the first event.

Those two params now declare `nonEmpty`, so the prompt re-asks instead of accepting an empty answer. Each kit's suite pins that invariant for the params it guards.

## Verification

All 13 kits build and their suites pass. Each of the 68 required params was blanked individually and blocked in firebase-tools' own `detectFromPort` discovery, and both changed declarations were driven through its `resolveParams`: `""` before, re-asked after. On a real project, a blanked `.env` entry now aborts the deploy naming the param.

## For the reviewer

The guard is duplicated per kit because there is no shared package under `kits/`.


CorieW:
**Blocking, parity gap:** `firestore-translate-text` lists the optional secret `GOOGLE_AI_API_KEY` in `REQUIRED_PARAMS`, so an empty placeholder secret would throw on every cold start.

Why:
- The extension marks the param optional: [`extension.yaml` L177-185](https://github.com/firebase/extensions/blob/a063eecd7719527278497735295959266792189c/firestore-translate-text/extension.yaml#L177-L185) has `required: false`. The kit requires it: [`src/config.ts` L159-166](https://github.com/firebase/extensions/blob/a063eecd7719527278497735295959266792189c/kits/firestore-translate-text/src/config.ts#L159-L166).
- Secrets are not in `.env`, so the guard cannot catch an empty value at deploy time. It only runs at cold start ([`src/index.ts` L37](https://github.com/firebase/extensions/blob/a063eecd7719527278497735295959266792189c/kits/firestore-translate-text/src/index.ts#L37)).
- The secret is bound unconditionally ([`src/index.ts` L122](https://github.com/firebase/extensions/blob/a063eecd7719527278497735295959266792189c/kits/firestore-translate-text/src/index.ts#L122)), so Cloud Translation and Vertex AI users must still create it. A placeholder of `""` or `" "` is plausible for a key the description calls "not required", and the extension accepts one.

Fix: remove `GOOGLE_AI_API_KEY` from `REQUIRED_PARAMS` in `src/config.ts`, and move it from `REQUIRED` to `OPTIONAL` in [`tests/required-params.test.ts` L19-28](https://github.com/firebase/extensions/blob/a063eecd7719527278497735295959266792189c/kits/firestore-translate-text/tests/required-params.test.ts#L19-L28). No other kit lists a secret.

CorieW (approved):
LGTM

Issue #3194, opened 2026-09-15
rtdb-limit-child-nodes: README documents RTDB_NODE_PATH and MAX_COUNT defaults that no longer exist


cabljac:
The kit README says the kit "falls back to its default of `messages`" when `NODE_PATH` is copied from an extension config (README.md:126-131), and has a section "`MAX_COUNT` now defaults to 100, and 0 is rejected" claiming both params have defaults so an incomplete config deploys (README.md:136-140).

Neither default exists. c4735460 (2026-08-24) removed them for parity: the extension declared both params required with no default, and a deploy with no `MAX_COUNT` had pruned every watched node down to 100. The README text is from 4c011e29 (2026-08-17) and was not updated.

Current behaviour on `kits` at 9cc453c5: `kits/rtdb-limit-child-nodes/src/config.ts:47-78` declares both params without a default, so the CLI prompts for them; `src/export-config.ts:46-51` throws at runtime on an empty path or a non-positive count.

Fix is docs only: drop the `messages` fallback sentence, and replace the defaults section with the one thing that is still true, that `MAX_COUNT` of 0 passes the extension's regex but is rejected at runtime.

IzaakGough:
Closing as fixed by #3196

PR #3196, opened 2026-09-15
docs(rtdb-limit-child-nodes): remove README claims of RTDB_NODE_PATH and MAX_COUNT defaults

cabljac:
The `rtdb-limit-child-nodes` kit README documented defaults that the code does
not have. It claimed the kit falls back to a node path of `messages` when only
`NODE_PATH` is set, and carried a section titled "`MAX_COUNT` now defaults to
100, and 0 is rejected" saying an incomplete config deploys instead of prompting.
c4735460 removed both defaults so the kit matches the extension, which declared
`NODE_PATH` and `MAX_COUNT` required with no default, but the README text from
4c011e29 was never updated.

This is a docs-only change. The `NODE_PATH` section now says what really happens
when the old key is copied over: the kit ignores it, `RTDB_NODE_PATH` stays
unset, and the CLI prompts for the path at deploy. The defaults section is
retitled and reduced to what is still true, that `MAX_COUNT` is an integer
setting whose `^\d+$` regex is kept verbatim from the extension, so `0` passes
validation and is rejected at runtime with "maxCount must be a positive
integer." The config table and the usage paragraph are corrected to match. No
code or tests were touched; `prettier --check` passes.

Fixes #3194


kits/rtdb-limit-child-nodes/README.md
```diff
@@ -41,8 +41,9 @@ and configure it with a `.env` (or `.env.<projectId>`).
 Importing the package without exporting its functions deploys nothing — the CLI
 only deploys what your entry file exports.
 
-Put `RTDB_NODE_PATH` and `SELECTED_DATABASE_INSTANCE` in `.env` so
-the trigger binds to the right database path and instance.
+Put `RTDB_NODE_PATH`, `MAX_COUNT` and `SELECTED_DATABASE_INSTANCE` in `.env` so
+the trigger binds to the right database path and instance. `RTDB_NODE_PATH` and
+`MAX_COUNT` have no default, so the CLI prompts for either one you leave out.
```
IzaakGough: `MAX_COUNT` is in the list now, but the purpose clause after it still only covers the path and the instance, so it reads as though `MAX_COUNT` takes part in binding the trigger.

Something like "... so the trigger binds to the right database path and instance, and the kit knows how many children to keep" would cover all three.

kits/rtdb-limit-child-nodes/README.md
```diff
@@ -126,20 +127,21 @@ bad values are caught, and where the function runs.
 
 Node.js reserves `NODE_PATH` for its own module resolution and overwrites it in
 the function runtime, so the setting had to be renamed. Copying `NODE_PATH` from
-an installed instance's config has no effect: the kit ignores it and falls back
-to its default of `messages`, so it watches the wrong path and silently trims
-nothing you care about. Rename the key to `RTDB_NODE_PATH` in your `.env`.
+an installed instance's config has no effect: the kit ignores it, which leaves
+`RTDB_NODE_PATH` unset, and a param with no default is prompted for, so the CLI
+asks you for the path at deploy time. Rename the key to `RTDB_NODE_PATH` in your
```
IzaakGough: This describes the interactive deploy. With no TTY, or in non-interactive mode, the CLI cannot prompt and the deploy fails on the missing value instead.

Worth a clause here, since someone copying `NODE_PATH` into a CI `.env` hits the failure rather than the prompt, and the failure is the case where the old key silently doing nothing is hardest to spot.

kits/rtdb-limit-child-nodes/README.md
```diff
@@ -126,20 +127,21 @@ bad values are caught, and where the function runs.
...
-instead of stopping to ask you. `MAX_COUNT` is also a proper integer setting now.
-The extension accepted `0`, which meant "delete every child on every write"; the
-kit rejects it along with negative and non-integer values.
+`MAX_COUNT` is a proper integer setting now, but the extension's `^\d+$`
+validation regex is kept verbatim, so `0` still passes validation. The extension
+took a `MAX_COUNT` of `0` to mean "delete every child on every write"; the kit
+rejects it, along with negative and non-integer values, with
+`maxCount must be a positive integer.` on the first write to the watched path.
 
 ### Bad settings surface on the first write, not at install
```
IzaakGough: This section describes the behaviour from before 051dd0a5 (`fix(kits): restore extension param validation regexes`, #2979). That commit put all three regexes back: `^\S+$` on the path (`src/config.ts:59`), `^\d+$` on `MAX_COUNT` (`:76`), and the instance regex in `DATABASE_INSTANCE_VALIDATION` (`:36-42`). The install prompt rejects those values again, so "a bad value deploys cleanly" does not hold, and as written it says the opposite of the section directly above it.

Two things are still true and worth keeping:

- A value set in `.env` skips validation. In firebase-tools 15.30.0, `resolveParams` assigns dotenv values straight through, and only params that are absent reach `promptParam`, which is the one place `validationRegex` is applied.
- `MAX_COUNT: 0` passes `^\d+$` even at the prompt, and throws `maxCount must be a positive integer.` on the first write.

`assertRequiredParams` (`src/config.ts:147-161`) narrows it further, rejecting a required param explicitly set to an empty value at module scope rather than on the first write.

Proposed replacement for 146-158:

```markdown
### Bad settings in `.env` surface on the first write

The install prompts reject a path containing spaces, a non-numeric `MAX_COUNT`
and an invalid database instance id before anything is deployed, the same as the
extension did. Those checks only run when the CLI prompts you, though: a value
you write into `.env` is used as-is, and only a required param left empty is
caught, at deploy, by `assertRequiredParams`.

`MAX_COUNT: 0` reaches the handler either way, since the extension's `^\d+$`
accepts it, and throws on every write to the watched path:

    maxCount must be a positive integer.

The parent node is not trimmed, and the only sign is the error in your function
logs.
```

I verified the `MAX_COUNT: 0` path end to end. I have not run a deploy with a path containing spaces or a non-numeric `MAX_COUNT` in `.env`, so the text above deliberately does not say what those do after they skip validation.

IzaakGough (commented):
Checked the diff against the code on this branch. The two corrections it makes are right: `src/config.ts:47,64` declare `RTDB_NODE_PATH` and `MAX_COUNT` with no `default`, and `resolveRtdbLimitConfig` only runs from `getConfig()` (`src/index.ts:47-55`), which is called from the handler, so "on the first write to the watched path" is accurate. The config table matches `RtdbLimitConfig`, where all three fields are non-optional.

One gap and two small wording notes inline. The gap is that the section immediately below the rewritten one carries the same kind of stale claim this PR is fixing.

IzaakGough:
I merged `kits` into this branch so the review references line up with what you will see. It was clean and your README diff is untouched; the five commits were #3193 and a translate-text fix, which add `assertRequiredParams` to this kit's `src/config.ts`. That function is relevant to the section I commented on.

kits/rtdb-limit-child-nodes/README.md
IzaakGough: Pushed ce24093f, which adds the non-interactive case to this paragraph:

> ... so the CLI asks you for the path at deploy time. A deploy that cannot prompt, such as one from CI, fails on the missing value instead.

For the record on where that comes from: `lib/command.js:162-166` in firebase-tools 15.30.0 forces `nonInteractive` when stdin is not a TTY, or on `--non-interactive` or `--json`, and `lib/deploy/functions/params.js:181-186` then throws rather than prompting. That is from reading 15.30.0, not from running a deploy.

kits/rtdb-limit-child-nodes/README.md
IzaakGough: Fixed in de685722. The clause now reads "... so the trigger binds to the right database path and instance, and the kit knows how many children to keep."

kits/rtdb-limit-child-nodes/README.md
IzaakGough: Fixed in 3abea8f2, using the replacement above with two small changes: `assertRequiredParams` is named without the line reference so it does not go stale, and the `MAX_COUNT` sentence says "regex accepts it" to make clear which check is doing the accepting.

The section no longer claims anything about what a path with spaces or a non-numeric `MAX_COUNT` does after skipping `.env` validation, since I have not run those.

IzaakGough:
A second review pass caught two things in the text I pushed earlier, both fixed in 496b6386.

`MAX_COUNT` said the kit rejects "non-integer values". `IntParam.runtimeValue()` in firebase-functions is `parseInt(process.env[name] || "0", 10) || 0`, so a fractional value from `.env` is truncated, not rejected: `10.7` keeps 10. `abc` does become 0 and is rejected, so the text now says "negative and non-numeric" and names the truncation.

The `.env` section heading said bad settings surface on the first write. Only `MAX_COUNT` has a runtime check. `resolveRtdbLimitConfig` strips slashes from the path and rejects it only when empty, and trims the instance id and rejects it only when empty, so a spaced path or a bad instance id from `.env` does not throw there. Retitled to "Values set in `.env` skip the install prompt's validation", which is what the body already described.

CorieW (approved):

Issue #3195, opened 2026-09-15
firestore-bigquery-export: WILDCARD_IDS prompt lost the extension's No/Yes labels


cabljac:
The extension declares `WILDCARD_IDS` as a select with labels `No` / `Yes` and values `false` / `true`, default `false` (`firestore-bigquery-export/extension.yaml:365-377`). The kit declares it as a bare `defineBoolean` with no `input` (`kits/firestore-bigquery-export/src/config.ts:394-400`), so the CLI shows a generic boolean prompt instead of the labelled select.

Stored values are unchanged, so a migrated `.env` reads correctly. This is the prompt only.

#3148 restored the extension's labels on the other seven `true` / `false` selects with `input: select({ Yes: true, No: false })` but did not touch this kit. Same change here, with `No` listed first to match the extension's order.

PR #3197, opened 2026-09-15
fix(firestore-bigquery-export): restore No/Yes labels on the WILDCARD_IDS prompt

cabljac:
The extension declared `WILDCARD_IDS` as a select with the labels `No` and `Yes` over the values `false` and `true`, default `false` (`firestore-bigquery-export/extension.yaml:365-377`). The kit declared it as a bare `defineBoolean` with no `input`, so the Firebase CLI showed a generic boolean prompt instead of the labelled select. #3148 restored the labels on the other seven yes/no selects across the kits but did not cover this one.

The param now carries `input: select({ No: false, Yes: true })`, with `No` first to match the extension's option order. The default and the stored values are untouched, so this is label-only and no `.env` from an earlier deploy needs editing.

A new `tests/param-declarations.test.ts` reads the declaration off `declaredParams` and pins the type, default, and option order; the kit's `config.test.ts` fakes `firebase-functions/params`, so the new file runs against the real module. `npx vitest run` passes (102 tests), `tsc -b` and prettier are clean. The CLI prompt itself was not exercised live - no `firebase deploy` was run - so the rendered prompt is inferred from the declaration shape, which matches the kits #3148 already fixed.

Fixes #3195


IzaakGough (approved):
lgtm!

Issue #3199, opened 2026-09-16
🐛 Eventarc custom event publishing crashes functions (unhandled rejections) & missing Declarative Security APIs/roles


shettyvarun268:
### Description
When an existing Firebase Extension with Eventarc custom events enabled is migrated via `firebase ext:migrate`, the CLI exports `EVENTARC_CHANNEL` and `EXT_SELECTED_EVENTS` into `.env.<project>`.

Currently, across the kits that emit Eventarc events, calling `eventChannel.publish(...)` throws an unhandled rejection if the channel is deleted, missing, or fails permissions. In several kits, `recordStartEvent()` is awaited *before* the main business logic (e.g. before resizing an image or before exporting to BigQuery). When Eventarc fails, the entire Cloud Function crashes and the actual work is never performed.

In addition, several kits are missing the required Eventarc publishing API and IAM publisher role in their Declarative Security blocks (`src/index.ts`), causing `HTTP 403 Forbidden` responses out-of-the-box.

---

#### 1. Wrap `eventChannel.publish()` in a try/catch
In `src/events.ts` (or wherever `eventChannel.publish` is called), wrap the publish calls in a `try / catch` so event publishing errors are logged as warnings and **never crash the Cloud Function**:

```ts
export const recordStartEvent = async (data: any) => {
  if (!eventChannel) return;
  try {
    return await eventChannel.publish({
      type: '...',
      data,
    });
  } catch (err) {
    logger.warn('Failed to publish Eventarc event (continuing execution):', err);
  }
};
```
*(Apply the same pattern to `recordCompletionEvent`, `recordSuccessEvent`, `recordErrorEvent`, etc.)*

#### 2. Ensure `eventarcpublishing.googleapis.com` is in `REQUIRED_APIS`
Add to `src/index.ts` so Declarative Security automatically enables the API on deploy:
```ts
{
  api: "eventarcpublishing.googleapis.com",
  reason: "Publishes custom events to Eventarc.",
}
```

#### 3. Ensure `roles/eventarc.publisher` is in `REQUIRED_ROLES`
Add to `src/index.ts` so the kit runtime service account has permission to publish to the channel:
```ts
requiresRole("roles/eventarc.publisher");
```

Note: We can wrap requiresAPI('eventarcpublishing.googleapis.com') and requiresRole('roles/eventarc.publisher') inside an if (process.env.EVENTARC_CHANNEL) check.

[...]

PR #3201, opened 2026-09-17
fix(kits): keep Eventarc publish failures from failing the function

cabljac:
Seven kits awaited `channel.publish()` with no error handling; five before the work itself. A deleted channel answers PERMISSION_DENIED, so the image was never resized or the document never exported.

- Each kit routes its `record*Event` helpers through one `publish` helper that warns and resolves. The `firestore-bigquery-export` rethrow is gone.
- `eventarcpublishing.googleapis.com` added to `REQUIRED_APIS` in all seven.
- No role changed: `roles/eventarc.publisher` is already declared in all seven, including the four #3199 lists as missing.
- Left unconditional rather than guarded by `EVENTARC_CHANNEL`: discovery runs without user `.env` (`firebase-tools` `prepare.js:200,506`), so the guard would always drop the declaration and every publish would 403.
- Tests: `events.test.ts` and `required-apis.test.ts` per kit, plus `eventarc-outage.test.ts` driving the real handler against a rejecting channel. Confirmed red pre-fix; all suites pass.
- Not verified against a live deleted channel. The `delete-user-data` `tsc` break predates this branch.

Fixes #3199


kits/firestore-bigquery-export/src/events.ts
```diff
@@ -50,21 +51,35 @@ export const setupEventChannel = () => {
...
 
   // Publish events for both OLD and NEW event types
-  return Promise.all(
+  await Promise.all(
     eventTypes.map((type) =>
-      eventChannel.publish({
+      publish({
         type,
         data,
       })
```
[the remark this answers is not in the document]
cabljac: Declining. The guard sits in `publish` on purpose: one choke point per kit is what makes the try/catch impossible to forget on the next `record*Event`, and a per-function guard duplicated across ~30 call sites in seven kits is the shape that let the missing error handling slip in originally. The allocation it saves is two template strings and two object literals on an invocation that also serializes a Firestore snapshot and inserts into BigQuery.

kits/storage-resize-images/tests/eventarc-outage.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  .fn()
+  .mockRejectedValue(
+    Object.assign(new Error("Permission denied"), { code: 403 })
+  );
+
+vi.mock("firebase-admin/eventarc", () => ({
+  getEventarc: () => ({ channel: () => ({ publish }) }),
+}));
+vi.mock("../src/logs");
+vi.mock("../src/filters", () => ({ shouldResize: vi.fn(() => false) }));
```
IzaakGough: With `shouldResize` mocked to `false`, `generateResizedImageHandler` returns right after `logs.start`, so this only proves `recordStartEvent` no longer throws. `recordStartResizeEvent`, `recordSuccessEvent` and `recordErrorEvent` are never reached, and neither are `handleFailedImage` or `deleteRemoteFile`. That makes the file header's "The resize must still run" a bit stronger than what is actually exercised, and a regression reintroducing a throwing `recordSuccessEvent` would leave this green.

The `firestore-bigquery-export` counterpart does drive the real path (it asserts `tracker.record` ran with `ChangeType.CREATE`). Letting `shouldResize` return `true` here and asserting on the resize would close the gap.

IzaakGough (approved):
Three small things, none blocking. One is inline; the other two are in `firestore-bigquery-export/src/handlers.ts` outside the diff, so they are here instead.

**`src/handlers.ts:268`, the `catch` in `handleSyncBigQueryTask` is now unreachable.** `recordSuccessEvent` delegates to the new `publish`, which swallows every rejection, and nothing else in the `try` can throw. So `logs.error(false, "Failed to record success event", ...)` can never fire, and the comment above it about Cloud Tasks retrying past the dedupe window describes a state the code cannot reach. If an `onSuccess` event is ever dropped, the only trace is the `logger.warn` in `events.ts`.

**`src/handlers.ts:114`, the comment in `enqueueForSync` is stale for the same reason.** It says "the event publish can itself reject", but `recordErrorEvent` no longer can, so the ordering constraint it states no longer exists.

kits/storage-resize-images/tests/eventarc-outage.test.ts
cabljac: Good catch, you're right. With `shouldResize` false the test only proved `recordStartEvent` no longer throws.

599a05e2 drives the real path: `shouldResize` returns true, with `content-filter`, `resize-image` and `file-operations` mocked at the boundary, and `deleteOriginal: "on_success"` so `deleteRemoteFile` is reached. Three cases now, the resize completing, `resizeImages` rejecting (`onError`), and a resize result reporting failure (`handleFailedImage`). All five `record*` helpers now publish against the denying channel, and all three go red if `publish` rethrows.

PR #3205, opened 2026-09-18
docs(firestore-bigquery-export-kit): document recovery for writes missed during migration

IzaakGough:
`ext:migrate` deploys the kit and then uninstalls the extension. The extension's trigger stops delivering as it is uninstalled while the kit's newly created trigger is not yet delivering reliably, so writes made in between reach neither exporter and appear in neither the changelog nor `BACKUP_COLLECTION`. The README did not mention this, or how to recover.

This PR adds a "Migrating from the extension" section, in two parts: migrating without a gap (decline the uninstall step, confirm from the kit's logs that it is delivering, then uninstall by hand), and recovering afterwards with `fs-bq-import-collection`, including what a re-import cannot restore. Documentation only.


cabljac (approved):
lgtm

PR #3207, opened 2026-09-18
chore(kits): Update all kits to regular dependency on firebase-functions@^7.4.0.

ajperel:
This is the first release version that fully supports kits features.


kits/bigquery-firestore-export/package.json
```diff
@@ -35,19 +35,18 @@
     "@firebase/app": "^0.16.0",
     "@google-cloud/bigquery": "^8.3.1",
     "@google-cloud/bigquery-data-transfer": "^5.0.1",
-    "@google-cloud/pubsub": "^4.11.0"
+    "@google-cloud/pubsub": "^4.11.0",
+    "firebase-functions": "^7.4.0"
   },
   "devDependencies": {
     "firebase-admin": "^14.4.0",
-    "firebase-functions": "^7.3.3-rc.3",
```
wandamora, line 42: was this mistakenly left as a dev/peer dependency before?

wandamora (approved):
I think this LG but want to understand why bigquery-firestore-export looks different.

kits/bigquery-firestore-export/package.json
ajperel: Yes. From an npm package perspective at this point things work equally well as regular or peer dependency. Because we use `npm-shrinkwrap.json` peer dependency has little value because unless the exact versions installed would be identical shrinkwrap forces two copies anyway. If we got rid of shrinkwrap there would be some benefits to using peer dependencies at the cost of making it much harder to test kits by just deploying them as regular functions.

For now I've made the decision to launch with regular dependencies.

ajperel: I'm even less concerned since we at the moment don't plan to launch this kit anyway. I could have left it alone but I figured better to have it in the state we think it _should_ be.

wandamora: Thanks for clarifying, and I agree this change is small enough to may as well make it.

